PR #5, opened 2018-08-28
Operator innards 1

gabemontero:
@openshift/sig-developer-experience ptal

broke changes out into 3 commits  ... hopefully the commit text sufficiently explains the rationale; in particular, I suspect the `operator-sdk build` could be generated by the repo build longer term, but I to confirm the state of api.ci and inclusion of the `operator-sdk` binary

Short term at least, the docker.io/gmontero/ose-cluster-samples-operator:v0.0.1 image should be accessible to anyone who want to try this out along with the various yaml in the deploy subdir.

also, I'm still adding unit tests, but decided at this point it would be good to continue that in parallel with implementation review and iteration


pkg/apis/samplesoperator/v1alpha1/types.go
```diff
@@ -47,30 +55,26 @@ type SamplesResourceSpec struct {
 type SamplesResourceStatus struct {
 	// Conditions represents the available maintenance status of the sample
 	// imagestreams and templates.
-	Conditions []SamplesResourceCondition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,8,rep,name=conditions"`
+	Conditions    []ResourceCondition
+	SecretCreated bool
```
bparees: why isn't this a condition?

pkg/stub/handler.go
```diff
@@ -2,96 +2,401 @@ package stub
...
+		_, err = h.secreteclientwrapper.Create("openshift", &secretToCreate)
+		if err != nil {
+			return h.processError(v1alpha1.SecretUpdateFailed, err, "failed to create registry dockerconfig secret in openshift namespace : %v")
+		}
+		logrus.Printf("created/updated dockerconfig secret %s in openshift namespace", h.operatorCfg.Spec.RegistryCredentials.Name)
+
+		h.operatorCfg.Status.SecretCreated = true
+		h.sdkwrapper.Update()
+		// TODO - do we want to duplicate "oc secrets link default <pull_secret_name> --for=pull"
+		// via client-go logic for the openshift namespace's default SA here as well ??
```
bparees: i don't think it is needed.

pkg/stub/handler.go
```diff
@@ -2,96 +2,401 @@ package stub
...
+			h.registrySecret = dockercfgSecret
+			return nil
+		}
+
+		if h.operatorCfg.Spec.RegistryCredentials.Name == dockercfgSecret.Name {
+			if h.registrySecret != nil &&
+				h.registrySecret.ResourceVersion == dockercfgSecret.ResourceVersion &&
+				h.operatorCfg.Status.SecretCreated &&
+				!event.Deleted {
+				return nil
```
bparees: since this could be a relist event and you might have missed the "modified" event when the secret was updated, I think you have to update the secret, just in case.

pkg/stub/handler.go
```diff
@@ -2,96 +2,401 @@ package stub
...
-					Image:   "busybox",
-					Command: []string{"sleep", "3600"},
-				},
-			},
-		},
-	}
-}*/
+func (h *Handler) updateDockerPullSpec(old string, imagestream *imagev1.ImageStream) {
+	if len(h.operatorCfg.Spec.SamplesRegistry) > 0 {
+		if strings.HasPrefix(imagestream.Spec.DockerImageRepository, "docker.io") {
```
bparees: all these docker.io constants should be "old" instead, right?

gabemontero: yep, and I am about the push unit tests that validate this path as well

also had to change this function's algorithm for proper use of `strings.Replace`, as well as better deal with the replacing of two possible prefixes in the RHEL registry case.

Will push shortly.

pkg/apis/samplesoperator/v1alpha1/types.go
gabemontero: there actually is a `SecretUpdateComplete` condition, but I found maintaining this separate boolean easier/cleaner looking than traversing the conditions array

if you feel strongly about this little piece of duplicate information, I'll remove and replace the use of this boolean with searching  the condition array 

I'll wait for your confirmation ...

pkg/stub/handler.go
gabemontero: thx for the confirmation

pkg/stub/handler.go
gabemontero: wouldn't the resource version change in that case ?  and if so, the resource version comparison would catch that ...

gabemontero:
pushed unit tests (with minor fixes) and changes resulting from a subset of @bparees comments, awaiting answers to my questions, will push further changes based on the response

sidebar: the vscode golang unit test support is pretty cool .... aside from easy debug and execution of individual test cases, or the entire suite, it can provide an examination of which code paths were executed during the test cases

pkg/apis/samplesoperator/v1alpha1/types.go
```diff
@@ -8,31 +8,39 @@ import (
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
 type SamplesResourceList struct {
-	metav1.TypeMeta `json:",inline"`
-	metav1.ListMeta `json:"metadata"`
-	Items           []SamplesResource `json:"items"`
```
bparees: why strip away the json metadata?

pkg/apis/samplesoperator/v1alpha1/types.go
bparees: I feel kinda strongly about it, yeah :)

I don't want us to have to maintain redundant info.

pkg/apis/samplesoperator/v1alpha1/types.go
```diff
@@ -47,30 +55,26 @@ type SamplesResourceSpec struct {
...
-	Status corev1.ConditionStatus `json:"status" protobuf:"bytes,2,opt,name=status,casttype=k8s.io/kubernetes/pkg/api/v1.ConditionStatus"`
-	// The last time this condition was updated.
-	LastUpdateTime metav1.Time `json:"lastUpdateTime,omitempty" protobuf:"bytes,6,opt,name=lastUpdateTime"`
-	// The last time the condition transitioned from one status to another.
-	LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty" protobuf:"bytes,3,opt,name=lastTransitionTime"`
-	// The reason for the condition's last transition.
-	Reason string `json:"reason,omitempty" protobuf:"bytes,4,opt,name=reason"`
+	Type ConditionType
+	// The time this condition occurred
+	Time metav1.Time
```
bparees: any reason for not following the pattern used elsewhere (w/ lastupdatetime, lasttransition, etc)?  I think the condition pattern was done that way to make it easier for cli tools to handle any object that exposed conditions, in a consistent way, so we probably want to follow it.

pkg/stub/handler.go
```diff
@@ -2,96 +2,410 @@ package stub
...
+	registrySecret *corev1.Secret
+
+	restconfig  *restclient.Config
+	tempclient  *templatev1client.TemplateV1Client
+	imageclient *imagev1client.ImageV1Client
+	coreclient  *corev1client.CoreV1Client
+
+	imageclientwrapper    ImageStreamClientWrapper
+	templateclientwrapper TemplateClientWrapper
+	secreteclientwrapper  SecretClientWrapper
```
bparees: extraneous `e` in that name.

pkg/stub/handler.go
```diff
@@ -2,96 +2,410 @@ package stub
...
+	filefinder         ResourceFileLister
+
+	namespace string
+
+	skippedTemplates    map[string]bool
+	skippedImagestreams map[string]bool
+}
+
+func (h *Handler) manageDockerCfgSecret(deleted bool) error {
+	err := h.secreteclientwrapper.Delete("openshift", h.registrySecret.Name, &metav1.DeleteOptions{})
```
bparees: I think this needs a // TODO to not just always delete the secret before creating/updating it.

ideally we'd only delete it when the operator's secret is deleted.

pkg/stub/handler.go
```diff
@@ -2,96 +2,410 @@ package stub
...
+		h.operatorCfg.Status.SecretCreated = true
+		h.sdkwrapper.Update()
+
+	}
+
+	status := v1alpha1.ResourceCondition{
+		Type: v1alpha1.SecretUpdateComplete,
+		Time: kapis.Now(),
+	}
+	h.operatorCfg.Status.Conditions = append(h.operatorCfg.Status.Conditions, status)
```
bparees: this is going to be an ever growing list.  I think the right way to use conditions is a fixed set of them, each having a true/false value (and a timestamp indicating when they were updated to that state).

In which case we'd want a condition like:
ImportSecretExists = true/false

I'm not sure if we care about having an "ImportSecretUpdated" condition or not, you can always refresh the "Exists" condition timestamp each time you update the secret.

bparees: (take a look at how other resources use conditions, but the above is at least my understanding)

pkg/stub/handler.go
```diff
@@ -2,96 +2,410 @@ package stub
...
+	err = h.sdkwrapper.Update()
+	if err != nil {
+		if !kerrors.IsConflict(err) {
+			return h.processError(v1alpha1.SamplesUpdateFailed, err, "failed adding success condition to config: %v")
+		}
+		logrus.Println("got conflict error on success status update, just going to forgo this time vs. retry")
```
bparees: // TODO - should requeue this event w/ a backoff for retry.

pkg/stub/handler.go
```diff
@@ -2,96 +2,410 @@ package stub
...
 }
 
 func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
-	switch o := event.Object.(type) {
+	switch event.Object.(type) {
+	case *corev1.Secret:
+		dockercfgSecret, _ := event.Object.(*corev1.Secret)
+
+		if h.operatorCfg == nil {
+			h.registrySecret = dockercfgSecret
```
bparees: holding the pointer seems risky.  should probably make a copy.

pkg/stub/handler.go
bparees: yeah, sorry, I guess it would. but see my new comment about not holding registrySecret as a pointer.

pkg/stub/handler.go
```diff
@@ -2,96 +2,410 @@ package stub
...
+		}
+
 	case *v1alpha1.SamplesResource:
-		logrus.Printf("ctrl cfg %#v", o)
-		restconfig, err := getRestConfig()
-		if err != nil {
-			logrus.Errorf("failed to get rest config : %v", err)
-			return err
+		if event.Deleted {
+			h.operatorCfg = nil
```
bparees: what actually happens in this scenario?  I think we should be recreating the default SamplesResource

pkg/stub/handler.go
```diff
@@ -2,96 +2,410 @@ package stub
...
-		if err != nil {
-			logrus.Errorf("failed to get template client : %v", err)
-			return err
+
+		opcfg, _ := event.Object.(*v1alpha1.SamplesResource)
+
+		if h.operatorCfg != nil && h.operatorCfg.ResourceVersion == opcfg.ResourceVersion {
+			return nil
+		}
+		h.operatorCfg = opcfg
```
bparees: since you're caching the resource version so you can "no-op" the next time you see this resource(unchanged), I don't think you can set the cache until you're sure you've successfully taken all necessary action to deal with it.

The same is true for where you're caching the secret.

If you were to cache the secret, then fail to copy the secret into the openshift namespace, you'd never have another chance to do it (until the operator was restarted)

pkg/stub/handler.go
```diff
@@ -2,96 +2,410 @@ package stub
...
+
+		if len(h.operatorCfg.Spec.RegistryCredentials.Name) > 0 &&
+			h.registrySecret != nil &&
+			!h.operatorCfg.Status.SecretCreated {
+			h.manageDockerCfgSecret(false)
```
bparees: is this needed at all?  Since you're watching for the creation of the secret, can't you just handle it there?

pkg/apis/samplesoperator/v1alpha1/types.go
bparees: were those internal types?  It should always be specified on external types.

pkg/apis/samplesoperator/v1alpha1/types.go
gabemontero: change part of next push

pkg/apis/samplesoperator/v1alpha1/types.go
gabemontero: It seemed overkill in that this is *NOT* a FSM like deployments.  In my mind these are event logs that can occur multiple times, independent of each other.

Perhaps I should call change the use of Condition to Event with the various fields/types to avoid confusion?

pkg/stub/handler.go
gabemontero: change part of next push

pkg/stub/handler.go
gabemontero: Since there is not "Status" with the secret, there wasn't any data to preserve in my mind between changes.

So rather than attempting to retrieve (an RPC), then creating or updating based on the results, seemed cumbersome.

I thought a `replace --force` was appropriate.  go-client does not wrapper that semantic in an API call.  So I implemented it manually.

Convinced?

pkg/apis/samplesoperator/v1alpha1/types.go
bparees: logging events (via the event recorder) is not entirely unreasonably, but i'd say that's orthogonal to the fact that we need to have conditions that reflect the current state/health of the operator.

Conditions I think we should have:
ImportCredentialsExist true/false
SamplesExist true/false

pkg/stub/handler.go
bparees: my primary concern here is we're creating a window where the secret doesn't exist and imports can fail.

pkg/stub/handler.go
gabemontero: Yeah, I did, and diverted per my reasoning in the other comment .... I was more interested in an audit/event log for debug purposes.  

I can certainly put a cap on the length of the array.

I still feel some resistance to conforming here, but digest my rationale, and I'll see what your response does to my resistance.

pkg/stub/handler.go
bparees: (or worse, we successfully delete the old secret and fail to create the new one)

pkg/stub/handler.go
bparees: yeah, i think if you want an audit log, that's what events are for. conditions should just reflect the current state, not the history.  And we need conditions (all operators must report conditions so we know the health of the system).  Adding event recording i'd say is bonus points (ie for now just putting that information in the operator's log output is probably sufficient)

pkg/stub/handler.go
gabemontero: Ok, in operator-sdk parlance, that would be returning an error.  The sdk's informer-sync.go error handling does the requeue, and employs rate limiting, using the k8s client-go APIs.

change to return the error part of the next push ... still going to distinguish conflict and non-conflict, so I can log the weird errors separately

gabemontero: the sdk does not expose a way for me to initiate requeue etc.

pkg/stub/handler.go
gabemontero: Deleting the SamplesResource object would allow a human to turn off the operator.

You'd have to clarify exactly how we recreate the default operator before I could weigh in on the value of it.  Just the version cached in memory?  Recreate the empty object in etcd and assume our architecture and distribution (centos/rhel) defaults?

bparees: Turning off/pausing the operator will be handled differently.  (Still being discussed exactly what the mechanism will be).  

But the general expectation is that every operator, when it starts up and does not see a relevant operator-resource instance (in this case samplesresource), will create a default one, and then get an event for that operator-resource being created, and handle it by instantiating whatever it's supposed to operate. 

So for the samples operator the auto-generated samplesresource would have to contain appropriate arch+distribution config, which we'd need to get from the "global cluster config" we've discussed.  For now you can hardcode it has x86+centos though :)

This is "4a" in the "startup" workflow of the samples operator proposal.  But since we can be restarted at anytime, our reaction to seeing the resource deleted should be the same as if we came up and it was missing.  So either we should do nothing at all (pending a restart), or we should do the same logic we would do on startup.

pkg/stub/handler.go
gabemontero: changes part of next push

on the secrets, since it is less disruptive, I'm not blocking an repeat update, and only using the field to manage the secret event arriving before the CR event

pkg/stub/handler.go
gabemontero: Yes this is needed if a Secret event arrives before the initial SamplesResource event

bparees: ah, good point, we don't want the import to fail due to the missing secret since nothing will automatically retry the import.  (We probably need to mark our imagestreams for scheduled imports at some point, but that's a side thread).

gabemontero:
I pushed a set of changes handling a subset of your 2nd set of comments @bparees 

I'll get to the json metadata and the condition stuff at a later date (we'll see how PTO goes next week).

pkg/stub/handler.go
gabemontero: id *does* result in a requeue with backoff

bparees: ok, i was confused by "the sdk does not expose a way for me to initiate requeue etc."

pkg/stub/handler.go
bparees: I was not suggesting we start marking imagestreams for scheduled import for now, that's a change that's going to have far reaching impacts on users.

Allowing them to override the samples to allow insecure import is a good point...I don't think ansible install handles that today, but it would be a nice to have here.  I wouldn't deem it critical (i'd really prefer to avoid doing much mutation of the imagestreams if we can help it, beyond the registry replacement you're already doing)

pkg/stub/handler.go
```diff
@@ -2,96 +2,436 @@ package stub
...
+		if h.operatorCfg.Spec.RegistryCredentials.Name == dockercfgSecret.Name {
+			err := h.manageDockerCfgSecret(event.Deleted, h.operatorCfg, dockercfgSecret)
+			if err != nil {
+				return err
+			}
+		}
+
+	case *v1alpha1.SamplesResource:
+		newStatus := corev1.ConditionTrue
+		var opcfg *v1alpha1.SamplesResource
```
bparees: same naming issue here (it's the samples config, not the operator's config)

pkg/stub/handler.go
```diff
@@ -2,96 +2,436 @@ package stub
...
-		/*err := sdk.Create(newbusyBoxPod(o))
-		if err != nil && !errors.IsAlreadyExists(err) {
-			logrus.Errorf("failed to create busybox pod : %v", err)
-			return err
-		}*/
+			if !kerrors.IsConflict(err) {
+				return h.processError(opcfg, v1alpha1.SamplesUpdateFailed, err, "failed adding success condition to config: %v")
+			}
+			logrus.Println("got conflict error on success status update, just going to forgo this time vs. retry")
+		}
```
bparees: same comment as the other status update block... I think all this can be wrapped inside the "if status!=newstatus" block?

pkg/stub/handler.go
```diff
@@ -2,96 +2,436 @@ package stub
...
+		status.Status = corev1.ConditionUnknown
+		status.LastTransitionTime = now
+	}
+	status.Message = log
+	opcfg.ConditionUpdate(status)
+	err2 := h.sdkwrapper.Update(opcfg)
+	if err2 != nil {
+		// just log this error
+		logrus.Printf("failed to add error condition to config status: %v", err2)
+	}
```
bparees: same here, wrap this inside the if status!=Unknown check.  No reason to update it if it isn't changing.

pkg/stub/handler.go
```diff
@@ -2,96 +2,436 @@ package stub
...
+		}
+
+	case *v1alpha1.SamplesResource:
+		newStatus := corev1.ConditionTrue
+		var opcfg *v1alpha1.SamplesResource
+		if event.Deleted {
+			h.operatorCfg = nil
+			// "4a" in the "startup" workflow, just create default
+			// resource and set up that way
+			opcfg = &v1alpha1.SamplesResource{}
```
bparees: Where do you actually create this resource?
Also we should be creating the resource w/ the default values set, rather than determing the defaults in code.  That way someone who looks at the resource instance can see what it is going to create.

bparees: basically in case of event.Deleted (or on initial startup if no SamplesResource exists) you should instantiate a new SamplesResource (with the defaults populated) and return.  Then you'll get a "SamplesResource created" event which you'll process as normal.

pkg/stub/handler.go
bparees: (Unless you're going to start setting the status.LastUpdateTime every time you re-assert the same status, but i'm concerned about us spamming the api server if we start doing that)

pkg/stub/handler.go
```diff
@@ -2,96 +2,436 @@ package stub
...
+				if opcfg.Spec.InstallType == v1alpha1.RHELSamplesDistribution {
+					h.updateDockerPullSpec([]string{"registry.redhat.io", "registry.access.redhat.com"}, imagestream, opcfg)
+				}
+				is, err := h.imageclientwrapper.Get("openshift", imagestream.Name, metav1.GetOptions{})
+				if err != nil && !kerrors.IsNotFound(err) {
+					return h.processError(opcfg, v1alpha1.SamplesUpdateFailed, err, "unexpected imagestream get error: %v")
+				}
+
+				if kerrors.IsNotFound(err) {
+					// testing showed that we get an empty is vs. nil in this case
```
bparees: that's...surprising.

bparees:
@gabemontero there's no logic in this operator yet to create the initial SamplesResource if one does not exist(as distinct from handling the delete event), is there?  if there is missed where it is.

gabemontero:
Correct @bparees re: https://github.com/openshift/cluster-samples-operator/pull/5#issuecomment-419472577

I'll add that to the next round of changes.

pkg/apis/samplesoperator/v1alpha1/types.go
```diff
@@ -22,42 +22,50 @@ type SamplesResource struct {
...
+)
+
 type SamplesResourceSpec struct {
 	// defaults to registry.redhat.io
-	SamplesRegistry string
-	// whether to install the rhel imagestreams
-	InstallRHEL bool
-	// whether to install the centos imagestreams
-	InstallCentos bool
+	SamplesRegistry string `json:"sampleRegistry,omitempty" protobuf:"bytes,1,opt,name=sampleRegistry"`
```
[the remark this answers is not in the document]
gabemontero: part of next push

Also, what the default is needs to be tweaked, as the InstallType has bearing.

pkg/stub/handler.go
```diff
@@ -2,96 +2,436 @@ package stub
...
+	operatorCfg.ConditionUpdate(status)
+	err := h.sdkwrapper.Update(operatorCfg)
+	if err != nil {
+		if !kerrors.IsConflict(err) {
+			return h.processError(operatorCfg, v1alpha1.SamplesUpdateFailed, err, "failed adding success condition to config: %v")
+		}
+		logrus.Println("got conflict error on success status update, just going to forgo this time vs. retry")
+		// just return the raw error and initiate the sdk's requeue/ratelimiter setup
+		return err
+	}
```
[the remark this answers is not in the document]
gabemontero: In the event of duplicate events, the Last Update Time would be more accurate if we did the updating regardless of the status equality.  It would correspond to the last event to come in.

Depends on how close together the duplicate events were.

Hence I did it this way.

Though I don't feel strongly about it if you are not concerned about updating the last access time in this case.... just let me know.

pkg/stub/handler.go
```diff
@@ -2,96 +2,436 @@ package stub
...
+
+		if h.operatorCfg != nil && h.operatorCfg.ResourceVersion == opcfg.ResourceVersion {
+			return nil
+		}
+
+		if len(opcfg.Spec.RegistryCredentials.Name) > 0 &&
+			len(h.registrySecret.ResourceVersion) > 0 &&
+			!opcfg.ConditionOccurred(v1alpha1.ImportCredentialsExists) {
+			h.manageDockerCfgSecret(false, opcfg, &h.registrySecret)
+		}
```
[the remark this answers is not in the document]
gabemontero: comment added

pkg/stub/handler.go
```diff
@@ -2,96 +2,436 @@ package stub
...
+	return &h
 }
 
 type Handler struct {
-	// Fill me
+	initter InClusterInitter
+
+	sdkwrapper SDKWrapper
+
+	operatorCfg    *v1alpha1.SamplesResource
```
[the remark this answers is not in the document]

pkg/apis/samplesoperator/v1alpha1/types.go
```diff
@@ -66,11 +74,43 @@ type SamplesResourceCondition struct {
...
+	LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty" protobuf:"bytes,4,opt,name=lastTransitionTime"`
 	// The reason for the condition's last transition.
-	Reason string `json:"reason,omitempty" protobuf:"bytes,4,opt,name=reason"`
+	Reason string `json:"reason,omitempty" protobuf:"bytes,5,opt,name=reason"`
 	// A human readable message indicating details about the transition.
-	Message string `json:"message,omitempty" protobuf:"bytes,5,opt,name=message"`
+	Message string `json:"message,omitempty" protobuf:"bytes,6,opt,name=message"`
+}
+
+func (s *SamplesResource) ConditionOccurred(c SamplesResourceConditionType) bool {
```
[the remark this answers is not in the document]
gabemontero: part of next push

pkg/apis/samplesoperator/v1alpha1/types.go
```diff
@@ -22,42 +22,50 @@ type SamplesResource struct {
...
 // the valid conditions of the SamplesResource
 
 const (
-	SamplesUpToDate SamplesResourceConditionType = "UpToDate"
-
-	SamplesUpdateFailed SamplesResourceConditionType = "UpdateFailed"
+	SamplesUpdateFailed     SamplesResourceConditionType = "SamplesResourceUpdateFailed"
+	SecretUpdateFailed      SamplesResourceConditionType = "SecretUpdateFailed"
+	ImportCredentialsExists SamplesResourceConditionType = "ImportCredentialsExists"
+	SamplesExists           SamplesResourceConditionType = "SamplesExists"
```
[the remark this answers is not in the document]
gabemontero: part of next push

pkg/apis/samplesoperator/v1alpha1/types.go
```diff
@@ -22,42 +22,50 @@ type SamplesResource struct {
...
+	// whether to install RHEL or Centos
+	InstallType SamplesDistributionType `json:"installType,omitempty" protobuf:"bytes,2,opt,name=installType"`
+
 	// which architecture(s) to install
-	Architectures []string
-	// reference to a secret containing a username+password/token
-	// for the registry.  Will be turned into a dockerconfig secret
+	Architectures []string `json:"architectures,omitempty" protobuf:"bytes,3,opt,name=architectures"`
+	// for the registry.  Required to already be in dockerconfig secret
+	// foramt so that it can be simply copied into the openshift namespace
```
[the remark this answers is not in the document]
gabemontero: part of next push

pkg/stub/handler.go
```diff
@@ -2,96 +2,436 @@ package stub
...
+			opcfg, _ = event.Object.(*v1alpha1.SamplesResource)
+		}
+
+		if h.operatorCfg != nil && h.operatorCfg.ResourceVersion == opcfg.ResourceVersion {
+			return nil
+		}
+
+		if len(opcfg.Spec.RegistryCredentials.Name) > 0 &&
+			len(h.registrySecret.ResourceVersion) > 0 &&
+			!opcfg.ConditionOccurred(v1alpha1.ImportCredentialsExists) {
```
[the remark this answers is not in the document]
gabemontero: Suppose it could have in theory been added then deleted before the first samples resource event came though

Added in next push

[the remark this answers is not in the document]
gabemontero: Though in the CondtionOccurred/ConditionsExists logic ... will rename to ConditionTrue

pkg/stub/handler.go
```diff
@@ -2,96 +2,436 @@ package stub
...
+	switch event.Object.(type) {
+	case *corev1.Secret:
+		dockercfgSecret, _ := event.Object.(*corev1.Secret)
+
+		if h.operatorCfg == nil {
+			dockercfgSecret.DeepCopyInto(&h.registrySecret)
+			return nil
+		}
+
+		if h.operatorCfg.Spec.RegistryCredentials.Name == dockercfgSecret.Name {
```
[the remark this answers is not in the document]
gabemontero: Not a big deal IMO ... let's leave it as is

pkg/stub/handler.go

pkg/stub/handler.go
gabemontero: OK that answers my earlier question ... will not worry about last update time.

pkg/stub/handler.go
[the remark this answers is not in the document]
gabemontero: saw comment below re: spamming master ... changing

pkg/stub/handler.go

pkg/stub/handler.go
gabemontero: OK, in the case of event.Deleted, seems straight forward.

The initial start up is trickier, as they may be creating the deployment and samples resource at the same time.  Because of that don't think you can just test for existence during the launch in main.go.

Something like a one time timer pop after a few seconds or something seems better.

Opinions?

pkg/stub/handler.go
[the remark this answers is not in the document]
bparees: ok for now, just keep in mind that if someone updates the name of the RegistryCredentials secret in the resource, the update logic will have to handle that scenario (go find the new secret, copy it over, etc).

pkg/stub/handler.go
bparees: fair point.  in general i think our expected pattern will be:
1) create operator
2) let operator create default resource
3) user tweaks resource if needed

but i can envision someone wanting to do what you described (create the resource and operator together because they are opinionated about the resource they want to create and they don't want to mutate the default one after it is created).  (though our installer will likely never do it that way).

I think your suggestion of having the operator startup logic set a timer and wait 5-10s before creating the default resource (assuming a resource hasn't been created by then) is a decent compromise to allow for that.

gabemontero:
ok, I've pushed an update for everything except the "if no SamplesResource available on startup" piece, while I wait for @bparees 's take on my timer need/idea

gabemontero:
On Sun, Sep 9, 2018 at 10:48 PM, Ben Parees <notifications@github.com>
wrote:

> *@bparees* commented on this pull request.
>
> In pkg/stub/handler.go
> <https://github.com/openshift/cluster-samples-operator/pull/5#discussion_r216186663>
> :
>
> > +
> +		if h.operatorCfg.Spec.RegistryCredentials.Name == dockercfgSecret.Name {
> +			err := h.manageDockerCfgSecret(event.Deleted, h.operatorCfg, dockercfgSecret)
> +			if err != nil {
> +				return err
> +			}
> +		}
> +
> +	case *v1alpha1.SamplesResource:
> +		newStatus := corev1.ConditionTrue
> +		var opcfg *v1alpha1.SamplesResource
> +		if event.Deleted {
> +			h.operatorCfg = nil
> +			// "4a" in the "startup" workflow, just create default
> +			// resource and set up that way
> +			opcfg = &v1alpha1.SamplesResource{}
>
> fair point. in general i think our expected pattern will be:
>
>    1. create operator
>    2. let operator create default resource
>    3. user tweaks resource if needed
>
> but i can envision someone wanting to do what you described (create the
> resource and operator together because they are opinionated about the
> resource they want to create and they don't want to mutate the default one
> after it is created). (though our installer will likely never do it that
> way).
>
> I think your suggestion of having the operator startup logic set a timer
> and wait 5-10s before creating the default resource (assuming a resource
> hasn't been created by then) is a decent compromise to allow for that.
>
Ah changes passing in the ether ... OK I'll put this in place on Monday.

bparees:
> while I wait for @bparees 's take on my timer need/idea

i'm ok with it

gabemontero:
On Sun, Sep 9, 2018 at 10:44 PM, Ben Parees <notifications@github.com>
wrote:

> *@bparees* commented on this pull request.
>
> In pkg/stub/handler.go
> <https://github.com/openshift/cluster-samples-operator/pull/5#discussion_r216186261>
> :
>
> > +	}
> +
> +	return nil
> +}
> +
> +func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
> +	switch event.Object.(type) {
> +	case *corev1.Secret:
> +		dockercfgSecret, _ := event.Object.(*corev1.Secret)
> +
> +		if h.operatorCfg == nil {
> +			dockercfgSecret.DeepCopyInto(&h.registrySecret)
> +			return nil
> +		}
> +
> +		if h.operatorCfg.Spec.RegistryCredentials.Name == dockercfgSecret.Name {
>
> ok for now, just keep in mind that if someone updates the name of the
> RegistryCredentials secret in the resource, the update logic will have to
> handle that scenario (go find the new secret, copy it over, etc).
>
Good point will need to handle this as well next round.

pkg/apis/samplesoperator/v1alpha1/types.go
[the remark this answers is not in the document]
bparees: i was thinking about how this would work when i was reviewing the PR.  I think there should be no default for this field.

if it's empty string, you don't touch the imagestreams you are loading.  (and you will either load the rhel or centos ones depending on the installtype).  the imagestreams have the correct registry baked in.

If you set this, you are saying "replace all the registry values in the imagestreams with this value".  So most people would not set it and it would be fine.  Only someone who wanted to get the imagestream content from some internal registry where they'd cloned all the images would set it.

pkg/apis/samplesoperator/v1alpha1/types.go
```diff
@@ -30,26 +30,38 @@ const (
...
+	// An InstallType of 'rhel' defaults to registry.redhat.io, and an InstallType of
+	// 'centos' defaults to docker.io.
 	SamplesRegistry string `json:"sampleRegistry,omitempty" protobuf:"bytes,1,opt,name=sampleRegistry"`
 
-	// whether to install RHEL or Centos
+	// InstallType specifies whether to install the RHEL or Centos distributions.
 	InstallType SamplesDistributionType `json:"installType,omitempty" protobuf:"bytes,2,opt,name=installType"`
 
-	// which architecture(s) to install
+	// Architectures determine which hardware architecture(s) to install, where x86_65 and ppc64le are the
```
bparees: x86_64.

pkg/stub/handler.go
```diff
@@ -152,88 +155,98 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
 	case *corev1.Secret:
 		dockercfgSecret, _ := event.Object.(*corev1.Secret)
 
-		if h.operatorCfg == nil {
+		if h.samplesResource == nil {
 			dockercfgSecret.DeepCopyInto(&h.registrySecret)
```
bparees: since we don't know the secretname that's going to be referenced by the sampleresource (because sampleresource is nil), how do we know we actually care about this secret?  (there could be multiple secrets in the namespace).

This is part of where i was going w/ wondering if us hardcoding the secret name would make your life easier.

Alternatively, if we want to keep the secretname as part of the resource spec, maybe it would make more sense to implement the logic as:

1) if we see a secret event, and we don't know the name of the secret we care about yet (because we haven't seen the samples resource yet) do nothing.
2) if we see a secret event we have seen the resource and this secret is the one we want, process it
3) if we see the sampleresource object and we *haven't* processed/found the secret yet, check to see if the secret referenced by the samplesresource exists.  If so, retrieve it and process it immediately (because clearly we already missed/ignored the creation event and we don't want to wait for a relist).

pkg/stub/handler.go

pkg/apis/samplesoperator/v1alpha1/types.go
gabemontero: part of next push

pkg/stub/handler.go
gabemontero: I'm going to switch to having the secret name be a constant

gabemontero:
While I have the changes ready based on @bparees 's last set of comments, continued testing is turning up update conflicts on the samples resource status update more consistently/frequently (though still somewhat intermittent).

When a conflict cycle occurs , the resource version no-op to filter duplicate events proves ineffective, and we end up in a looping cycle of sorts.

I'll push the comment response updates to continue the review process, but will be working some combination of retry on conflict and/or some for of deep equal examination of the Spec as a possible new version of filtering duplicate updates.

gabemontero:
ok comment updates pushed

bparees:
> continued testing is turning up update conflicts on the samples resource status update more consistently/frequently (though still somewhat intermittent).

any idea what could be causing that?  you should be the only process updating that resource...

pkg/apis/samplesoperator/v1alpha1/types.go
```diff
@@ -29,6 +29,16 @@ const (
...
 )
 
+const (
+	// SamplesRegistryCredentials is the name for a secret that contains a username+password/token
+	// for the registry, where if the secret is present, will be used for authentication.
+	// The corresponding secret is required to already be formatted as a
+	// dockerconfig secret so that it can just be copied
+	// to the openshift namespace
+	// for use during imagestream import.
+	SamplesRegistryCredentials = "override-samples-registry-credentials"
```
bparees: s/override//

pkg/stub/handler.go
```diff
@@ -60,6 +61,12 @@ func NewHandler() sdk.Handler {
 	h.skippedImagestreams = make(map[string]bool)
 	h.skippedTemplates = make(map[string]bool)
 
+	h.defaultResourceTimer = time.NewTimer(5 * time.Second)
+	go func() {
+		<-h.defaultResourceTimer.C
+		h.CreateDefaultResourceIfNeeded()
+	}()
```
bparees: @legionus this is a pattern you might want to copy in your operator.. basically a short grace period in case someone is trying to create their own resource object before the operator creates the default one.

pkg/stub/handler.go
```diff
@@ -155,44 +226,45 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
...
-			if err != nil {
-				return err
-			}
+		err := h.manageDockerCfgSecret(event.Deleted, h.samplesResource, dockercfgSecret)
+		if err != nil {
+			return err
 		}
 
 	case *v1alpha1.SamplesResource:
+		h.defaultResourceTimer.Stop()
```
bparees: seems unnecessary, i'd just let the timer fire and then no-op when it sees the resource already exists.

that way you don't need to add the defaultResourceTimer to the handler struct.

pkg/stub/handler.go
```diff
@@ -88,16 +95,77 @@ type Handler struct {
...
 	skippedTemplates    map[string]bool
 	skippedImagestreams map[string]bool
+
+	defaultResourceTimer *time.Timer
 }
 
-func (h *Handler) manageDockerCfgSecret(deleted bool, operatorCfg *v1alpha1.SamplesResource, secret *corev1.Secret) error {
-	// clear out temporary secret cache
-	h.registrySecret = corev1.Secret{}
+func (h *Handler) GoodStatusUpdate(srcfg *v1alpha1.SamplesResource, newStatus corev1.ConditionStatus) error {
```
bparees: how is this different from a "not good" status update?

why is it not used for the CredentialsExist status?

pkg/stub/handler.go
```diff
@@ -88,16 +95,77 @@ type Handler struct {
...
+		logrus.Println("")
+		logrus.Println("")
+		logrus.Println("")
+	}
+	h.samplesResource = srcfg
+	return nil
+}
+
+func (h *Handler) CreateDefaultResourceIfNeeded() error {
+	if h.samplesResource == nil {
```
bparees: feels racy, do we need a mutex around references to this?

pkg/stub/handler.go
```diff
@@ -155,44 +226,45 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
 	case *corev1.Secret:
 		dockercfgSecret, _ := event.Object.(*corev1.Secret)
 
+		if dockercfgSecret.Name != v1alpha1.SamplesRegistryCredentials {
+			return nil
+		}
+
 		if h.samplesResource == nil {
```
bparees: shouldn't you only be doing this copy/assignment if it's not a delete event?

pkg/stub/handler.go
gabemontero: originally did that, but it was sub optimal

the competing concerns:
- from earlier in our review, we don't set `h.sampleResource` on the first event until we've completed all the processing for the event 
- the actual time it takes to process all the imagestreams and templates from disk and update etcd can take more than 5 seconds
- so I would see the event processing and the default create happening at the same time, and we would then have to deal with concurrency at that level, including deciding which object to use

This was much more seamless IMO

pkg/stub/handler.go
gabemontero: the `processError` functionality is the "not good" status update  

I'm certainly open to another name

and yes, I could leverage this for the CredentialsExists status

pkg/stub/handler.go
gabemontero: will add that conditional

pkg/stub/handler.go
gabemontero: sure will add mutex

gabemontero:
of course the conflicts stopped happening after I added the get of the samples resource to compare if a conflict occurs .... we'll leave it in for now, see if it shows up again

gabemontero:
OK so I got some data on a conflict.  Weird to me.

The resource versions are different, but when comparing field by field, the only differences are the timestamps in the status condition.

The one I'm trying to update with:

```
				LastUpdateTime:v1.Time{Time:time.Time{wall:0xbede2a92ac41121c, ext:12589094030, loc:(*time.Location)(0x1a4e9c0)}}, 
				LastTransitionTime:v1.Time{Time:time.Time{wall:0xbede2a92ac41121c, ext:12589094030, loc:(*time.Location)(0x1a4e9c0)}}, 

```

And the one that I fetched from etcd after the conflict (where it has a higher resource version) is:

```
				LastUpdateTime:v1.Time{Time:time.Time{wall:0x0, ext:63672297923, loc:(*time.Location)(0x1a4e9c0)}}, 
				LastTransitionTime:v1.Time{Time:time.Time{wall:0x0, ext:63672297923, loc:(*time.Location)(0x1a4e9c0)}}, 

```

It is like something tweaked my status condition's 2 time stamps

Definitely is not consistent ... this is not always happening

pkg/stub/handler.go
gabemontero: with mutex wrapping the setting of h.samplesResource, getting rid of call to Stop, storage of timer

gabemontero:
updates pushed

pkg/stub/handler.go
```diff
@@ -169,7 +195,15 @@ func (h *Handler) manageDockerCfgSecret(deleted bool, samplesResource *v1alpha1.
 		}
 		logrus.Printf("registry dockerconfig secret %s was deleted", secret.Name)
 		newStatus = corev1.ConditionFalse
+		h.registrySecret = nil
 	} else {
+		if h.registrySecret != nil {
+			currentVersion, _ := strconv.Atoi(h.registrySecret.ResourceVersion)
+			newVersion, _ := strconv.Atoi(secret.ResourceVersion)
+			if newVersion <= currentVersion {
```
bparees: when would new version be less than current version?

gabemontero: when a paranoid developer thought he saw such an event, albeit very intermittently, during a flurry of testing on top of all this new stuff 

this is definitely erring on the side of caution / paranoia

bparees: ok

deploy/openshift-rbac.yaml
```diff
@@ -0,0 +1,14 @@
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: cluster-samples-operator-openshift-edit
+  namespace: openshift
+subjects:
+- kind: ServiceAccount
+  name: default
+  # set namespace to wherever the operator is running
+  namespace: test-operator
```
bparees: set this to openshift-cluster-samples-operator

tmp/build/Dockerfile
```diff
@@ -1,6 +1,24 @@
-FROM alpine:3.6
+#FROM alpine:3.6
+FROM openshift/origin
 
-RUN adduser -D cluster-samples-operator
+RUN INSTALL_PKGS="dejavu-sans-fonts wget rsync gettext git tar zip unzip shadow-utils" && \
```
bparees: you actually need all these packages?

tmp/build/Dockerfile
```diff
@@ -1,6 +1,24 @@
-FROM alpine:3.6
+#FROM alpine:3.6
```
bparees: we should be basically ignoring this file at this point.  The one that matters is the Dockerfile in the root of the repo.

pkg/stub/handler.go
```diff
@@ -2,96 +2,531 @@ package stub
...
+	mutex *sync.Mutex
 }
 
-func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
-	switch o := event.Object.(type) {
-	case *v1alpha1.SamplesResource:
-		logrus.Printf("ctrl cfg %#v", o)
-		restconfig, err := getRestConfig()
+func (h *Handler) GoodStatusUpdate(srcfg *v1alpha1.SamplesResource, newStatus corev1.ConditionStatus, condition v1alpha1.SamplesResourceConditionType) error {
+	status := srcfg.Condition(condition)
```
bparees: s/status/condition/

pkg/stub/handler.go
```diff
@@ -2,96 +2,531 @@ package stub
...
+			srcfg.ConditionUpdate(status)
+			err = h.sdkwrapper.Update(srcfg)
+			if err != nil {
+				// just give up this time
+				return h.processError(srcfg, v1alpha1.SamplesUpdateFailed, err, "failed to update status after conflict retry: %v")
+			}
+		} else {
+			// get the latest copy in case something modified the object underneath us
+			// per observations during testing
+			sr, _ := h.sdkwrapper.Get(srcfg.Name, srcfg.Namespace)
```
bparees: we just did an update of the object, why would we need to fetch it back? if it does change, we'd get a modified event for it, no?

pkg/stub/handler.go
```diff
@@ -2,96 +2,531 @@ package stub
...
-		logrus.Printf("template list %#v", templist)
-		/*err := sdk.Create(newbusyBoxPod(o))
-		if err != nil && !errors.IsAlreadyExists(err) {
-			logrus.Errorf("failed to create busybox pod : %v", err)
+
+		if h.samplesResource == nil {
+			if !event.Deleted {
+				h.registrySecret = dockercfgSecret
+			}
+			return nil
```
bparees: why don't we still want to setup the secret in the openshift namespace in this scenario?

pkg/stub/handler.go
```diff
@@ -2,96 +2,531 @@ package stub
...
+
+		// coordinate with timer's check on creating
+		// default resource ... looks at h.sampleResource,
+		// which is not set until this whole case is completed
+		h.mutex.Lock()
+		defer h.mutex.Unlock()
+
+		if h.samplesResource != nil &&
+			// in case a new, non-default resource is created
+			h.samplesResource.Name == srcfg.Name {
```
bparees: i don't know what we'd do if someone did create a second samplesresource instance?

pkg/stub/handler.go
```diff
@@ -2,96 +2,531 @@ package stub
...
+			currVersion, _ := strconv.Atoi(h.samplesResource.ResourceVersion)
+			newVersion, _ := strconv.Atoi(srcfg.ResourceVersion)
+			if newVersion <= currVersion {
+				return nil
+			}
+		}
+
+		// if the secret event came in before the samples resource event,
+		// it could not be processed (though it would have been cached in
+		// the Handler struct);  process it now
```
bparees: so why could it not be processed?  i think this logic is a holdover from before you switched to a hardcoded secret name,where you would have needed to see the samplesresource to determine the secret name.

bparees:
something else we're going to have to deal with at some point:  if we miss a delete event, a "relist" won't help us.  Something needs to periodically check if the Secret or SamplesResource has gone missing in case we missed the delete event.

tmp/build/Dockerfile
gabemontero: did not port to new Dockerfile ... I started with them as possible placeholders for future need (dev or debug or support)

pkg/stub/handler.go
gabemontero: Yeah similar to the secret name, I was wondering if we should pin down via constant the name of the samplesresource we deal with, and document that is the only name we consider.  Just stopped short of implementing it, and decided this minimal step forward was enough to generate the discussion.

Thoughts on the static, well defined name ?

bparees: @jwforres is there any current thinking around how "singleton" operators are supposed to deal with this, or even per-namespace operators that can still only handle a single instance (per namespace) of the thing they instantiate?

What are we supposed to do if someone creates 2+ CRs?  Can our operator simply say "we only look at the CR named "default" and ignore all the others, and you must name your CR instance "foo"?  Is there some other pattern that's been established?

pkg/stub/handler.go
gabemontero: Yeah this stems from the intermittent, repeated conflict scenario, I cited earlier.

I'll try to concisely describe:
1) we have an event come in at version A, but in processing, it  gets a conflict cause etcd has version B (where numerically B > A) ... version B's status conditions timestamps have been modified by something, and are different than version A's ... everything else between the 2 versions same
2) we return error 
3) sdk does requeue / retry based on us returning error 
4) cause of retry, we get the *same* event again, i.e. the same now stale version
5) sdk is hardwired to retry 15 times .. so lots of thrashing as we update the samples content again, since we did not update our samplesresource cache, since we did not successfully update the status conditions

With this change, I have since seen this occur (i.e. item 1 above) at least twice now, and by doing this, have avoided the churn around 2) through 5) 

So, yes, *WHAT* changes the timestamp on the condition ???  I could certainly use any suggestions on how to debug, and sure, we could open a bug somewhere (where exactly?), but with this being intermittent, the reality is unless we can get the precise debug data needed, the bug will sit unresolved.

And this workaround is simple in the interim and is working in my testing.

pkg/stub/handler.go
gabemontero: we can't update at the status without the samplesresource ... so we either 
a) create secret now, then save data somewhere so we know to add the secret import condition when we get the sample resource
b) create the secret and add the condition at the same time

b) seemed cleaner to me, as well as less churn from our prior form

pkg/stub/handler.go
bparees: but you're already re-getting the fresh object if you get an update conflict and retrying the update, above, so doesn't that already handle the conflict?

pkg/stub/handler.go
bparees: hum. ok.

pkg/stub/handler.go
gabemontero: see rationale at https://github.com/openshift/cluster-samples-operator/pull/5#discussion_r217518521

a) create secret, then update conditions when we get samplesresource
b) create secret and update conditions at same time when we get samplesresource

I liked b) better

pkg/stub/handler.go
gabemontero: After the fact I was worried about my status update possibly updating the version yet again, but yeah, in hindsight that is too paranoid (even for me) .... only spec/objectmeta/metadata updates should result in the version getting updated.

I'll remove that `Get` in the non-error path.

bparees: well i would expect the status update to change the resource version because the resourceversion just comes from etcd as far as i know and doesn't care what the change was... but you'd still get a new modified event due to that update and then your handler would process it, as normal (ideally process it as a no-op since nothing in the spec changed).

gabemontero:
pushed initial updates for round 5 ... the multiple SR instances question is the outstanding piece

pkg/stub/handler.go
gabemontero: That is not what I see in practice ... when thing run without issue
1) I get the initial event
2) process it, update the status
3) i don't get a new version event as a result of the updated status ... the operator is quiet until the relist happens

Other than resource version, and not comparing the specs for deltas before proceding

bparees: ok, i may not understand how resourceversion is managed by k8s, not going to worry about it for now if things are working as you'd hope.

pkg/stub/handler.go
```diff
@@ -2,96 +2,524 @@ package stub
...
+		if h.registrySecret != nil &&
+			!srcfg.ConditionTrue(v1alpha1.ImportCredentialsExist) &&
+			h.registrySecret.Name == v1alpha1.SamplesRegistryCredentials {
```
bparees: this check seems unnecessary.  we should not have assigned h.registrySecret if secret.Name was not the one we care about.

pkg/stub/handler.go
jwforres: Using a static name is one option, the other one i've heard discussed is marking the first one you see with an annotation or something and that becomes the one the operator is managing. I can see downsides to both. But the annotation method seems like it would be easier for an admin to screw things up and the operator to not know which one it is supposed to be managing, although since you in theory have to be able to take a CR in and out of operator management anyway, it should be resolvable if they do mess it up.

bparees: static name it is.

As for the name, how about "openshift-samples"?

pkg/stub/handler.go
gabemontero: Well I had the check here and line 242 in the main `Handle` case block for the secret watch.

I'll centralize it in `manageDockerCfgSecret`

bparees: well my point was because you had the other check, you already knew that if h.registrySecret was non-nil, it was because we set it to a value where the name matched the name we wanted.

at least i think that was my reading of the code.  either way, yeah it should only need to be checked in one place.

gabemontero:
OK pushed updates to complete round 5

deploy/cr.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: samplesoperator.config.openshift.io/v1alpha1
 kind: SamplesResource
 metadata:
-  name: example
+  name: openshift-samples
```
bparees: openshift-cluster-samples-operator

bparees: sorry, nm

bparees:
/lgtm

let's see if a new image gets pushed to dockerhub when this merges...it's supposed to.

gabemontero:
Looks like it did get updated @bparees : 
```
gmontero ~ $ docker pull docker.io/openshift/origin-cluster-samples-operator
Using default tag: latest
latest: Pulling from openshift/origin-cluster-samples-operator



Digest: sha256:65c03022a959a2c1051a12d1856da8a3155bcf50ccc0c3033b2f36c99e825a63
Status: Downloaded newer image for openshift/origin-cluster-samples-operator:latest
gmontero ~ $ docker images | grep cluster 
openshift/origin-cluster-samples-operator            latest              0b26d87c20ea        About an hour ago   415MB
gmontero ~ $ 
```

PR #6, opened 2018-08-30
Library output

gabemontero:
@bparees - fyi / ptal

### DIFF
--- a/tmp/build/assets/arch/ppc64le/official/ruby/imagestreams/ruby_ppc64le-rhel7.json
+++ b/tmp/build/assets/arch/ppc64le/official/ruby/imagestreams/ruby_ppc64le-rhel7.json
@@ -1,0 +1,54 @@
+{
+    "apiVersion": "v1",
+    "kind": "ImageStream",
+    "metadata": {
+        "annotations": {
+            "openshift.io/display-name": "Ruby (ppc64le)"
+        },
+        "name": "ruby_ppc64le"
+    },
+    "spec": {
+        "tags": [
+            {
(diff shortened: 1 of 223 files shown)


bparees:
we shouldn't need both "official" and "arch/x86_64/official"

bparees:
also let's give the toplevel dir a better name.  "assets" or "library" or something.   `/tmp/build/library-master` is an odd/meaningless path.

`library/x86_64` and `library/ppc64le` should be sufficient?

gabemontero:
the `tmp/build` sudirs stems from how `operator-sdk` works ... I could not get the image building to work if it was not under `tmp/build`

gabemontero:
`library-master` was trying to convey the master branch of library ... despite our prior discussion about always pulling from openshift/library:master, I wanted to clarify

that said, we can always rename if need be later on

I'll change `library-master` to `assets`

gabemontero:
@bparees pushed updates for you comments, except the `tmp/build` piece I explained in https://github.com/openshift/cluster-samples-operator/pull/6#issuecomment-417366637

bparees:
> the tmp/build sudirs stems from how operator-sdk works

i think we're going to have to talk about how we build the operator image, because the way operator-sdk does it isn't a way we can easily support when we are building the operator images from within openshift, I do not believe.  But sure, whatever we need to do for now.

bparees:
looks better, i'd say merge this and we can iterate if/when we can resolve the "/tmp/build" requirement.

(i'm also not sure we really need the "arch" dir in the assets path)

PR #13, opened 2018-09-19
deploy: add namespace to Role and RoleBindings

abhinavdahiya:
Anybody using these manfiests doesn't have information or context
to find which namespace these Namespaced Object need to be installed.
For example, ClusterVersionOperator wouldn't know which namespace to choose.

### DIFF
--- a/deploy/02-rbac.yaml
+++ b/deploy/02-rbac.yaml
@@ -2,6 +2,7 @@ kind: Role
 apiVersion: rbac.authorization.k8s.io/v1beta1
 metadata:
   name: cluster-samples-operator
+  namespace: openshift-cluster-samples-operator
 rules:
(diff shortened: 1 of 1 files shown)


gabemontero:
my understanding from @bparees was the namespace for the operator in 4.0 would be constant, always this particular name provided, and that would be the only namespace identifier supported

but let's wait until he gets online and can chime in specifically before making an decisions wrt this PR

bparees:
> my understanding from @bparees was the namespace for the operator in 4.0 would be constant, always this particular name provided, and that would be the only namespace identifier supported

that's correct, but @abhinavdahiya's PR is also correct as it will ensure the operators resources get created in that constant namespace.
 
/lgtm

PR #15, opened 2018-09-19
Group skip dir name

gabemontero:
@openshift/sig-developer-experience fyi/ptal

a collection of changes based on ideas of mine or discussions with @bparees on #aos-devexp

### DIFF
--- a/pkg/apis/samplesoperator/v1alpha1/types.go
+++ b/pkg/apis/samplesoperator/v1alpha1/types.go
@@ -80,12 +80,21 @@ type SamplesResourceConditionType string
 // the valid conditions of the SamplesResource
 
 const (
-	SamplesUpdateFailed    SamplesResourceConditionType = "SamplesResourceUpdateFailed"
-	SecretUpdateFailed     SamplesResourceConditionType = "SecretUpdateFailed"
+	// ImportCredentialsExists represents the state of any credentials specified by
+	// the SamplesRegistry field in the Spec.
 	ImportCredentialsExist SamplesResourceConditionType = "ImportCredentialsExists"
-	SamplesExist           SamplesResourceConditionType = "SamplesExists"
+	// SamplesExist represents whether an incoming SamplesResource has been successfully
+	// processed or not all, or whether the last SamplesResource to come in has been
+	// successfully processed.
+	SamplesExist SamplesResourceConditionType = "SamplesExists"
+	// ConfigurationValid represents whether the latest SamplesResource to come in
+	// tried to make a support configuration change.  Currently, changes to the
+	// InstallType and Architectures list after initial processing is not allowed.
+	ConfigurationValid SamplesResourceConditionType = "ConfigurationValid"
 )
 
+// SamplesResourceCondition captures various conditions of the SamplesResource
+// as entries are processed.
 type SamplesResourceCondition struct {
 	// Type of condition.
 	Type SamplesResourceConditionType `json:"type" protobuf:"bytes,1,opt,name=type,casttype=SamplesResourceConditionType"`
@@ -126,8 +135,7 @@ func (s *SamplesResource) Condition(c SamplesResourceConditionType) *SamplesReso
 		}
 	}
 	newCondition := SamplesResourceCondition{
-		Type:   c,
-		Status: corev1.ConditionUnknown,
+		Type: c,
 	}
 	s.Status.Conditions = append(s.Status.Conditions, newCondition)
 	return &newCondition
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -42,6 +42,7 @@ const (
 	ppc64OCPContentRootDir = "/opt/openshift/operator/ocp-ppc64le"
 	x86                    = "x86_64"
 	ppc                    = "ppc64le"
+	installtypekey         = "installtype"
 )
 
 func NewHandler() sdk.Handler {
@@ -59,6 +60,7 @@ func NewHandler() sdk.Handler {
 	h.imageclientwrapper = &defaultImageStreamClientWrapper{h: &h}
 	h.templateclientwrapper = &defaultTemplateClientWrapper{h: &h}
 	h.secretclientwrapper = &defaultSecretClientWrapper{h: &h}
+	h.configmapclientwrapper = &defaultConfigMapClientWrapper{h: &h}
 
 	h.namespace = getNamespace()
 
@@ -88,9 +90,10 @@ type Handler struct {
 	imageclient *imagev1client.ImageV1Client
 	coreclient  *corev1client.CoreV1Client
 
-	imageclientwrapper    ImageStreamClientWrapper
-	templateclientwrapper TemplateClientWrapper
-	secretclientwrapper   SecretClientWrapper
+	imageclientwrapper     ImageStreamClientWrapper
+	templateclientwrapper  TemplateClientWrapper
+	secretclientwrapper    SecretClientWrapper
+	configmapclientwrapper ConfigMapClientWrapper
 
 	fileimagegetter    ImageStreamFromFileGetter
 	filetemplategetter TemplateFromFileGetter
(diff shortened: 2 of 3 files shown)


pkg/apis/samplesoperator/v1alpha1/types.go
```diff
@@ -119,6 +128,15 @@ func (s *SamplesResource) ConditionUpdate(c *SamplesResourceCondition) {
 	}
 }
 
+func (s *SamplesResource) StatusUpdate(c SamplesResourceConditionType) {
+	if c == SamplesExist {
+		s.Status.InstallType = s.Spec.InstallType
+		for _, arch := range s.Spec.Architectures {
+			s.Status.Architectures = append(s.Status.Architectures, arch)
```
bparees: should do this in a way that ensures we don't acrue dupes in the future

pkg/stub/handler.go
```diff
@@ -278,6 +281,32 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
...
+
+			if len(srcfg.Spec.Architectures) == 0 &&
+				(len(h.samplesResource.Status.Architectures) != 1 ||
+					h.samplesResource.Status.Architectures[0] != x86) {
+				err := fmt.Errorf("cannot change from architecture list %#v to %#v",
+					h.samplesResource.Status.Architectures, srcfg.Spec.Architectures)
+				return h.processError(srcfg, v1alpha1.SamplesUpdateFailed, err, "%v")
+			}
+			if len(srcfg.Spec.Architectures) != 0 &&
+				!reflect.DeepEqual(srcfg.Spec.Architectures, h.samplesResource.Status.Architectures) {
```
bparees: this is an order dependent check, right?  seems fragile.

pkg/stub/handler.go
```diff
@@ -382,18 +411,24 @@ func (h *Handler) processFiles(dir string, files []os.FileInfo, opcfg *v1alpha1.
 		logrus.Printf("processing file %s from dir %s", file.Name(), dir)
 
 		if strings.HasSuffix(dir, "imagestreams") {
+			// allow setting of product dir as found in tmp/build/assests/operator as a way to skip as set of imagestreams
+			// i.e. .../php/imagestreams/...
+			// setting "php" will skip all the imagestreams under php/imagestreams
```
bparees: I worry about this having unintended consequences.  The users setting skippedImageStreams don't have any insight into the underlying directory structure the operator is processing, so they could inadvertently skip things they don't intend to/don't understand.  The names are really the only thing they are exposed to.

pkg/stub/handler.go
```diff
@@ -422,6 +457,17 @@ func (h *Handler) processFiles(dir string, files []os.FileInfo, opcfg *v1alpha1.
 		}
 
 		if strings.HasSuffix(dir, "templates") {
+			// allow setting of product dir as found in tmp/build/assests/operator as a way to skip as set of templates
+			// i.e. .../cakephp/templates/...
+			// setting "cakephp" will skip all the templates under cakephp/templates
```
bparees: same concern here.  What was the motivation for adding this?

pkg/stub/handler.go
```diff
@@ -460,23 +506,27 @@ func (h *Handler) processFiles(dir string, files []os.FileInfo, opcfg *v1alpha1.
...
-		if !replaced {
-			opcfg.Spec.SamplesRegistry = opcfg.Spec.SamplesRegistry + "/" + imagestream.Spec.DockerImageRepository
 		}
 
 		for _, tagref := range imagestream.Spec.Tags {
 			replaced := false
 			for _, old := range oldies {
 				if tagref.From != nil && tagref.From.Kind == "DockerImage" {
-					if strings.HasPrefix(tagref.From.Name, old) {
+					if !strings.HasPrefix(tagref.From.Name, opcfg.Spec.SamplesRegistry) &&
```
bparees: why not move this outside the "oldies" loop like you did for DockerImageRepository?

pkg/apis/samplesoperator/v1alpha1/types.go
gabemontero: Had a change ready right before lunch, but given the direction to move this out of status and use a config map, this will get reworked anyway.

Perhaps with config maps, we can use key/value (with well defined keys) vs. arrays, and can by nature prevent the array accrual concern.  I'll sort it out as I revisit the ConfigMap particulars.

pkg/stub/handler.go
gabemontero: Are you referring to the order of the conditionals within this if, or this if in relation to the prior if ?

That said, the switch to a k/v ConfigMap will reswizzle this anyway.

pkg/stub/handler.go
gabemontero: I admittedly was anticipating need here based on how ansible surfaced the on/off switches for which samples were installed.  It had higher level categories.

Sure, I can back this off, and we can wait and see if the request for skipping a group of content emerges, and based on how it manifests, warrants consideration.

pkg/stub/handler.go
gabemontero: explained above ... will remove

pkg/stub/handler.go
gabemontero: Just missed that ... thanks, part of next push

pkg/stub/handler.go
bparees: i'm referring to the order of the values in the arrays.  You're comparing two arrays, I assume that compare will fail if the two arrays contain the same values in a different order, whereas from our perspective they should be considered equal.

gabemontero: ah gotcha 

and yep, a stackoverflow lead me to believe it would not matter, but just tried it in go playground and order does matter

But as I said before, I think the switch to configmap will negate this.

gabemontero:
ok ... pushed the change to use configmap in the "error out on changes to arch/distro" commit

anything not related to that is in the ben 1 commit 

@bparees ^^

pkg/stub/handler.go
```diff
@@ -104,6 +107,99 @@ type Handler struct {
...
+		return nil
+	}
+	cm, err := h.configmapclientwrapper.Get(h.namespace, v1alpha1.SamplesResourceName)
+	if cm == nil || err != nil {
+		return h.processError(srcfg, v1alpha1.SamplesUpdateFailed, err, "problem getting configmap: %v")
+	}
+
+	switch srcfg.Spec.InstallType {
+	case "":
+		fallthrough
```
bparees: what's the intent here?

pkg/stub/handler.go
```diff
@@ -104,6 +107,99 @@ type Handler struct {
...
+	case v1alpha1.CentosSamplesDistribution:
+		installtype, ok := cm.Data[installtypekey]
+		if !ok || installtype != string(v1alpha1.CentosSamplesDistribution) {
+			err = fmt.Errorf("cannot change installtype from %s to %s", installtype, srcfg.Spec.InstallType)
+			return h.processError(srcfg, v1alpha1.SamplesUpdateFailed, err, "%v")
```
bparees: what status conditions will this set?  i think it's going to set SamplesExist==false which is not (necessarily) true.

The samples still exist (we didn't delete anything that was previously created).


Same for the other places you call this.

We need to distinguish between "your current spec is being rejected" and "the current state of what the operator has done"

pkg/stub/handler.go
```diff
@@ -104,6 +107,99 @@ type Handler struct {
 	mutex *sync.Mutex
 }
 
+func (h *Handler) StatusUpdate(condition v1alpha1.SamplesResourceConditionType, srcfg *v1alpha1.SamplesResource) error {
+	if condition == v1alpha1.SamplesExist {
+		cm, err := h.configmapclientwrapper.Get(h.namespace, v1alpha1.SamplesResourceName)
+		if err != nil && !kerrors.IsNotFound(err) {
+			return h.processError(srcfg, v1alpha1.SamplesUpdateFailed, err, "unable to get configm map: %v")
```
bparees: s/configm/config/

pkg/stub/handler.go
gabemontero: safety precaution ... in case this method is ever called with InstallType empty, we employ the defaulting convention we have for an empty samplesresource

pkg/stub/handler.go
gabemontero: yeah I debated that ... so to be clear, to me, that says a new condition ... an error one, perhaps called "ValidationSucceeded" or "ValidationFailed"

please confirm

pkg/stub/handler.go
bparees: `case "",v1alpha1.CentosSamplesDistribution:` is a cleaner way to write this imho

fallthrough makes more sense when you want to do some logic in this case statement *and* execute the logic in the next case statement.

gabemontero:
@bparees latest round pushed in "ben 2"

pkg/stub/handler_test.go
```diff
@@ -555,7 +555,7 @@ func validate(succeed bool, err error, errstr string, sr *v1alpha1.SamplesResour
 			case v1alpha1.ImportCredentialsExist:
 				fallthrough
 
-			case v1alpha1.SamplesExist:
+			case v1alpha1.SamplesExist, v1alpha1.ConfigurationValid:
```
bparees: you can bring the fallthrough case above down to this line too.

gabemontero: yep part of next push

gabemontero:
fallthrough removed

pkg/stub/handler.go
```diff
@@ -460,31 +561,36 @@ func (h *Handler) processFiles(dir string, files []os.FileInfo, opcfg *v1alpha1.
...
+			replaced := false
+			for _, old := range oldies {
+				if strings.HasPrefix(imagestream.Spec.DockerImageRepository, old) {
+					imagestream.Spec.DockerImageRepository = strings.Replace(imagestream.Spec.DockerImageRepository, old, opcfg.Spec.SamplesRegistry, 1)
+					replaced = true
+					break
+				}
+			}
+			if !replaced {
+				imagestream.Spec.DockerImageRepository = opcfg.Spec.SamplesRegistry + "/" + imagestream.Spec.DockerImageRepository
```
bparees: i'm confused by this.  

if my DockerImageRepository value is "myprivateregistry/myrepo/myimage"

you're going to replace it with "samplesregistry/myprivateregistry/myrepo/myimage"?

what's the intent here?

Same question about the logic for tags below.

pkg/stub/handler.go
```diff
@@ -104,6 +107,99 @@ type Handler struct {
...
+	if cm == nil || err != nil {
+		// just return error to sdk for retry
+		return err
+	}
+
+	switch srcfg.Spec.InstallType {
+	case "", v1alpha1.CentosSamplesDistribution:
+		installtype, ok := cm.Data[installtypekey]
+		if !ok || installtype != string(v1alpha1.CentosSamplesDistribution) {
+			err = fmt.Errorf("cannot change installtype from %s to %s", installtype, srcfg.Spec.InstallType)
```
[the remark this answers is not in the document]

pkg/stub/handler.go
```diff
@@ -104,6 +107,99 @@ type Handler struct {
...
+			return h.processError(srcfg, v1alpha1.InstallTypeChangeDetected, err, "%v")
+		}
+	case v1alpha1.RHELSamplesDistribution:
+		installtype, ok := cm.Data[installtypekey]
+		if !ok || installtype != string(v1alpha1.RHELSamplesDistribution) {
+			err = fmt.Errorf("cannot change installtype from %s to %s", installtype, srcfg.Spec.InstallType)
```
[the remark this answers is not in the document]

pkg/stub/handler.go
gabemontero: I was trying to distinguish between myprivateregistry/myrepo/myimage and myreop/myimage

Though at one point I had code to count the "/" characters as a way to distinguish between the two and that got lost in the shuffle

Now, that said, we are calling this function on the imagestream read from the file

If we are always confident that content from openshift/library will have myregistry/myrepo/myimage, then I'll forgo fixing my attempt to address the 2 and 1 "/" cases, and just simplify

thoughts?

pkg/stub/handler.go
```diff
@@ -432,32 +435,22 @@ func (h *Handler) buildSkipFilters(opcfg *v1alpha1.SamplesResource) {
...
-	case v1alpha1.SecretUpdateFailed:
-		status = opcfg.Condition(v1alpha1.ImportCredentialsExist)
-	case v1alpha1.SamplesUpdateFailed:
-		status = opcfg.Condition(v1alpha1.SamplesExist)
-	case v1alpha1.InstallTypeChangeDetected, v1alpha1.ArchitectureChangeDetected:
-		status = opcfg.Condition(v1alpha1.ConfigurationValid)
-	default:
-		return err
-	}
+	status := opcfg.Condition(ctype)
```
bparees: i think there is a bug here.

If opcfg does not contain the condition "ctype", Condition() manufactures one and returns it.

In the case that one was manufactured, there's no guarantee you ultimately proceed to call sdkwrapper.Update() to actually update the resource w/ the new status that got manufactured.

pkg/stub/handler.go
```diff
@@ -559,38 +552,47 @@ func (h *Handler) processFiles(dir string, files []os.FileInfo, opcfg *v1alpha1.
 	return nil
 }
 
+func (h *Handler) coreUpdateDockerPullSpec(oldreg, newreg string, oldies []string) string {
+	// see if the imagestream on file (i.e. the openshift/library content) is
+	// of the form "reg/repo/img" or "repo/img"
```
bparees: since our centos imagestreams probably are of the form "openshift/myimage" (ie no docker.io prefix) i think this logic is necessary/good.  

it's also nicely encapsulated now, so +1 from me.

pkg/stub/handler.go
```diff
@@ -559,38 +552,47 @@ func (h *Handler) processFiles(dir string, files []os.FileInfo, opcfg *v1alpha1.
...
+	}
+	if !replaced {
+		if !hasRegistry {
+			oldreg = newreg + "/" + oldreg
+		} else {
+			// the content from openshift/library has something odd in in ... replace the registry piece
+			parts := strings.Split(oldreg, "/")
+			oldreg = newreg + "/" + parts[1] + "/" + parts[2]
+		}
+	}
```
bparees: I would do this logic slightly differently. I think it's a mistake to call strings.Replace() on a value that we previously concluded does not have a registry prefix.  (it's unlikely to replace anything, but it could).

Should probably do:

```
if hasRegistry {
   // do search+replace
} else {
  // prepend the value w/ the replacement registry value
}
```

pkg/stub/handler.go
gabemontero: ok cool thx

pkg/stub/handler.go
gabemontero: I changed `Condition(ctype)` to return a condition with the Status field set to an empty string to account for this.

Callers should be passing in one of the Condition constants.  None of those are empty string.  They would have to pass in `ConditionStatus("")`  to cause a case where one that was manufactured did not lead to a resource update.

Do you want me to account for that case?

pkg/stub/handler.go
gabemontero: part of next push ... chose to still include the openshift/library content hasRegistry, but not one of our presumed constants, piece

pkg/stub/handler.go
bparees: i guess it's sufficient.

pkg/stub/handler.go
```diff
@@ -104,6 +107,102 @@ type Handler struct {
...
+		wantsppc != hasppc {
+		err = fmt.Errorf("cannot change architectures from %#v to %#v", cm.Data, srcfg.Spec.Architectures)
+		return h.processError(srcfg, v1alpha1.ConfigurationValid, corev1.ConditionFalse, err, "%v")
+	}
+
+	return h.GoodConditionUpdate(srcfg, corev1.ConditionTrue, v1alpha1.ConfigurationValid)
+}
+
 func (h *Handler) GoodConditionUpdate(srcfg *v1alpha1.SamplesResource, newStatus corev1.ConditionStatus, condition v1alpha1.SamplesResourceConditionType) error {
 	status := srcfg.Condition(condition)
```
bparees: nit but status should probably be named condition and condition should probably be named conditionType

bparees:
one nit, squash and lgtm

PR #16, opened 2018-09-25
fix configmap creation; add bad post create cfg change test cases

gabemontero:
@bparees ptal

### DIFF
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -114,6 +114,10 @@ func (h *Handler) StatusUpdate(condition v1alpha1.SamplesResourceConditionType, 
 			// just return error to sdk for retry
 			return err
 		}
+		if kerrors.IsNotFound(err) {
+			// 4.0 testing showed that we were getting empty ConfigMaps
+			cm = nil
+		}
 		if cm != nil {
 			// just return ... we only update the config map once
 			return nil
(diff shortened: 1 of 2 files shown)


pkg/stub/handler.go
```diff
@@ -154,7 +158,7 @@ func (h *Handler) SpecValidation(srcfg *v1alpha1.SamplesResource) error {
 		return nil
 	}
 	cm, err := h.configmapclientwrapper.Get(h.namespace, v1alpha1.SamplesResourceName)
-	if cm == nil || err != nil {
+	if err != nil && !kerrors.IsNotFound(err) {
```
bparees: if it is an IsNotFound err, i think you need to make sure cm gets set to a non-nil value.  I wouldn't count on get always returning an empty one (i find it weird that it ever does)

bparees: also i'm not sure what you expect to do if it is an IsNotFoundErr?

pkg/stub/handler.go
```diff
@@ -114,6 +114,10 @@ func (h *Handler) StatusUpdate(condition v1alpha1.SamplesResourceConditionType,
 			// just return error to sdk for retry
 			return err
 		}
+		if kerrors.IsNotFound(err) {
+			// 4.0 testing showed that we were getting emtpy ConfigMaps
```
bparees: typo in empty

pkg/stub/handler.go
gabemontero: Yeah need to error out with a ConditionUnknown status for the ConfigurationValid condition ... presumably somebody deleted the configmap underneath us, or some such pre-existing problem

bparees:
lgtm

PR #20, opened 2018-09-27
add finalizer processing to ensure we see delete events

gabemontero:

### DIFF
--- a/pkg/apis/samplesoperator/v1alpha1/types.go
+++ b/pkg/apis/samplesoperator/v1alpha1/types.go
@@ -39,6 +39,18 @@ const (
 	SamplesRegistryCredentials = "samples-registry-credentials"
 	// SamplesResourceName is the name/identifier of the static, singleton operator employed for the samples.
 	SamplesResourceName = "openshift-samples"
+	// X86Architecture is the value used to specify the x86_64 hardware architecture
+	// in the Architectures array field.
+	X86Architecture = "x86_64"
+	// PPCArchitecture is the value used to specify the ppc64le hardware architecture
+	// in the Architectures array field.
+	PPCArchitecture = "ppc64le"
+	// SamplesResourceFinalizer is the text added to the SamplesResource.Finalizer field
+	// to enable finalizer processing.
+	SamplesResourceFinalizer = "samplesoperator.config.openshift.io/finalizer"
+	// SamplesResourceLabel is the key for a label added to all the imagestreams and templates
+	// in the openshift namespace that the SamplesResource is managing.
+	SamplesResourceLabel = "samplesoperator.config.openshift.io/managed"
 )
 
 type SamplesResourceSpec struct {
@@ -111,6 +123,9 @@ type SamplesResourceCondition struct {
 }
 
 func (s *SamplesResource) ConditionTrue(c SamplesResourceConditionType) bool {
+	if s.Status.Conditions == nil {
+		return false
+	}
 	for _, rc := range s.Status.Conditions {
 		if rc.Type == c && rc.Status == corev1.ConditionTrue {
 			return true
@@ -120,6 +135,9 @@ func (s *SamplesResource) ConditionTrue(c SamplesResourceConditionType) bool {
 }
 
 func (s *SamplesResource) ConditionUpdate(c *SamplesResourceCondition) {
+	if s.Status.Conditions == nil {
+		return
+	}
 	for i, ec := range s.Status.Conditions {
 		if ec.Type == c.Type {
 			s.Status.Conditions[i] = *c
@@ -129,9 +147,11 @@ func (s *SamplesResource) ConditionUpdate(c *SamplesResourceCondition) {
 }
 
 func (s *SamplesResource) Condition(c SamplesResourceConditionType) *SamplesResourceCondition {
-	for _, rc := range s.Status.Conditions {
-		if rc.Type == c {
-			return &rc
+	if s.Status.Conditions != nil {
+		for _, rc := range s.Status.Conditions {
+			if rc.Type == c {
+				return &rc
+			}
 		}
 	}
 	newCondition := SamplesResourceCondition{
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -19,10 +19,11 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 
-	kapis "k8s.io/apimachinery/pkg/apis/meta/v1"
-
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	kapis "k8s.io/apimachinery/pkg/apis/meta/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilnet "k8s.io/apimachinery/pkg/util/net"
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	restclient "k8s.io/client-go/rest"
 
@@ -37,8 +38,6 @@ const (
 	x86OCPContentRootDir   = "/opt/openshift/operator/ocp-x86_64"
 	x86OKDContentRootDir   = "/opt/openshift/operator/okd-x86_64"
 	ppc64OCPContentRootDir = "/opt/openshift/operator/ocp-ppc64le"
-	x86                    = "x86_64"
-	ppc                    = "ppc64le"
 	installtypekey         = "installtype"
 )
 
@@ -101,6 +100,8 @@ type Handler struct {
 	skippedTemplates    map[string]bool
 	skippedImagestreams map[string]bool
 
+	deleteInProgress bool
+
 	mutex *sync.Mutex
 }
 
@@ -128,14 +129,14 @@ func (h *Handler) StatusUpdate(condition v1alpha1.SamplesResourceConditionType, 
 			cm.Data[installtypekey] = string(srcfg.Spec.InstallType)
 		}
 		if len(srcfg.Spec.Architectures) == 0 {
-			cm.Data[x86] = x86
+			cm.Data[v1alpha1.X86Architecture] = v1alpha1.X86Architecture
 		} else {
 			for _, arch := range srcfg.Spec.Architectures {
 				switch arch {
-				case x86:
-					cm.Data[x86] = x86
-				case ppc:
-					cm.Data[ppc] = ppc
+				case v1alpha1.X86Architecture:
+					cm.Data[v1alpha1.X86Architecture] = v1alpha1.X86Architecture
+				case v1alpha1.PPCArchitecture:
+					cm.Data[v1alpha1.PPCArchitecture] = v1alpha1.PPCArchitecture
 				}
 			}
 		}
@@ -185,8 +186,8 @@ func (h *Handler) SpecValidation(srcfg *v1alpha1.SamplesResource) error {
 		return h.processError(srcfg, v1alpha1.ConfigurationValid, corev1.ConditionFalse, err, "%v")
 	}
 
-	_, hasx86 := cm.Data[x86]
-	_, hasppc := cm.Data[ppc]
+	_, hasx86 := cm.Data[v1alpha1.X86Architecture]
+	_, hasppc := cm.Data[v1alpha1.PPCArchitecture]
 
 	wantsx86 := false
 	wantsppc := false
@@ -195,9 +196,9 @@ func (h *Handler) SpecValidation(srcfg *v1alpha1.SamplesResource) error {
 	}
 	for _, arch := range srcfg.Spec.Architectures {
 		switch arch {
-		case x86:
+		case v1alpha1.X86Architecture:
 			wantsx86 = true
-		case ppc:
+		case v1alpha1.PPCArchitecture:
 			wantsppc = true
 		default:
 			err = fmt.Errorf("trying to change architecture, which is not allowed, but also specified an unsupported architecture %s", arch)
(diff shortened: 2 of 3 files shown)


pkg/stub/handler.go
```diff
@@ -372,10 +472,29 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
...
+		}
+		if srcfg.DeletionTimestamp != nil {
+			if h.NeedsFinalizing(srcfg) {
+				logrus.Println("Initiating finalizer processing for a SampleResource delete attempt")
+				h.RemoveFinalizer(srcfg)
+				err := h.GoodConditionUpdate(srcfg, corev1.ConditionFalse, v1alpha1.SamplesExist)
+				h.samplesResource = nil
+				timer := time.NewTimer(15 * time.Second)
+				go func() {
+					<-timer.C
```
bparees: what's the purpose of the 15s delay?

pkg/stub/handler.go
```diff
@@ -372,10 +472,29 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
...
-			logrus.Println("SampleResource deleted")
-			h.samplesResource = nil
-			return h.CreateDefaultResourceIfNeeded()
+			// possibly tell poller to stop
+			logrus.Info("A previous delete attempt has been successfully completed")
+			return nil
+		}
+		if srcfg.DeletionTimestamp != nil {
+			if h.NeedsFinalizing(srcfg) {
+				logrus.Println("Initiating finalizer processing for a SampleResource delete attempt")
```
bparees: strictly speaking, finalizing should also remove the imagestreams, templates, and secrets we created in the openshift namespace.  (but only the ones we created/own).

gabemontero: By "created/own", does that mean existing ones we've updated ?

pkg/stub/handler.go
gabemontero: conservative delay

would you prefer we launch immediately?  Or some small interval?

pkg/stub/handler.go
```diff
@@ -372,10 +472,29 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
 		if srcfg.Name != v1alpha1.SamplesResourceName {
 			return nil
 		}
+
+		// pattern is 1) come in with delete timestample, event delete flag false
```
bparees: timestamp.

pkg/stub/handler.go
```diff
@@ -258,22 +302,73 @@ func (h *Handler) GoodConditionUpdate(srcfg *v1alpha1.SamplesResource, newStatus
...
+			if kerrors.IsNotFound(e) {
+				return true, nil
+			}
+			if err != nil && h.IsRetryableAPIError(err) {
+				return false, nil
+			}
+			if err != nil {
+				return false, err
+			}
+			// based on 4.0 testing, we've been seeing emtpy resources returned
```
bparees: empty.

pkg/stub/handler.go
```diff
@@ -258,22 +302,73 @@ func (h *Handler) GoodConditionUpdate(srcfg *v1alpha1.SamplesResource, newStatus
...
 	// where it will set h.sampleResource
 	// when it completely updates all imagestreams/templates/statuses
 	h.mutex.Lock()
 	defer h.mutex.Unlock()
-	if h.samplesResource == nil {
+	deleteInProgress := h.deleteInProgress
+
+	var err error
+	var srcfg *v1alpha1.SamplesResource
+	if deleteInProgress {
```
bparees: seems like you can just pass deleteInProgress as an arg to CreateDefaultResourceIfNeeded rather than shuffling it around as a field on the handler.

bparees: well, I take that back since you're using it in a variety of places.  something about this feels a little too loose to me though.  i'll keep thinking about it.

pkg/stub/handler.go
```diff
@@ -282,12 +377,17 @@ func (h *Handler) CreateDefaultResourceIfNeeded() error {
...
 			logrus.Println("got already exists error on create default, just going to forgo this time vs. retry")
 			// just return the raw error and initiate the sdk's requeue/ratelimiter setup
-			return err
 		}
-		return h.GoodConditionUpdate(srcfg, corev1.ConditionTrue, v1alpha1.SamplesExist)
+		h.deleteInProgress = false
+		return err
+	} else {
+		if deleteInProgress {
+			return h.processError(srcfg, v1alpha1.SamplesExist, corev1.ConditionUnknown, err, "issues waiting for delete to complete: %v")
```
bparees: you can do this return immediately after the wait.Poll call if err comes back !nil.

pkg/stub/handler.go
```diff
@@ -282,12 +377,17 @@ func (h *Handler) CreateDefaultResourceIfNeeded() error {
...
+		} else {
+			logrus.Println("SampleResource already received, not creating default")
```
bparees: and you should do this return on line 351 (if srccfg!=nil)

bparees:
```
if deleteinprogress {
  all logic related to waiting for delete, returning if we get an error waiting for deletion
  set h.SamplesResource=nil  // any reason not to do this here?
}
if h.samplesResource!=nil {
  return nil // we already have a samples resource
}

// logic to create default sampleresource.
```

pkg/stub/handler.go
```diff
@@ -372,10 +472,29 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
...
-			return h.CreateDefaultResourceIfNeeded()
+			// possibly tell poller to stop
+			logrus.Info("A previous delete attempt has been successfully completed")
+			return nil
+		}
+		if srcfg.DeletionTimestamp != nil {
+			if h.NeedsFinalizing(srcfg) {
+				logrus.Println("Initiating finalizer processing for a SampleResource delete attempt")
+				h.RemoveFinalizer(srcfg)
+				err := h.GoodConditionUpdate(srcfg, corev1.ConditionFalse, v1alpha1.SamplesExist)
```
bparees: i think this should be done before removing the finalizer.  You want to complete all the work you want to do related to the deletion of the object, before you remove the finalizer.

pkg/stub/handler.go
bparees: yes.  Anything that the samplesoperator knows how to create and which isn't in the Skipped list.

I think the best thing to do would be for the operator to put a label or annotation on the things it creates.  If someone later adds one of those items to the Skipped list, we should remove the label/annotation.

bparees: (if we weren't going cross-namespace, i'd suggest an ownerref and rely on GC)

pkg/stub/handler.go
bparees: I don't see any reason not to launch immediately...you were just looking to avoid unnecessary polling iterations i guess?

pkg/stub/handler.go
gabemontero: moved setting of h.SamplesResource to nil at that spot, worked fine, along with the rest of this

pkg/stub/handler.go
gabemontero: the `h.GoodConditionUpdate` is what persists the changes from `h.RemoveFinalizer` to etcd ... `h.RemoveFinalizer` just changes the local copy

bparees: ok

gabemontero:
pushed updates except for the delete imagestream/template content, label/annotate items we touch @bparees

gabemontero:
@bparees pushed labelling and deleting openshift imagestreams/templates/secret on SamplesResource delete

gabemontero:
Just realized ... we might want to copy the skipped lists from the old SamplesResource to the new default one created after deletion?

thoughts @bparees ? ... I could see pros and cons to both choices ....

I felt less inclined about the install type and archs, but I suppose it warrants at least a brief discussion to confirm/deny.

bparees:
> Just realized ... we might want to copy the skipped lists from the old SamplesResource to the new default one created after deletion?

no, if someone deletes the samplesresource we should create a new one in the same way we would have originally bootstrapped it.

gabemontero:
On Fri, Sep 28, 2018 at 11:02 AM Ben Parees <notifications@github.com>
wrote:

> Just realized ... we might want to copy the skipped lists from the old
> SamplesResource to the new default one created after deletion?
>
> no, if someone deletes the samplesresource we should create a new one in
> the same way we would have originally bootstrapped it.
>
OK cool the current commit does that.

pkg/stub/handler.go
```diff
@@ -258,36 +299,90 @@ func (h *Handler) GoodConditionUpdate(srcfg *v1alpha1.SamplesResource, newStatus
...
+	// GoodConditionUpdate, so we do everything in the create vs.
+	// create then immediately update
+	condition := srcfg.Condition(v1alpha1.SamplesExist)
+	now := kapis.Now()
+	condition.LastUpdateTime = now
+	condition.Status = corev1.ConditionTrue
+	condition.LastTransitionTime = now
+	srcfg.ConditionUpdate(condition)
+	logrus.Println("creating default SamplesResource")
+	err = h.sdkwrapper.Create(srcfg)
```
bparees: wait, so we're creating the initial SamplesResource with a condition.SamplesExist==true?

pkg/stub/handler.go
```diff
@@ -258,36 +299,90 @@ func (h *Handler) GoodConditionUpdate(srcfg *v1alpha1.SamplesResource, newStatus
...
+	if err != nil {
+		if !kerrors.IsAlreadyExists(err) {
+			return h.processError(srcfg, v1alpha1.SamplesExist, corev1.ConditionUnknown, err, "failed creating default resource: %v")
```
bparees: since we failed to create the SamplesResource, what is processError actually going to do when it tries to update a status field on SamplesResource?

pkg/stub/handler.go
```diff
@@ -372,10 +508,27 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
...
-			return h.CreateDefaultResourceIfNeeded()
+			// possibly tell poller to stop
+			logrus.Info("A previous delete attempt has been successfully completed")
+			return nil
+		}
+		if srcfg.DeletionTimestamp != nil {
+			if h.NeedsFinalizing(srcfg) {
+				logrus.Println("Initiating finalizer processing for a SampleResource delete attempt")
+				h.RemoveFinalizer(srcfg)
+				h.CleanUpOpenshiftNamespaceOnDelete(srcfg)
```
bparees: you also need to delete the configmap, right?  Either via this finalizer or via an ownerreference relationship to the samplesresource.

pkg/stub/handler.go
```diff
@@ -497,18 +651,23 @@ func (h *Handler) processFiles(dir string, files []os.FileInfo, opcfg *v1alpha1.
...
-					return h.processError(opcfg, v1alpha1.SamplesExist, corev1.ConditionUnknown, err, "unexpected imagestream get error: %v")
-				}
 
-				if kerrors.IsNotFound(err) {
-					// testing showed that we get an empty is vs. nil in this case
-					is = nil
+				if imagestream.Labels == nil {
+					imagestream.Labels = map[string]string{}
 				}
+				imagestream.Labels[v1alpha1.SamplesResourceLabel] = ""
```
bparees: should set it to something, e.g. "true"

pkg/stub/handler.go
```diff
@@ -533,16 +701,21 @@ func (h *Handler) processFiles(dir string, files []os.FileInfo, opcfg *v1alpha1.
...
+			}
 
-				if kerrors.IsNotFound(err) {
-					// testing showed that we get an empty is vs. nil in this case
-					t = nil
+			if _, tok := h.skippedTemplates[template.Name]; !tok {
+				if template.Labels == nil {
+					template.Labels = map[string]string{}
 				}
+				template.Labels[v1alpha1.SamplesResourceLabel] = ""
```
bparees: set to true

pkg/apis/samplesoperator/v1alpha1/types.go
```diff
@@ -39,6 +39,18 @@ const (
...
+	X86Architecture = "x86_64"
+	// PPCArchitecture is the value used to specify the ppc64le hardware architecture
+	// in the Architectures array field.
+	PPCArchitecture = "ppc64le"
+	// SamplesResourceFinalizer is the text added to the SamplesResource.Finalizer field
+	// to enable finalizer processing.
+	SamplesResourceFinalizer = "samplesoperator.config.openshift.io/finalizer"
+	// SamplesResourceLabel is the key for a label added to all the imagestreams and templates
+	// in the openshift namespace that the SamplesResource is managing.
+	SamplesResourceLabel = "samplesoperator.config.openshift.io/managed-label"
```
bparees: "samplesoperator.config.openshift.io/managed"

pkg/stub/handler.go
bparees: the label not being set, and the label being set to false(or anything that isn't "true"), would both have the same meaning for the operator, effectively.

Setting it to true allows us to set it to false in the future if we want to (or just remove it as you're doing today, which is also fine).

So i think i'd rather see it explicitly set to false (And the selector to select on =true, per your (c)) so we don't paint ourselves into a corner w/ respect to how the label is used in the future.

gabemontero:
updates pushed @bparees

pkg/stub/handler.go
```diff
@@ -128,17 +129,24 @@ func (h *Handler) StatusUpdate(condition v1alpha1.SamplesResourceConditionType,
...
 				}
 			}
 		}
+		or := metav1.OwnerReference{
+			APIVersion: "samplesoperator.config.openshift.io/v1alpha1",
+			Kind:       "SamplesResource",
+			Name:       v1alpha1.SamplesResourceName,
+			UID:        srcfg.UID,
+		}
+		cm.OwnerReferences = append(cm.OwnerReferences, or)
```
bparees: so one caveat about relying on ownerreferences that might push us to handle the deletion explicitly in the finalizer instead:

Depending how the samplesresource is deleted(the deleteoptions passed during the delete request), the configmap may or may not be deleted before the samplesresource is deleted.  So if you really want to ensure the old configmap is gone before the samplesresource gets deleted/recreated(which i think we need to), I think you need to handle it explicitly in your finalizer instead.

Sorry I didn't think about that when i suggested using GC.

gabemontero:
push to explicitly delete pushed @bparees

Issue #21, opened 2018-10-01
status/processing of changing installtype is incorrect


bparees:
1) starting from a default CR w/ installtype: centos

2) change the installtype to rhel8, status correctly reports:
```
 message: trying to change installtype, which is not allowed, but also specificed
      an unsupported installtype rhel8
```
3) change the installtype to rhel.  Status message does not change, I would expect it to just say "changing the install type is not allowed", it shouldn't be complaining about "rhel8" anymore.

4) change the installtype to "centos".  Status message does not change.  I would expect it to be back to a valid/non-error state.

@gabemontero

bparees:
(not surprisingly, the architecture field behaves the same way)

bparees:
looks like "ConfigurationValid" is true in (4) though.  So just the message isn't being updated.

PR #22, opened 2018-10-01
proper toggling of error messages with ConfigurationValid condition

gabemontero:
Fixes https://github.com/openshift/cluster-samples-operator/issues/21

Looks good with live testing in addition to unit test

### DIFF
--- a/pkg/stub/handler_test.go
+++ b/pkg/stub/handler_test.go
@@ -128,6 +128,27 @@ func TestWithBadArch(t *testing.T) {
 	validate(false, err, "architecture bad unsupported", sr, []v1alpha1.SamplesResourceConditionType{v1alpha1.SamplesExist}, []corev1.ConditionStatus{corev1.ConditionUnknown}, t)
 }
 
+func TestConfigurationValidCondition(t *testing.T) {
+	h, sr, event := setup()
+	err := h.Handle(nil, event)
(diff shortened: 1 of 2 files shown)


pkg/stub/handler_test.go
```diff
@@ -128,6 +128,27 @@ func TestWithBadArch(t *testing.T) {
...
 
+func TestConfigurationValidCondition(t *testing.T) {
+	h, sr, event := setup()
+	err := h.Handle(nil, event)
+	validate(true, err, "", sr, []v1alpha1.SamplesResourceConditionType{v1alpha1.SamplesExist}, []corev1.ConditionStatus{corev1.ConditionTrue}, t)
+	_, sr, event = setup()
+	sr.Spec.InstallType = "rhel8"
+	sr.ResourceVersion = "2"
+	err = h.Handle(nil, event)
+	validate(false, err, "trying to change installtype, which is not allowed, but also specificed", sr, []v1alpha1.SamplesResourceConditionType{v1alpha1.ConfigurationValid}, []corev1.ConditionStatus{corev1.ConditionFalse}, t)
```
bparees: s/specificed/specified/

bparees: (hopefully that would have failed your test..)

bparees:
nit, otherwise lgtm

pkg/stub/handler_test.go
gabemontero: it was misspelled in the src code too ... adjusting in both places

Issue #24, opened 2018-10-04
report error if registry is registry.redhat.io and no secret provided


bparees:
If the configuration is to install rhel imagestreams, and no override registry is provided (or the override registry is registry.redhat.io), and there is no imageimport secret available, we should report an error and refuse to proceed.

Otherwise we're going to create the imagestreams and they're going to fail to import and then the user has to manually re-import them after creating the secret.

PR #27, opened 2018-10-05
check for credential before process rhel install with default registry

gabemontero:
Fixes https://github.com/openshift/cluster-samples-operator/issues/24

### DIFF
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -555,6 +555,17 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
 			h.manageDockerCfgSecret(false, srcfg, h.registrySecret)
 		}
 
+		// if trying to do rhel to the default registry.redhat.io registry requires the secret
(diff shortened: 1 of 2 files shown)


bparees:
/lgtm

at some point i'd like to better understand the reconciliation loop (ie when we go through processError here, when do we next try?  is that driven by the operator sdk requeue/retry loop?  I'd hate for us to have to wait 10 mins (or whatever our relist/resync interval is) before attempting to create the samples again just because a race condition caused us to not see the credentials initially).

gabemontero:
The update of the bad status seemed to be initiating a new event @bparees independent of our current 10 minute resync interval.

However I'm not detecting a rapid fire, CPU burning hot loop when the error condition occurs.

I'll be constructing a new change to avoid the hot loop.  In doing so, I'll add a little more sophistication (perhaps using the mutex to coordinate concurrent secret/sampleresource update, probably along with a tweak to our event caching) to avoid the situation your are concerned about.

PR #28, opened 2018-10-08
protect against hotloop for rhel lack of cred error; add some mutex u…

gabemontero:
…sage for concurrent secret/samplesresoure/default activity

Follow up from https://github.com/openshift/cluster-samples-operator/pull/27

Reminder from snippet that PR.  Updates  to the SampleResource status are generating SamplesResource event.

So the secret getting successfully processed resulting in adding the ImporCredentials condition on the SamplesResource, leads to a cycle of SamplesResource event processing that will complete the original rhel processing desired.

### DIFF
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -81,6 +81,14 @@ type Handler struct {
 	samplesResource *v1alpha1.SamplesResource
 	registrySecret  *corev1.Secret
 
+	// we maintain a separate SamplesResource cache
+	// entry when this is a default rhel install and we
+	// are waiting for the credential since we are
+	// in an erorr state; our current approach is
+	// to only update the standard samplesResource
+	// cache entry for valid / non-error states
+	waitingForCredential *v1alpha1.SamplesResource
+
 	restconfig  *restclient.Config
 	tempclient  *templatev1client.TemplateV1Client
 	imageclient *imagev1client.ImageV1Client
@@ -320,6 +328,10 @@ func (h *Handler) CreateDefaultResourceIfNeeded() error {
 	// when it completely updates all imagestreams/templates/statuses
 	h.mutex.Lock()
 	defer h.mutex.Unlock()
+	if h.waitingForCredential != nil {
+		return nil
+	}
+
 	deleteInProgress := h.deleteInProgress
 
 	var err error
@@ -381,7 +393,8 @@ func (h *Handler) manageDockerCfgSecret(deleted bool, samplesResource *v1alpha1.
 	if secret.Name != v1alpha1.SamplesRegistryCredentials {
 		return nil
 	}
-	if h.samplesResource == nil {
+
+	if h.samplesResource == nil && h.waitingForCredential == nil {
 		if !deleted {
 			h.registrySecret = secret
 		}
@@ -438,7 +451,7 @@ func (h *Handler) manageDockerCfgSecret(deleted bool, samplesResource *v1alpha1.
 	}
 
 	h.registrySecret = secret
-	h.samplesResource = samplesResource
+	h.waitingForCredential = nil
 
 	return nil
 }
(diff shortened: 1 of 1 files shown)


pkg/stub/handler.go
```diff
@@ -562,8 +583,14 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
...
 			!srcfg.ConditionTrue(v1alpha1.ImportCredentialsExist) &&
 			h.registrySecret == nil {
-			err := fmt.Errorf("Cannot create rhel imagestreams to registry.redhat.io without the credentials being available: %#v", srcfg)
-			return h.processError(srcfg, v1alpha1.SamplesExist, corev1.ConditionFalse, err, "%v")
+			if h.waitingForCredential == nil {
+				h.waitingForCredential = srcfg
+				err := fmt.Errorf("Cannot create rhel imagestreams to registry.redhat.io without the credentials being available: %#v", srcfg)
+				return h.processError(srcfg, v1alpha1.SamplesExist, corev1.ConditionFalse, err, "%v")
+			} else {
+				// already recorded need, simply wait for credential to appear
```
bparees: don't we always want to update h.waitingForCredential?

If I do the following sequence:

1) create the samplesresource (such that it needs a secret)
  - we cache the samplesresource and report an error
2) update the samplesresource (such that it still needs a secret, but perhaps i update the skiplist)
 - currently we ignore that update entirely(?)
3) i create the missing secret
  - we create the samples from the cached value from (1), not the current value from (2).

(I haven't walked the logic carefully so perhaps i'm missing something in the flow where we get the updated samplesresource before doing the samples create)

gabemontero: Interesting points  ... taking them on not in sequential order, but by the importance I took for each of them ... high level summary, the answer to your question is yes, but not for the reasons you noted ... at least IMO / AFAIK

- we never create the samples from the cached SamplesResource

neither the traditional `samplesResource` cache, or the new `waitingForCredentials` cache

 ... at most, at most,  cached SamplesResource drives status condition updates of a not entirely ready variety

-  So to build upon your sequence
    1) SR event comes in for rhel, no cred, so cache in `waitingForCredentials`, but event otherwise ignored
    2) SR event updates skipped, still no cred, event still ignored
    3) secret created ... `waitingForCredentials` used to create the ImportCredentialsExists condition
   4) that `ImportCredentialsExists` update leads to a new SR event, which has the updates to the skipped list
    5) now that the cred exists, process the samples
    6) assuming successful, `samplesRegistry` cache item is updated.

So ... why do we still need to update waitingForCredentials then :-) ?

It is the config map for spec validation   ....  suppose they updated the arch's instead of the skipped list before we laid down the initial set of samples ... we want to work off the latest config there.

bparees: yeah, so I think you're saying if they choose to update the samplesresource such that creds are no longer needed, we would need to process/react to that and update status/proceed accordingly, which the code currently wouldn't do.

gabemontero: Actually I spoke too soon ... the status update of the config map won't happen on the cached value, but on the event ... but I'll update the cache anyway, doesn't hurt, as well as update the comment, to provide some of the detail I noted here, as there is complexity here

gabemontero: yeah since the sdk does not surface an UpdateStatus method, keeping the cache up to date with the most recent resource version is very helpful :-)

gabemontero:
updates pushed @bparees

pkg/stub/handler.go
```diff
@@ -438,7 +451,7 @@ func (h *Handler) manageDockerCfgSecret(deleted bool, samplesResource *v1alpha1.
 	}
 
 	h.registrySecret = secret
-	h.samplesResource = samplesResource
```
bparees: why is this  no longer needed?  or was it already dead code?

pkg/stub/handler.go
```diff
@@ -81,6 +81,14 @@ type Handler struct {
 	samplesResource *v1alpha1.SamplesResource
 	registrySecret  *corev1.Secret
 
+	// we maintain a separate SamplesResource cache
+	// entry when this is a default rhel install and we
+	// are waiting for the credential since we are
+	// in an erorr state; our current approach is
+	// to only update the standard samplesResource
+	// cache entry for valid / non-error states
+	waitingForCredential *v1alpha1.SamplesResource
```
bparees: is there a reason this is better than just setting a "waitingForCredential" boolean value, and holding the pending samplesResource in the existing "samplesResource" field?

Or setting a waitingForCredential boolean and when the credential shows up, simply fetching the current SamplesResource and processing it (after having processed the credential of course)?

pkg/stub/handler.go
gabemontero: it was not needed previously ... and in general I want to minimize where we set `h.samplesResource`

pkg/stub/handler.go
gabemontero: the complete separation of concerns was cleaner and easier to keep separate in my mind this way 

it wanted to keep samplesResource as only ever for "good" samplesresource

PR #30, opened 2018-10-15
run the samples operator on the master node

bparees:


bparees:
@deads2k ptal, need confirmation this is the right collection of settings (will do the same for image-registry)

gabemontero:
will wait on @deads2k before adding looks good to merge label

bparees:
this seems to have worked for the registry, so lgtm'ing.

Issue #31, opened 2018-10-15
move samples custom resource to be cluster scoped


bparees:
Since we only want one set of samples per cluster anyway and because finalizers deadlock themselves when the finalizer and the resource being finalized run in the same namespace, we need to switch the samples resource CRD to be cluster scoped instead of namespace scoped.

/cc @deads2k 

/assign @gabemontero

gabemontero:
So looking at various existing operators for comparison.

My interpretation on what needs to be done:
- of course in the crd change scope from Namespaces to Cluster 
- create a clusterrolebinding file, where the subject is a new SA defined in the sample operator namespace
- new file to create said SA which has the clusterrolebinding
- add the SA ref to the operator's deployment definition, at the bottom, after the container def

the cluster-ingress-operator has both a cluster role / cluster role binding and a role / role binding ... so perhaps keep the existing rbac file .... but its crd is still Namespaced

A little worried if the SDK handles this seamlessly ... none of the current cluster scoped CRDs I have found are using it.... should find out quickly assuming I've set up the yamls correctly.

gabemontero:
samples operator is not current defining an SA ... the role is being added to the `default` SA

do you want me to just add the cluster role binding to the `default` SA ?

bparees:
no it's probably best to get aligned w/ the image-registry operator and start creating an SA and using it to run the operator.

gabemontero:
no issues on the sdk front when running locally ... pr up

PR #32, opened 2018-10-16
convert to crd cluster role, add specific SA

gabemontero:
Fixes https://github.com/openshift/cluster-samples-operator/issues/31


deploy/02-rbac.yaml
```diff
@@ -1,5 +1,5 @@
-kind: Role
-apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRole
```
bparees: i think you need two roles (one a clusterrole for the CRD, and the other this namespace scoped role for the various resources the operator needs to modify in its own namespace)


we don't want to give the samples operator cluster-wide access to all these resources.

deploy/02-sa.yaml
```diff
@@ -0,0 +1,6 @@
+---
```
bparees: fix the name on this, we don't want any overlap on the numbering.  The numbering should control the order these resources are created in  (so you also probably want to create the SA before the rolebindings, though strictly speaking i don't think it matters)

gabemontero: yeah in comparing with some other repos I saw where several cases where steps that could happen in either order had the same number

i.e. the SA and the rolebinding where as you noted they can occur in either order (my manual testing saw that)

but sure, I can be explicit, and will put the SA before the rolebinding

deploy/02-rbac.yaml
gabemontero: yeah I was on that path with my first comment, then somehow got turned around ... updating ...

gabemontero:
updates pushed @bparees

gabemontero:
ah ... forgot dockerfile change on filename changes

deploy/03-rbac.yaml
```diff
@@ -0,0 +1,72 @@
...
+  - apps
+  resources:
+  - deployments
+  - daemonsets
+  - replicasets
+  - statefulsets
+  verbs:
+  - "*"
+
+kind: ClusterRole
```
bparees: missing a --- above this.

deploy/03-rbac.yaml
```diff
@@ -0,0 +1,72 @@
...
+  name: cluster-samples-operator
+subjects:
+- kind: ServiceAccount
+  name: cluster-samples-operator
+  namespace: openshift-cluster-samples-operator
+roleRef:
+  kind: Role
+  name: cluster-samples-operator
+
+kind: ClusterRoleBinding
```
bparees: missing a --- above this

deploy/03-rbac.yaml
gabemontero: ah ... thought those were optional

bparees: you're combining multiple resources into a single file, so you need to split them.

deploy/03-rbac.yaml
```diff
@@ -0,0 +1,72 @@
+kind: Role
```
bparees: for consistency should probably put --- up here too, though i guess it's implicit.

gabemontero: yeah it is / wasn't there before

gabemontero:
dockerfile and `---` update pushed

bparees:
@gabemontero this change doesn't require any code updates to watch the correct SamplesResource?

gabemontero:
@bparees it did not previously per https://github.com/openshift/cluster-samples-operator/issues/31#issuecomment-430380906

gabemontero:
updates per face to face pushed @bparees

bparees:
@gabemontero please open an issue on openshift/release for that flake, we can't afford to have that job start out flaky, it's barely doing anything.

gabemontero:
@bparees according to Trevor King on #forum-testplatform it probably resulted from a clean up script he was running  .... if it persists I'll open

PR #33, opened 2018-10-16
set cluster operator status

bparees:
should note most of this code was borrowed from @legionus's PR on image-registry operator.


pkg/apis/samplesoperator/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -13,8 +13,8 @@ func (in *SamplesResource) DeepCopyInto(out *SamplesResource) {
 	*out = *in
 	out.TypeMeta = in.TypeMeta
 	in.ObjectMeta.DeepCopyInto(&out.ObjectMeta)
-	out.Spec = in.Spec
-	out.Status = in.Status
+	in.Spec.DeepCopyInto(&out.Spec)
+	in.Status.DeepCopyInto(&out.Status)
```
bparees: @gabemontero i'm not sure why this is changing...  i just did a full generate, so either:

1) you haven't done it in a while and i have a newer sdk-generator than you
2) my sdk-generator is newer than yours

pkg/operatorstatus/operatorstatus.go
```diff
@@ -0,0 +1,124 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/util/retry"
+
+	osapi "github.com/openshift/cluster-version-operator/pkg/apis/operatorstatus.openshift.io/v1"
+	"github.com/operator-framework/operator-sdk/pkg/sdk"
+
+	"github.com/openshift/cluster-samples-operator/pkg/apis/samplesoperator/v1alpha1"
+)
+
+func UpdateOperatorStatus(sampleResource *v1alpha1.SamplesResource) error {
```
bparees: @gabemontero i expect this logic to evolve over time, but hopefully it's a sufficient stake in the ground for now.

pkg/apis/samplesoperator/v1alpha1/zz_generated.deepcopy.go
gabemontero: could be either or both ... I have not done any changes in the vendor or generated code spaces since initially seeding the repo for SDK and getting the vendoring of client-go sorted out with that whole `glide` vs. `dep` foray.

gabemontero:
Fyi the images job failed with an api.ci registry flake (unable to connect) building the image:

```
2018/10/16 22:43:36 Copying artifacts from release-latest into /logs/artifacts/release-latest
info: Using registry public hostname registry.svc.ci.openshift.org
info: Found 78 images in image stream
info: Manifests will be extracted to /tmp/release-image-2018-10-16T224302Z616505967
warning: Could not load current user information: user: unknown userid 1002020000
Unable to connect to the server: net/http: TLS handshake timeout
2018/10/16 22:43:36 Container release in pod release-latest failed, exit code 1, reason Error
```

gabemontero:
The unit test failures are related to being unable to get new type:

`failed to get resource client for (apiVersion:operatorstatus.openshift.io/v1, kind:ClusterOperator, ns:): failed to get resource type: failed to get the resource REST mapping for GroupVersionKind(operatorstatus.openshift.io/v1, Kind=ClusterOperator): no matches for kind "ClusterOperator" in version "operatorstatus.openshift.io/v1"`

We'll need to add the appropriate interface / abstraction layer so that we can fake out REST calls in the unit test.

pkg/stub/handler.go
```diff
@@ -114,6 +115,12 @@ type Handler struct {
 }
 
 func (h *Handler) StatusUpdate(condition v1alpha1.SamplesResourceConditionType, srcfg *v1alpha1.SamplesResource) error {
+
+	switch condition {
+	case v1alpha1.SamplesExist:
+
+	}
+
```
gabemontero: Was  this you deciding you'd rather have a switch vs. an if check, but you did not complete the conversion ?

pkg/stub/handler.go
```diff
@@ -527,6 +534,13 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
 			return nil
 		}
 
+		// Every time we see a change to the SamplesResource object, update the ClusterOperator status
+		// based on the current conditions of the SamplesResource.
+		err := operatorstatus.UpdateOperatorStatus(srcfg)
+		if err != nil {
+			return err
```
gabemontero: We should have a new Condition on the SamplesResource status that says whether we updated the CVO status or not, no?

pkg/operatorstatus/operatorstatus.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	if sampleResource.ConditionTrue(v1alpha1.ConfigurationValid) {
+		err = setOperatorStatus(sampleResource.Namespace, sampleResource.Name, osapi.OperatorFailing, osapi.ConditionFalse, "The samples operator configuration is invalid")
+	} else {
+		err = setOperatorStatus(sampleResource.Namespace, sampleResource.Name, osapi.OperatorFailing, osapi.ConditionTrue, "The samples operator configuraiton is valid")
+	}
+	if err != nil {
+		return err
+	}
```
gabemontero: Should add ImportCredentials as well IMO

Now, if we do decide to add an update CVO status condition on the SamplesResource, we would forgo setting that here.

pkg/stub/handler.go
bparees: i don't think that's a thing an end-user needs to care about.  for now i'm not inclined to add it.

pkg/stub/handler.go
bparees: other way around. I started out writing a switch.  Anyway it should have been removed, thanks.

bparees: (actually it was me debating where to put the logic)

pkg/operatorstatus/operatorstatus.go
bparees: Add it how/where?

My understanding of the samples operator flow is/was that SamplesExist can't be true until ImportCredentials is true, unless ImportCredentials is not needed(because we're not referencing registry.redhat.io)

So ImportCredentials is a secondary concern, all the CVO cares about is "has the operator done it's job" where "it's job" is "Create the samples".

In the fullness of time i'd expect this implementation to evolve to:

Check if the imagestreams have been created + imported successfully -> Available==true

but again, that's not directly dependent on looking at ImportCredentials.

gabemontero: as to where, I was thinking within the SamplesExists block because
it could provide some detail in the rhel case where we are not doing the import cause the credentials are not there yet (perhaps there was a problem trying to read in the credential that could be helpful)

but if we want to keep this coarse grained that is fine ... as much of anything, getting a lay of the land here

pkg/stub/handler.go
gabemontero: 10-4

pkg/stub/handler.go
gabemontero: ok

pkg/operatorstatus/operatorstatus.go
bparees: the status we promote up to the ClusterOperator resource is intended to be very course grained, not a duplication of the status we're putting on our own CR.  At least that's the direction i've gotten so far.

bparees:
> After any discussion on those comments, do you want / have the cycles to drive this to conclusion, or would you rather I pick up these changes into my own branch/PR and complete things?

happy to have you push it over the edge, i've got a few other things i need to go look at anyway.

pkg/operatorstatus/operatorstatus.go
gabemontero: 10-4

pkg/stub/handler.go
bparees: fixed

gabemontero:
OK I have your branch checked out in a local branch of mine now @bparees 

I'll start hacking on the unit tests and get a new PR up when I have them ready.

bparees:
replaced by https://github.com/openshift/cluster-samples-operator/pull/35

PR #36, opened 2018-10-18
create/update clusteroperator status object

bparees:


pkg/stub/handler.go
```diff
@@ -113,60 +116,67 @@ type Handler struct {
...
+	// are "valid" (the architecture name is known, the distribution name is known, etc)
+	// if that fails, we should immediately error out and set ConfigValid to false.
+
+	// only if the values being requested are valid, should we then proceed to check
+	// them against the previous values(if we've stored any previous values)
+
 	// if we have not had a valid SamplesResource processed, allow caller to try with
 	// the srcfg contents
-	if h.samplesResource == nil || !h.samplesResource.ConditionTrue(v1alpha1.SamplesExist) {
+	if !srcfg.ConditionTrue(v1alpha1.SamplesExist) {
```
bparees: this looked suspect to me.  i'd like to get rid of as much of this "h.samplesResource" caching logic as possible as it's difficult to reason about the state of that value when the operator has restarted.

`srcfg` represents the current state of the system (the event we just got), so use that to drive decision making about whether we need to compare with an old config or not.

pkg/stub/handler.go
```diff
@@ -113,60 +116,67 @@ type Handler struct {
...
+		logrus.Println("Spec is valid because samples are not created yet")
 		return nil
 	}
 	cm, err := h.configmapclientwrapper.Get(h.namespace, v1alpha1.SamplesResourceName)
 	if err != nil && !kerrors.IsNotFound(err) {
-		// just return error to sdk for retry
-		return err
+		err = fmt.Errorf("error retrieving previous configuration: %v", err)
+		return h.processError(srcfg, v1alpha1.ConfigurationValid, corev1.ConditionUnknown, err, "%v")
```
bparees: we tried to validate the configuration, we failed to do so, the condition is now unknown (even if it was previously explicitly valid or invalid).

pkg/stub/handler.go
```diff
@@ -554,17 +574,10 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
...
-			newVersion, _ := strconv.Atoi(srcfg.ResourceVersion)
-			if newVersion <= currVersion {
-				return nil
-			}
-
-			err := h.SpecValidation(srcfg)
-			if err != nil {
-				return err
-			}
+		err = h.SpecValidation(srcfg)
```
bparees: the operator-sdk Update() function mutates the object you pass in, updating it to the result of the update call.  This meant that when we would do something like:

Update(h.samplesResource), the cached value would get updated w/ the new resource version.

That meant when we then got back to this point based on the resulting update event, we'd ignore the update event.

So when we went through a flow of:

1) samples resource created
2) event handler see create event and instantiates samples and updates condition to samplesexist
3) event handler sees update event, checks resource version and ignores it (never calling SpecValidation and thus never setting ConfigValid to true)

i'm not entirely sure whether i broke something to cause that, or it was the state i was in, but regardless I think we should just do this every time we see the object.  I know i previously argued to avoid that, but i'm now leaning towards correctness/guarantees over performance, at least until proven otherwise.

bparees:
@gabemontero ok, at least in a quick smoketest this worked, ptal and also note the numerous // TODOs i've scattered throughout.

bparees:
i've made one of your unit tests unhappy because of my change to how spec validation behaves, working on it.

bparees:
unit tests fixed.  In many cases i changed them to carry the same SamplesResource through multiple invocations of `Handle` which i think is more realistic than the way the tests were operating (creating a new samplesresource but expecting the handler to maintain certain state/behavior).

Fundamentally the changes were needed because the flow is now:

1) SR is created
2) handler processes it, creates the samples and marks condition samples created, and stores the configuration
3) handler sees update to samples resource (the one that marked the condition samples created)
4) handler confirms the configuration is valid (against the saved config form (2) and applies the configvalid condition.  - it only checks the configuration if it sees the SamplesExist condition is already asserted, which is why we need to retain the samplesresource through multiple handle calls (or set SamplesExist on the new resource)

ideally i want to change this so (4) happens as part of (2) but right now GoodConditionUpdate() doesn't allow for updating 2 conditions in a single write to the SR.  Refactor for another day.

pkg/stub/handler.go
```diff
@@ -526,14 +531,18 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
...
+
 		if srcfg.Name != v1alpha1.SamplesResourceName || srcfg.Namespace != "" {
 			return nil
 		}
 
 		// Every time we see a change to the SamplesResource object, update the ClusterOperator status
 		// based on the current conditions of the SamplesResource.
 		err := h.cvowrapper.UpdateOperatorStatus(srcfg)
+		// TODO make sure an error here does not prevent us from doing other necessary processing
+		// of the samples resource object.
```
gabemontero: So help me reconcile this comment with the fact that you have an error is nil check and returning the error to the SDK if it exists ...

... is the answer to the TODO comment not to just log the error and continue processing (and perhaps attempt to update operator status at a later date)?

If not, I need help understanding what you are getting at here ....

pkg/stub/handler.go
```diff
@@ -116,61 +116,67 @@ type Handler struct {
...
 	}
-	return nil
+	_, err = h.configmapclientwrapper.Create(h.namespace, cm)
+	return err
 }
 
 func (h *Handler) SpecValidation(srcfg *v1alpha1.SamplesResource) error {
+	// TODO - the first thing this should do is check that all the config values
+	// are "valid" (the architecture name is known, the distribution name is known, etc)
+	// if that fails, we should immediately error out and set ConfigValid to false.
```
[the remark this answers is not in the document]
bparees: > if they specify empty strings for arch/distro, assuming we'll do the default, error out?

no, emptystring is fine.  but invalid values (ie a distribution or arch we don't understand) is not fine.

> at least for distro, the field is InstallType SamplesDistributionType ... so do you mean if they programmatically did SamplesDistributionType("foobar") ?

yes but of course for them it's not programmatic, it's just yaml editing of the resource (oc edit) so the types don't help them.  it's just a string and to them any value is "valid".  Nothing is going to block any valid string from getting set and you processing (or attempting to process) it.

> the arch is a string ... so sure, either add a new SamplesArchitectureType, or do explicit checks so that is it only the two valid

defining a type doesn't matter, for one thing that new "Type" is still just a string.  It's not like it's an enumeration of the valid values.  For another those types don't come into play when the user is editing yaml.  They're a programmatic convenience for your operator code and that's about it.
 
> so essentially just move those check here so the invalid config gets specified ... correct?

i believe so.  Essentially everytime we see a SamplesResource event, the first thing (well one of the first things) we should be doing is confirming the config it specifies is valid.  Valid in terms of standing on its own, and valid in terms of being compatible with the current applied config.

pkg/stub/handler.go
bparees: > So either here, or in a follow up PR, I think we are good for changing this to be a string, and remove any fears of a cached, full copy.

definitely a follow up.  Not 100% sure i understand the remaining use case (not sure why it would be a string vs a boolean?) but we can discuss that in the followup that reworks that logic/removes the cache field.

pkg/stub/handler.go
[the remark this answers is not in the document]
gabemontero: OK I will do that in my next follow up PR

pkg/stub/handler.go
gabemontero: OK, the second item for my follow up PR.

pkg/stub/handler.go
bparees: it could be that simple, but i wasn't sure if we'd want to do something more sophisticated (such as forcing the event to be requeued for retry, which is obviously what currently happens with the return err).  So i wanted to revisit that as we think about the general level-driven reconciliation pattern we want to get to here.

pkg/stub/handler.go
bparees: i'm still not convinced we should need to be capturing/using the last processed resource version at all, but again we can revisit that as we move forward.

pkg/stub/handler.go
gabemontero: It is the relist in particular I am concerned about wrt duplicate events.

And yes, the sdk.Update(...) and mutating / duplicate ... a bit more precisely, my experience was that I was *NOT* updating the spec, just the metadata for the keys, and the call to update to set the conditions was resulting in duplicate events ... so either
a) I missed a spot in the code, and my assertion is false ... in any event ,removing the "cached" version should help prove things out one way or the other
b) the sdk really needs an `UpdateStatus` like what I see in client-go, so it leaves the resource version alone

gabemontero: this will be a potential add on to the follow up PR

pkg/stub/handler.go
gabemontero: Do you want to reconcile this here, or in my follow up PR @bparees ?

gabemontero: ah missed your last comment before posting @bparees

gabemontero: For my follow up PR

pkg/stub/handler.go
gabemontero: ok sure thing ... my concern was that when the relist came in, if we did not ignore it, we would update the images streams again, albeit with the same spec, and I was afraid it would still be firing image change triggers everywhere

pkg/stub/handler.go
bparees: My understanding is that any change to the stored object will change the ResourceVersion value.  It's basically the etcd transaction identifier.  It has no understanding of whether you changed metadata, spec, or status.

So basically any time you update Status (ie setting a condition) you are going to get a new SampleResource event, with a new resourceversion, which your handler will need to process.  In general processing it should be a no-op (nothing has changed about the desired config), though since the Conditions changed, the ClusterOperator status update flow will likely be triggered to reflect the new Conditions into the ClusterOperator object.

bparees: (so again, I don't think you were seeing "duplicate events", i think you were seeing the update event which *appeared* to contain the same resourceversion as the one you had cached, because when you called sdk.Update() your cached object's resource version got updated to the new resource version that resulted from the update.  Then the update event comes in, also with that resource version that's in your cached resource)

gabemontero:
On Fri, Oct 19, 2018 at 2:45 PM Ben Parees <notifications@github.com> wrote:

> *@bparees* commented on this pull request.
>
> In pkg/stub/handler.go
> <https://github.com/openshift/cluster-samples-operator/pull/36#discussion_r226746480>
> :
>
> >  		return nil
>  	}
>  	cm, err := h.configmapclientwrapper.Get(h.namespace, v1alpha1.SamplesResourceName)
>  	if err != nil && !kerrors.IsNotFound(err) {
> -		// just return error to sdk for retry
> -		return err
> +		err = fmt.Errorf("error retrieving previous configuration: %v", err)
> +		return h.processError(srcfg, v1alpha1.ConfigurationValid, corev1.ConditionUnknown, err, "%v")
>
> My understanding is that any change to the stored object will change the
> ResourceVersion value. It's basically the etcd transaction identifier. It
> has no understanding of whether you changed metadata, spec, or status.
>
> So basically any time you update Status (ie setting a condition) you are
> going to get a new SampleResource event, with a new resourceversion, which
> your handler will need to process. In general processing it should be a
> no-op (nothing has changed about the desired config), though since the
> Conditions changed, the ClusterOperator status update flow will likely be
> triggered to reflect the new Conditions into the ClusterOperator object.
>
OK ... as part of removing the cache entry, I might update the config map
to list the  skipped lists, etc. in addition to the stuff that cannot
change,
and compare those before deciding to ignore relists or updates cause of
revision.

gabemontero:
On Fri, Oct 19, 2018 at 2:47 PM Ben Parees <notifications@github.com> wrote:

> *@bparees* commented on this pull request.
>
> In pkg/stub/handler.go
> <https://github.com/openshift/cluster-samples-operator/pull/36#discussion_r226746983>
> :
>
> >  		return nil
>  	}
>  	cm, err := h.configmapclientwrapper.Get(h.namespace, v1alpha1.SamplesResourceName)
>  	if err != nil && !kerrors.IsNotFound(err) {
> -		// just return error to sdk for retry
> -		return err
> +		err = fmt.Errorf("error retrieving previous configuration: %v", err)
> +		return h.processError(srcfg, v1alpha1.ConfigurationValid, corev1.ConditionUnknown, err, "%v")
>
> (so again, I don't think you were seeing "duplicate events", i think you
> were seeing the update event which *appeared* to contain the same
> resourceversion as the one you had cached, because when you called
> sdk.Update() your cached object's resource version got updated to the new
> resource version that resulted from the update. Then the update event comes
> in, also with that resource version that's in your cached resource)
>
I disagree in that I was seeing "duplicates" when I first started this, and
at that time, I was not updating status.

But bottom line we'll see when I start the next PR

bparees:
> OK ... as part of removing the cache entry, I might update the config map to list the skipped lists, etc. in addition to the stuff that cannot change,

makes sense.  essentially you should hold "what we've seen/processed before" somewhere, and whenever you are looking at the SamplesResource, determine anything we care about has changed from those held values and if so, take action.  Which i think is basically what you were trying to do w/ the resource cache and version comparison, but i think actually storing the values is safer, if only because it works better if the operator is restarted.

pkg/stub/handler.go
bparees: > my concern was that when the relist came in, if we did not ignore it, we would update the images streams again, albeit with the same spec, and I was afraid it would still be firing image change triggers everywhere

that should not be a problem.  we'd only fire the triggers if the import found a new sha.

now it's possible the "replace" action will trigger a re-import which could update a sha if there was a new image available (kinda like a scheduled import would) which would be undesirable behavior, but i think we need to think about that holistically (how the operator knows whether it needs to apply a new imagstream update or not.  After all, if we update the operator with new imagestreamtags for an existing imagestream, we absolutely *do* want to apply that change, even if nothing about SamplesResource itself has changed).  Again, item for future discussion.

PR #37, opened 2018-10-22
minor cluster operator fixes

gabemontero:
/assign bparees

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+REPO=openshift
+
 all: generate build build-image
 generate:
 	operator-sdk generate k8s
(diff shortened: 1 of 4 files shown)


gabemontero:
@bparees these changes include the verification we discussed in aos-devex IRC

Makefile
```diff
@@ -7,6 +9,10 @@ build-image:
 	# save some time setting up the docker build context by deleting this first.
 	rm -f cluster-samples-operator        
 	docker build -t docker.io/openshift/origin-cluster-samples-operator:latest .
+build-image-dev:
+	rm -f cluster-samples-operator
+	docker build -t docker.io/$(REPO)/origin-cluster-samples-operator:latest .
+	docker push docker.io/$(REPO)/origin-cluster-samples-operator:latest
```
bparees: i don't think this belongs here.

"make build-image && docker push foo" covers it, and this is going to fail by default for most people since they won't set REPO and can't push to the openshift repo.

also doing image pushing in a target named build is odd.

eventually we're probably going to want patterns for helping w/ these iterative workflows, but i don't think this is the one we want.

Updating the build-image target to use the new $(REPO) variable would be fine though.

bparees:
other than the makefile changes, lgtm

gabemontero:
makefile updated

PR #38, opened 2018-10-23
get rid of cached samplesresource; add cfg base duplicate event check

gabemontero:
a couple of the follow up items from the initial clusteroperator status PR @bparees and I sorted through last week

got motivated to do this now cause the reduction in imagestream/template update churn in my libvirt env really sped up the demo

### DIFF
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -39,7 +38,10 @@ const (
 	x86OCPContentRootDir   = "/opt/openshift/operator/ocp-x86_64"
 	x86OKDContentRootDir   = "/opt/openshift/operator/okd-x86_64"
 	ppc64OCPContentRootDir = "/opt/openshift/operator/ocp-ppc64le"
-	installtypekey         = "installtype"
+	installtypekey         = "keyForInstallTypeField"
+	regkey                 = "keyForSamplesRegistryField"
+	skippedstreamskey      = "keyForSkippedImageStreamsField"
+	skippedtempskey        = "keyForSkippedTemplatesField"
 )
 
 func NewHandler() sdk.Handler {
@@ -66,11 +68,7 @@ func NewHandler() sdk.Handler {
 	h.skippedTemplates = make(map[string]bool)
 
 	h.mutex = &sync.Mutex{}
-	timer := time.NewTimer(5 * time.Second)
-	go func() {
-		<-timer.C
-		h.CreateDefaultResourceIfNeeded()
-	}()
+	h.CreateDefaultResourceIfNeeded()
 
 	return &h
 }
@@ -80,17 +78,6 @@ type Handler struct {
 
 	sdkwrapper SDKWrapper
 	cvowrapper *operatorstatus.CVOOperatorStatusHandler
-
-	samplesResource *v1alpha1.SamplesResource
-	registrySecret  *corev1.Secret
-
-	// we maintain a separate SamplesResource cache
-	// entry when this is a default rhel install and we
-	// are waiting for the credential since we are
-	// in an erorr state; our current approach is
-	// to only update the standard samplesResource
-	// cache entry for valid / non-error states
-	waitingForCredential *v1alpha1.SamplesResource
 
 	restconfig  *restclient.Config
 	tempclient  *templatev1client.TemplateV1Client
@@ -116,6 +103,75 @@ type Handler struct {
 	mutex *sync.Mutex
 }
 
+func (h *Handler) VariableConfigChanged(srcfg *v1alpha1.SamplesResource, cm *corev1.ConfigMap) (bool, error) {
+	samplesRegistry, ok := cm.Data[regkey]
+	if !ok {
+		err := fmt.Errorf("could not find the registry setting in the config map %#v", cm.Data)
+		return false, h.processError(srcfg, v1alpha1.ConfigurationValid, corev1.ConditionUnknown, err, "%v")
+	}
+	if srcfg.Spec.SamplesRegistry != samplesRegistry {
+		logrus.Printf("SamplesRegistry changed from %s, processing %v", samplesRegistry, srcfg)
+		return true, nil
+	}
+
+	skippedStreams, ok := cm.Data[skippedstreamskey]
+	if !ok {
+		err := fmt.Errorf("could not find the skipped stream setting in the config map %#v", cm.Data)
+		return false, h.processError(srcfg, v1alpha1.ConfigurationValid, corev1.ConditionUnknown, err, "%v")
+	}
+	streams := []string{}
+	// Split will return an item of len 1 if skippedStreams is empty;
+	// verified via testing and godoc of method
+	if len(skippedStreams) > 0 {
+		streams = strings.Split(skippedStreams, " ")
+	}
+	if len(streams) != len(srcfg.Spec.SkippedImagestreams) {
+		logrus.Printf("SkippedImageStreams changed from %s, processing %v", skippedStreams, srcfg)
+		return true, nil
+	}
+
+	streamsMap := make(map[string]bool)
+	for _, stream := range streams {
+		streamsMap[stream] = true
+	}
+	for _, stream := range srcfg.Spec.SkippedImagestreams {
+		if _, ok := streamsMap[stream]; !ok {
+			logrus.Printf("SkippedImageStreams changed from %s, processing %v", skippedStreams, srcfg)
+			return true, nil
+		}
+	}
+
+	skippedTemps, ok := cm.Data[skippedtempskey]
+	if !ok {
+		err := fmt.Errorf("could not find the skipped template setting in the config map %#v", cm.Data)
+		return false, h.processError(srcfg, v1alpha1.ConfigurationValid, corev1.ConditionUnknown, err, "%v")
+	}
+	temps := []string{}
+	// Split will return an item of len 1 if skippedStreams is empty;
+	// verified via testing and godoc of method
+	if len(skippedTemps) > 0 {
+		temps = strings.Split(skippedTemps, " ")
+	}
+	if len(temps) != len(srcfg.Spec.SkippedTemplates) {
+		logrus.Printf("SkippedTemplates changed from %s, processing %v", skippedTemps, srcfg)
+		return true, nil
+	}
+
+	tempsMap := make(map[string]bool)
+	for _, temp := range temps {
+		tempsMap[temp] = true
+	}
+	for _, temp := range srcfg.Spec.SkippedTemplates {
+		if _, ok := tempsMap[temp]; !ok {
+			logrus.Printf("SkippedTemplates changed from %s, processing %v", skippedTemps, srcfg)
+			return true, nil
+		}
+	}
+
+	logrus.Printf("Incoming samples registry unchanged from last processed version")
+	return false, nil
+}
+
 func (h *Handler) StoreCurrentValidConfig(condition v1alpha1.SamplesResourceConditionType, srcfg *v1alpha1.SamplesResource) error {
 
 	if condition != v1alpha1.SamplesExist {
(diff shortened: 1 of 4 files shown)


pkg/stub/handler.go
```diff
@@ -40,6 +40,9 @@ const (
 	x86OKDContentRootDir   = "/opt/openshift/operator/okd-x86_64"
 	ppc64OCPContentRootDir = "/opt/openshift/operator/ocp-ppc64le"
 	installtypekey         = "installtype"
+	regkey                 = "reg"
```
bparees: let's do better with these names, we're going to be living with them for a long time.

pkg/stub/handler.go
```diff
@@ -81,16 +84,12 @@ type Handler struct {
...
-	// in an erorr state; our current approach is
-	// to only update the standard samplesResource
-	// cache entry for valid / non-error states
-	waitingForCredential *v1alpha1.SamplesResource
+	configActive   bool
+	registrySecret *corev1.Secret
+	// This signifies that this is rhel install and we
+	// are waiting for the credential since they are
+	// required to access registry.redhat.io
+	waitingForCredential bool
```
bparees: We should be able to determine this simply by looking at the SamplesResource event we recieved in the handler.

1) as long as the SamplesResource points to a registry that needs credentials, and we don't have credentials, and SamplesExist is false, requeue the event.

2) if the SamplesResource points to a registry that needs credentials and status.CredentialsExist==true, proceed with creating the samples

pkg/stub/handler.go
```diff
@@ -116,6 +115,75 @@ type Handler struct {
...
+	skippedStreams, ok := cm.Data[skippedstreamskey]
+	if !ok {
+		err := fmt.Errorf("could not find the skipped stream setting in the config map %#v", cm.Data)
+		return false, h.processError(srcfg, v1alpha1.ConfigurationValid, corev1.ConditionUnknown, err, "%v")
+	}
+	streams := []string{}
+	// Split will return an item of len 1 if skippedStreams is empty;
+	// verified via testing and godoc of method
+	if len(skippedStreams) > 0 {
+		strings.Split(skippedStreams, " ")
```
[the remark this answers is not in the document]
gabemontero: I did ... part of next push

pkg/stub/handler.go
gabemontero: attempt at better names part of next push

pkg/stub/handler.go
gabemontero: I'm getting rid of `h.waitingForCredential` and `h.registrySecret` ...  bit more swizzling but I think worth it

gabemontero:
updates pushed @bparees 

aside from your comments, I removed some additional cached information, with some tweaks to the tests to adjust coverage and additional Gets of secrets or SampleResources.

pkg/stub/handler.go
```diff
@@ -155,11 +215,22 @@ func (h *Handler) StoreCurrentValidConfig(condition v1alpha1.SamplesResourceCond
...
+	var value string
+	for _, val := range srcfg.Spec.SkippedImagestreams {
+		value = val + " "
+	}
+	cm.Data[skippedstreamskey] = value
+	value = ""
+	for _, val := range srcfg.Spec.SkippedTemplates {
+		value = val + " "
+	}
+	cm.Data[skippedtempskey] = value
```
bparees: The comments in this function say we only update the ConfigMap once...   but it seems like you've introduced logic here and above that expects the ConfigMap to store the last applied configuration (skipped content, registry name).

Storing the last applied content makes sense, but then this other logic that expects to only create the configmap exactly once and not update it does not make sense.

pkg/stub/handler.go
```diff
@@ -368,44 +443,62 @@ func (h *Handler) CreateDefaultResourceIfNeeded() error {
...
+	// log error state
+	if srcfg.Spec.InstallType == v1alpha1.RHELSamplesDistribution &&
+		(srcfg.Spec.SamplesRegistry == "" || srcfg.Spec.SamplesRegistry == "registry.redhat.io") {
+
+		if !srcfg.ConditionFalse(v1alpha1.SamplesExist) {
+			err := fmt.Errorf("Cannot create rhel imagestreams to registry.redhat.io without the credentials being available: %#v", srcfg)
+			return h.processError(srcfg, v1alpha1.SamplesExist, corev1.ConditionFalse, err, "%v")
+		} else {
+			// already recorded error/need
+			return fmt.Errorf("Previously noted that cannot create rhel imagestreams to registry.redhat.cio with the credentials being available %#v", srcfg)
```
bparees: how do we know this was previously noted?  Seems like SamplesExist could be false for any number of reasons when we get here, not necessary just because the credentials didn't exist (even if there isn't such a reason today, there could be in the future).

pkg/stub/handler.go
```diff
@@ -514,18 +595,17 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
...
-			err = h.manageDockerCfgSecret(event.Deleted, h.waitingForCredential, dockercfgSecret)
+		srcfg, _ := h.sdkwrapper.Get(v1alpha1.SamplesResourceName)
+		if srcfg != nil {
+			return h.manageDockerCfgSecret(event.Deleted, srcfg, dockercfgSecret)
 		} else {
-			err = h.manageDockerCfgSecret(event.Deleted, h.samplesResource, dockercfgSecret)
-		}
-		if err != nil {
-			return err
+			return fmt.Errorf("Received secret %s but do not have the SampleRegistry yet, requeing", dockercfgSecret.Name)
```
bparees: "requeuing"

pkg/stub/handler.go
```diff
@@ -574,47 +654,27 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
...
-				return nil
-			}
-		} else {
-			// if we change config in the interim to no longer require creds, nil out waitingForCredentials
-			// so it does not impact any future checks
-			h.waitingForCredential = nil
+		// error state will be logged by WaitingForCredential, so return error and requeue
+		err = h.WaitingForCredential(srcfg)
+		if err != nil {
+			return err
```
bparees: I don't think we need to requeue here.

When we see the user's secret, we should the openshift secret and update the samplesresource.status.conditions to include CredentialsExist=true

The handler should then see an update event for SamplesResource(the one that updated the status), recognize the credentials exist, and proceed to create the samples.

pkg/stub/handler.go
```diff
@@ -368,44 +443,62 @@ func (h *Handler) CreateDefaultResourceIfNeeded() error {
...
 
-	if h.samplesResource == nil && h.waitingForCredential == nil {
-		if !deleted {
-			h.registrySecret = secret
+	secret, err := h.secretclientwrapper.Get(h.namespace, v1alpha1.SamplesRegistryCredentials)
+	// have to check both since get will return an empty secret struct along with not found errs
+	// handle case where secret is created in sample operator namespace before SamplesResource
+	if secret != nil && !kerrors.IsNotFound(err) {
+		return h.manageDockerCfgSecret(false, srcfg, secret)
+	}
```
bparees: per my other comment, I don't think you need to do this (looking for the secret manually).

I think all this method has to do is:

1) if credentials exist==true, return false(we aren't waiting for credentials)
2) if credentials exist == false && we need credentials(the checking you do below), return true (we are waiting for credentials)
3) else return false

pkg/stub/handler.go
gabemontero: yep good catch ... need to fix comment, and move the setting of the install type and arch such that we only update that the first time, but allow for updates to the other fields multiple times.

part of next push

pkg/stub/handler.go
gabemontero: because all the h.processError call above does is set the SamplesExist condition to false

I saw no need to induce more churn on the object and update again.  Returning the error still kicks in relist, etc.

If the if check above evolves then yes this path might need to evolve with it

pkg/stub/handler.go

pkg/stub/handler.go
bparees: I would expect us to start defaulting SamplesExist to false, or setting SamplesExist to false any time we see that the configuration is invalid(and we haven't already created the samples), at which point this logic would be flawed.  We can proceed w/ this for now but i think it's another area of behavior we'll need to revisit in the future.

gabemontero: I'll update to always call with some clarifying comment about consciously not trying to minimize duplicate condition settings

pkg/stub/handler.go
gabemontero: not returning error / not initiating requeue part of next push

pkg/stub/handler.go
gabemontero: Might need a face to face conversation on this, but I'll try to capture here first.

- if secret event comes *AFTER* initial samplesresource event, then agreed, do not need the secret lookup nor call to `manageDockerCfgSecret` ... that will happen on the secret event processing

- if secret event comes *BEFORE* samplesresource event, then yeah agree but was uncomfortable because
 -- if the secret relists occur long enough, as we return errors in the secret path, then presumably on of the relists should have the samplesresource available and complete processing.
 -- at least with the SDK version I'm currently using, here is what is documented for relist in informer-sync.go:
```
	// Copy from deployment_controller.go:
	// maxRetries is the number of times a Vault will be retried before it is dropped out of the queue.
	// With the current rate-limiter in use (5ms*2^(maxRetries-1)) the following numbers represent the times
	// a Vault is going to be requeued:
	//
	// 5ms, 10ms, 20ms, 40ms, 80ms, 160ms, 320ms, 640ms, 1.3s, 2.6s, 5.1s, 10.2s, 20.4s, 41s, 82s
	maxRetries = 15

```

-- seems like it should be sufficient window ... but I'm still nervous about the underlying stuff changing on us ... and granted we may move to different underlying stuff anyway

I'll remove, but wanted to at least verbalize the rationale behind the presumably unneeded redundancy.

gabemontero:
updates stemming from comments, plus some additional ones that stemmed from the discussion in your PR last week and our face to face today, pushed

pkg/stub/handler.go
```diff
@@ -131,38 +187,74 @@ func (h *Handler) StoreCurrentValidConfig(condition v1alpha1.SamplesResourceCond
 		// 4.0 testing showed that we were getting empty ConfigMaps
 		cm = nil
 	}
-	if cm != nil {
-		// just return ... we only update the config map once
-		return nil
+	create := false
```
bparees: It's a bit paranoid, but since you are fetching the configmap within this function, there's no guarantee the changes are valid anymore... (the configmap could have changed since you called SpecValidation).

I think this needs to be a continuous flow:

1) confirm the sampleresource is valid on its own (doesn't contain totally invalid values) - stop here if it is
2) fetch the configmap - if it does not exist, create one w/ the values from the SR
3) if the configmap exists, compare the changes, stop here and throw error if changes are unacceptable
4) assuming we didn't get an error earlier, proceed to apply the SR configuration:
   a) mutate the configmap you fetched in (3) to reflect the new configuration and apply/update it (if you get an error here because someone else changed the configmap underneath you, error out and retry)
   b) update/create the samples based on the new configuration
   
I also wonder if we can find our way to getting rid of this configmap entirely(not in this PR, but thinking ahead)...   the reason we introduced it was to ensure we had a copy of the previous config if someone deleted the SR.  But if we take the position that upon deletion of the SR, we delete all the samples, then we do not need to worry about that scenario.  If we did that, we could simply store the current config settings in Status on the SR itself, which I think makes a lot of this logic easier.

pkg/stub/handler.go
```diff
@@ -331,23 +421,43 @@ func (h *Handler) IsRetryableAPIError(err error) bool {
...
 	deleteInProgress := h.deleteInProgress
 
 	var err error
-	var srcfg *v1alpha1.SamplesResource
+	// "4a" in the "startup" workflow, just create default
+	// resource and set up that way
+	srcfg := &v1alpha1.SamplesResource{}
+	srcfg.Name = v1alpha1.SamplesResourceName
+	srcfg.Kind = "SamplesResource"
+	srcfg.APIVersion = "samplesoperator.config.openshift.io/v1alpha1"
```
bparees: get this out of pkg/apis/samplesoperator/v1alpha1/register.go

pkg/stub/handler.go
```diff
@@ -366,46 +476,60 @@ func (h *Handler) CreateDefaultResourceIfNeeded() error {
...
-			h.registrySecret = secret
+	// if trying to do rhel to the default registry.redhat.io registry requires the secret
+	// be in place since registry.redhat.io requires auth to pull; since it is not ready
+	// log error state
+	if srcfg.Spec.InstallType == v1alpha1.RHELSamplesDistribution &&
+		(srcfg.Spec.SamplesRegistry == "" || srcfg.Spec.SamplesRegistry == "registry.redhat.io") {
+		// currently do not attempt to avoid multiple settings of this conditions/message prior to an event
+		// where the situation is addressed
+		err := fmt.Errorf("Cannot create rhel imagestreams to registry.redhat.io without the credentials being available: %#v", srcfg)
+		h.processError(srcfg, v1alpha1.SamplesExist, corev1.ConditionFalse, err, "%v")
```
bparees: this doesn't seem like the appropriate place to make this decision (that the samples do not exist).

this function's only purpose is to determine whether or not we are waiting for credentials, other areas of the code should decide whether the existence/non-existence of credentials has any implication regarding whether the samples exist.

pkg/stub/handler.go
bparees: (we may need to talk through 1-4 on a whiteboard, i suspect i've done an imperfect job explaining my thinking)

pkg/stub/handler.go
gabemontero: ah ... i mean to switch this to setting the import credentials to false, but then when we decided to set the conditions on the create default, I forgot to reconcile this.

addressed in next push

gabemontero: sorry did add the import cred to false (with the detailed message) but just forgot to delete

pkg/stub/handler.go
gabemontero: OK ... I will change the private constants the sdk setup command auto generated to public and do this

pkg/stub/handler.go
gabemontero: Also, looked at the code to refresh my memory, and we *ALREADY ARE* deleting:
1. the imagestreams in openshift namespace, but factor in skipped filter
2. the templates in openshift namespace, but factor in skipped filter
3. the secret in openshift namespace
4. the config map in the operator's namespace

Since the operator does not create the secret in its namespace, we left it alone.

gabemontero: OK @bparees and I decided to just stay pat on this thread for now, think more about it in the background.

gabemontero:
ok pushed latest updates in separate commit @bparees

gabemontero:
also, per prior discussions in https://github.com/openshift/cluster-samples-operator/pull/36, I've scanned all uses of `sdk.Update(..)` and am convinced we are not modifying the Spec struct in any of our interactions.

And other than the finalizer annotation, we are only modifying Status.

bparees:
squash and lgtm.

gabemontero:
ok clean tests again

gabemontero:
also, I have a commit locally that adds the manage state processing @bparees  ... do you have any preference on including it in this PR vs. a new PR ?

bparees:
> also, I have a commit locally that adds the manage state processing @bparees ... do you have any preference on including it in this PR vs. a new PR ?

new PR please

gabemontero:
OK saved the managed stuff, not in this PR

Pushed a small update based on installer testing (no selinux issues today, and the samples operator is getting installed again - no problem from the other operator crashing out)

separate commits ... will squash when appropriate

@bparees ptal when you get a chance

pkg/stub/handler.go
```diff
@@ -481,7 +481,7 @@ func (h *Handler) CreateDefaultResourceIfNeeded() (*v1alpha1.SamplesResource, er
 	}
 
 	s, err := h.sdkwrapper.Get(v1alpha1.SamplesResourceName)
-	if s != nil && !kerrors.IsNotFound(err) {
+	if s == nil || kerrors.IsNotFound(err) {
```
bparees: heh. doh.

bparees:
squash and still lgtm

PR #39, opened 2018-10-29
add managementstate;

gabemontero:
move sdk/samplesresource updates up to top of stack (allow grouping of changes);
genericize retry on api interactions;
move config map creation to startup

### DIFF
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -30,6 +30,7 @@ import (
 	imagev1client "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
 	templatev1client "github.com/openshift/client-go/template/clientset/versioned/typed/template/v1"
 
+	operatorsv1alpha1api "github.com/openshift/api/operator/v1alpha1"
 	"github.com/openshift/cluster-samples-operator/pkg/apis/samplesoperator/v1alpha1"
 	operatorstatus "github.com/openshift/cluster-samples-operator/pkg/operatorstatus"
 )
@@ -184,33 +185,34 @@ func (h *Handler) StoreCurrentValidConfig(condition v1alpha1.SamplesResourceCond
 		return err
 	}
 	if kerrors.IsNotFound(err) {
-		// 4.0 testing showed that we were getting empty ConfigMaps
-		cm = nil
+		err = fmt.Errorf("Operator in compromised state; Could not find config map even though samplesresource exists: %#v", srcfg)
+		h.processError(srcfg, v1alpha1.SamplesExist, corev1.ConditionUnknown, err, "%v")
+		h.processError(srcfg, v1alpha1.ConfigurationValid, corev1.ConditionUnknown, err, "%v")
+		return err
 	}
-	create := false
-	if cm == nil {
-		create = true
-		cm = &corev1.ConfigMap{}
-		cm.Name = v1alpha1.SamplesResourceName
+
+	if cm.Data == nil {
 		cm.Data = map[string]string{}
-		if len(srcfg.Spec.InstallType) == 0 {
-			cm.Data[installtypekey] = string(v1alpha1.CentosSamplesDistribution)
-		} else {
-			cm.Data[installtypekey] = string(srcfg.Spec.InstallType)
-		}
-		if len(srcfg.Spec.Architectures) == 0 {
-			cm.Data[v1alpha1.X86Architecture] = v1alpha1.X86Architecture
-		} else {
-			for _, arch := range srcfg.Spec.Architectures {
-				switch arch {
-				case v1alpha1.X86Architecture:
-					cm.Data[v1alpha1.X86Architecture] = v1alpha1.X86Architecture
-				case v1alpha1.PPCArchitecture:
-					cm.Data[v1alpha1.PPCArchitecture] = v1alpha1.PPCArchitecture
-				}
+	}
+
+	if len(srcfg.Spec.InstallType) == 0 {
+		cm.Data[installtypekey] = string(v1alpha1.CentosSamplesDistribution)
+	} else {
+		cm.Data[installtypekey] = string(srcfg.Spec.InstallType)
+	}
+	if len(srcfg.Spec.Architectures) == 0 {
+		cm.Data[v1alpha1.X86Architecture] = v1alpha1.X86Architecture
+	} else {
+		for _, arch := range srcfg.Spec.Architectures {
+			switch arch {
+			case v1alpha1.X86Architecture:
+				cm.Data[v1alpha1.X86Architecture] = v1alpha1.X86Architecture
+			case v1alpha1.PPCArchitecture:
+				cm.Data[v1alpha1.PPCArchitecture] = v1alpha1.PPCArchitecture
 			}
 		}
 	}
+
 	cm.Data[regkey] = srcfg.Spec.SamplesRegistry
 	var value string
 	for _, val := range srcfg.Spec.SkippedImagestreams {
@@ -222,11 +224,7 @@ func (h *Handler) StoreCurrentValidConfig(condition v1alpha1.SamplesResourceCond
 		value = val + " "
 	}
 	cm.Data[skippedtempskey] = value
-	if create {
-		_, err = h.configmapclientwrapper.Create(h.namespace, cm)
-	} else {
-		_, err = h.configmapclientwrapper.Update(h.namespace, cm)
-	}
+	_, err = h.configmapclientwrapper.Update(h.namespace, cm)
 	return err
 }
 
@@ -262,7 +260,7 @@ func (h *Handler) SpecValidation(srcfg *v1alpha1.SamplesResource) (*corev1.Confi
 	// if we have not had a valid SamplesResource processed, allow caller to try with
 	// the srcfg contents
 	if !srcfg.ConditionTrue(v1alpha1.SamplesExist) {
-		logrus.Println("Spec is valid because samples are not created yet")
+		logrus.Println("Spec is valid because this operator has not processed a config yet")
 		return nil, nil
 	}
 	cm, err := h.configmapclientwrapper.Get(h.namespace, v1alpha1.SamplesResourceName)
@@ -319,7 +317,8 @@ func (h *Handler) SpecValidation(srcfg *v1alpha1.SamplesResource) (*corev1.Confi
 		err = fmt.Errorf("cannot change architectures from %#v to %#v", cm.Data, srcfg.Spec.Architectures)
 		return nil, h.processError(srcfg, v1alpha1.ConfigurationValid, corev1.ConditionFalse, err, "%v")
 	}
-	return cm, h.GoodConditionUpdate(srcfg, corev1.ConditionTrue, v1alpha1.ConfigurationValid)
+	h.GoodConditionUpdate(srcfg, corev1.ConditionTrue, v1alpha1.ConfigurationValid)
+	return cm, nil
 }
 
 func (h *Handler) AddFinalizer(srcfg *v1alpha1.SamplesResource) {
@@ -362,7 +361,7 @@ func (h *Handler) NeedsFinalizing(srcfg *v1alpha1.SamplesResource) bool {
 	return false
 }
 
-func (h *Handler) GoodConditionUpdate(srcfg *v1alpha1.SamplesResource, newStatus corev1.ConditionStatus, conditionType v1alpha1.SamplesResourceConditionType) error {
+func (h *Handler) GoodConditionUpdate(srcfg *v1alpha1.SamplesResource, newStatus corev1.ConditionStatus, conditionType v1alpha1.SamplesResourceConditionType) {
 	logrus.Debugf("updating condition %s to %s", conditionType, newStatus)
 	condition := srcfg.Condition(conditionType)
 	// decision was made to not spam master if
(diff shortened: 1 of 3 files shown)


pkg/stub/handler.go
```diff
@@ -187,12 +189,23 @@ func (h *Handler) StoreCurrentValidConfig(condition v1alpha1.SamplesResourceCond
 		// 4.0 testing showed that we were getting empty ConfigMaps
 		cm = nil
 	}
+	//TODO our create default samplesresource logic attempls to create the config map,
+	// but for now we have redundancy here in case it still does not exists
+	// Is this valid ??
```
bparees: It is probably not valid.  It means we've created samples, but we don't have a record of the configuration that lead to those samples.  If we don't have that record, we can't validate changes to the configuration, which means we can't ever have another known valid configuration.

I would say if the configmap does not exist, and the samples do exist, we are in a broken state that requires user/admin intervention to reconcile.

pkg/stub/handler.go
```diff
@@ -187,12 +189,23 @@ func (h *Handler) StoreCurrentValidConfig(condition v1alpha1.SamplesResourceCond
...
 		cm.Name = v1alpha1.SamplesResourceName
+	}
+
+	if cm.Data == nil {
 		cm.Data = map[string]string{}
+	}
+
+	// if this is the initial processing, then store the values for the spec
+	// values that cannot change after the initial processing
+	if _, ok := cm.Data[initialprocessingkey]; !ok {
```
bparees: why wouldn't we just always store these values?  if it hasn't changed, it's a no-op.

pkg/stub/handler.go
```diff
@@ -606,18 +638,66 @@ func (h *Handler) CleanUpOpenshiftNamespaceOnDelete(srcfg *v1alpha1.SamplesResou
 			err = h.templateclientwrapper.Delete("openshift", temp.Name, &metav1.DeleteOptions{})
 			if err != nil && !kerrors.IsNotFound(err) {
 				logrus.Warnf("Problem deleting openshift template %s on SamplesResource delete: %#v", temp.Name, err)
+			} else {
+				err = nil
```
bparees: what's the point of this?  you don't do anything with this value.

pkg/stub/handler.go
```diff
@@ -606,18 +638,66 @@ func (h *Handler) CleanUpOpenshiftNamespaceOnDelete(srcfg *v1alpha1.SamplesResou
...
 			}
 		}
 	}
 
 	err = h.secretclientwrapper.Delete("openshift", v1alpha1.SamplesRegistryCredentials, &metav1.DeleteOptions{})
 	if err != nil && !kerrors.IsNotFound(err) {
 		logrus.Warnf("Problem deleting openshift secret %s on SamplesResource delete: %#v", v1alpha1.SamplesRegistryCredentials, err)
+	} else {
+		err = nil
```
bparees: same here, you never read this value of "err"

pkg/stub/handler.go
```diff
@@ -606,18 +638,66 @@ func (h *Handler) CleanUpOpenshiftNamespaceOnDelete(srcfg *v1alpha1.SamplesResou
...
 	}
 
 	err = h.configmapclientwrapper.Delete(h.namespace, v1alpha1.SamplesResourceName)
 	if err != nil && !kerrors.IsNotFound(err) {
 		logrus.Warnf("Problem deleting openshift configmap %s on SamplesResource delete: %#v", v1alpha1.SamplesResourceName, err)
+	} else {
+		err = nil
```
bparees: and here

pkg/stub/handler.go
```diff
@@ -606,18 +638,66 @@ func (h *Handler) CleanUpOpenshiftNamespaceOnDelete(srcfg *v1alpha1.SamplesResou
...
+	return nil
+}
+
+// ProcessManagementField returns true if the operator should handle the SampleResource event
+// and false if it should not, as well as an err in case we want to bubble that up to
+// the controller level logic for retry
+func (h *Handler) ProcessManagementField(srcfg *v1alpha1.SamplesResource) (bool, error) {
+	switch srcfg.Spec.ManagementState {
+	case operatorsv1alpha1api.Removed:
+		if srcfg.ConditionFalse(v1alpha1.SamplesExist) {
```
bparees: isn't it possible the credentials exist but the samples don't, in which case the credentials still need to be cleaned up?

pkg/stub/handler.go
```diff
@@ -649,8 +741,6 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
 		// Every time we see a change to the SamplesResource object, update the ClusterOperator status
 		// based on the current conditions of the SamplesResource.
 		err := h.cvowrapper.UpdateOperatorStatus(srcfg)
```
bparees: if the samplesresource is `unmanaged` we should not be doing this, or at least i think we need to revisit what we want to report for status. Probably need to think about the implications for `removed` too.

pkg/stub/handler.go
```diff
@@ -678,6 +769,18 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
...
+		if err != nil {
+			// flush status update
+			h.sdkwrapper.Update(srcfg)
+			return err
+		}
+		if !doit {
+			// flush status update
+			h.sdkwrapper.Update(srcfg)
+			return nil
+		}
```
bparees: combine these into a single if?  `(if !doit || err!=nil)`

pkg/stub/handler.go
```diff
@@ -1194,11 +1303,27 @@ type defaultSDKWrapper struct {
...
 func (g *defaultSDKWrapper) Update(samplesResource *v1alpha1.SamplesResource) error {
-	return sdk.Update(samplesResource)
+	err := sdk.Update(samplesResource)
+	count := 0
+	for count < 10 {
+		if !g.h.IsRetryableAPIError(err) {
+			break
+		}
+		err = sdk.Update(samplesResource)
+	}
```
bparees: should probably have a delay here.. use the wait.Poll helper.

pkg/stub/handler.go
```diff
@@ -1194,11 +1303,27 @@ type defaultSDKWrapper struct {
...
 func (g *defaultSDKWrapper) Create(samplesResource *v1alpha1.SamplesResource) error {
-	return sdk.Create(samplesResource)
+	err := sdk.Create(samplesResource)
+	count := 0
+	for count < 10 {
+		if !g.h.IsRetryableAPIError(err) {
+			break
+		}
+		err = sdk.Create(samplesResource)
+	}
```
bparees: wait.poll

pkg/stub/handler.go
```diff
@@ -1207,6 +1332,13 @@ func (g *defaultSDKWrapper) Get(name string) (*v1alpha1.SamplesResource, error)
 	sr.APIVersion = "samplesoperator.config.openshift.io/v1alpha1"
 	sr.Name = name
 	err := sdk.Get(&sr, sdk.WithGetOptions(&metav1.GetOptions{}))
+	count := 0
+	for count < 10 {
+		if !g.h.IsRetryableAPIError(err) {
+			break
+		}
+		err = sdk.Get(&sr, sdk.WithGetOptions(&metav1.GetOptions{}))
+	}
```
bparees: wait.poll

pkg/stub/handler.go
```diff
@@ -42,6 +43,7 @@ const (
 	regkey                 = "keyForSamplesRegistryField"
 	skippedstreamskey      = "keyForSkippedImageStreamsField"
 	skippedtempskey        = "keyForSkippedTemplatesField"
+	initialprocessingkey   = "keyForIndicatingInitialProcessing"
```
bparees: any sort of "we need to know this is the first time because we do something special" logic makes me nervous/wary.

pkg/stub/handler.go
gabemontero: sounds good ... will propagate as an error

pkg/stub/handler.go
gabemontero: I think I'm struggling with my paranoia around future changes subverting assumptions of the current flow

I'll remove `initialprocessingkey`

pkg/stub/handler.go
gabemontero: per comment below removing (resolving my paranoia)

bparees: yeah, paranoia is valid, but we should try to solve it by inspecting the current state of the system before adding new state (since that new state can in turn end up wrong/corrupt/invalid/racey)

pkg/stub/handler.go
gabemontero: ah leftover ... removing

pkg/stub/handler.go
gabemontero: No either deleting the samplesresource or marking it removed will delete the credential created by the operator as well as the samples.

pkg/stub/handler.go
gabemontero: We can certainly discuss, but initially I think management state should not stop the reporting of operator status....the operator is available to receive input regardless of whether or not is processing changes ....

pkg/stub/handler.go
gabemontero: I am varying whether I return an error or not ... that is why the are not combined

pkg/stub/handler.go
gabemontero: Yeah forgot about that (used it elsewhere) ... thanks

pkg/stub/handler.go
gabemontero: duh err is nil in the case where i return nil ... will change

pkg/stub/handler.go
bparees: I guess the real question is what should the ClusterOperator status be when the resource is unmanaged?

available:  reflect the last known state before the resource was marked unmanaged?  (is there any way this status can change once the resource is marked unmanaged?  I don't think so..?).  So this one probably "just works" w/o any special logic.

progressing:  do we keep monitoring resources we may have already installed (eg imagestreams we installed but weren't imported yet?) and continue to update this if/when those resources successfully import?  I think the answer may be yes...  Though I could also see an argument for immediately putting "progressing" to false when you mark the resource unmanaged, because the operator is no longer actively attempting to apply any changes you request.

failed:  presumably this never changes since we're not going to look at the config changes while unmanaged, so again it should never changed once the resource is marked unmanaged, so probably not a specific concern.

gabemontero: available: agreement with what you said, but with one add-on:
- deleting the samplesresource with the current code "takes precedence" and results in "mangement" activity occurring .... in the vacuum of me iterating over this code that seemed valid, but admin it is nuanced enough that reasonable minds could disagree.

Your take?

failed: agreement with what you said

progressing: yeah, the either/or 's you mentioned don't necessarily have a clear winner in my mind at first blush as well;  at this point, I'd rather revisit the relationship between progressing and unmanaged *AFTER* we nail down what progressing will entail.

Thoughts?

bparees: > deleting the samplesresource with the current code "takes precedence" and results in "mangement" activity occurring .... in the vacuum of me iterating over this code that seemed valid, but admin it is nuanced enough that reasonable minds could disagree.

hm.  yeah my inclination is to say if they have it marked "unmanaged" and then they delete it, we should not do anything (beyond deleting the samplesresource).  Unfortunately that means we're then going to create a new samplesresource which is managed, so the whole thing becomes a bit moot.

It may well be more consistent to treat deleting the resource as superceding unmanaged, as you suggest, since at least then the rest of the behavior is "logical".  So sure, let's go with that.

and i'm fine with deferring what to do w/ progressing until we actually have a valid progressing status.

gabemontero: ok cool thx

pkg/stub/handler.go
gabemontero: I think we still need to get/confirm consensus here as well @bparees

bparees: Sorry i'm not seeing it, what's the code flow that leads to the credential being deleted when the samplesresource is marked removed (particularly is samplesexist is false)?

gabemontero: Flow I tested:
- default bringup, mgmt state is managed, samples installed, exist is true
- admin creats secret in operator's namespace
- operator creates secret in openshift namespace
- admin sets mgmt to removed
- first time through, exists is still true, calls `CleanUpOpenshiftNamespaceOnDelete`
- `CleanUpOpenshiftNamespaceOnDelete` deletes the samples and the secret in the openshift namespace, marks exist/cred conditions as false, updates samplesreource
- samples resource update reflecting updated conditions ... ignores since already processed given exists is false ... checking credentials is false is redundant

Only `CleanUpOpenshiftNamespaceOnDelete` removes things ... it always removes both samples and secret.

bparees: right, but you could have a scenario where the secret exists but the samples do not exist, in which case this code is going to skip over the cleanup and not remove the secret.

gabemontero: actually, no, I do not think you can have such a scenario now ... can you describe how you think we can get in such a state?

bparees: 1) SamplesResource is created, references rhel.  We do not create samples because we don't have a secret yet.

2) User creates their secret, we copy the secret into the openshift namespace and update the resource to samplesexist=true

3) User marks the resource to "Removed" either before we attempt to create the samples, or we encounter a failure trying to create the samples and then the user marks the resource "Removed".

gabemontero: those are some tight timing windows but sure, will adjust if-check to account for them

bparees: the case where we are failing to create the samples for some reason isn't necessarily a tight timing window, we could be in that state for a long time if something gets screwed up in RBAC or we have broken samples or something else.

it's also possible in the future we introduce new scenarios where we end up there (perhaps we introduce a second gating condition on creating the samples, such that it becomes easy to get in a scenario where you've created the secret, but can't create the samples yet because you're waiting for another condition still).  So I want to make sure we're taking all possible states into account.

bparees: (I know you're convinced/acquiescing on this specific case, but in general I think those are the kinds of things we need to be considering when we are building our operator state machines)

gabemontero:
update pushed @bparees .... that is all the comments to date

bparees:
i'm gonna call it good for this iteration

gabemontero:
thanks @bparees ... squashed and labelled

PR #40, opened 2018-10-31
remove master node selector

bparees:


bparees:
This selector is causing frustration/confusion when people try to deploy the operator on clusters that do not have a master node labeled, or that have other default nodeselector's being applied to their resources.  Until we hear otherwise, I don't see a good reason to restrict ourselves to specifically the master node.

/assign @gabemontero

PR #41, opened 2018-11-01
add samples watches; fix sdk get/upd/create retry; other minor fixes

gabemontero:

/assign @bparees 

includes comment detailing my failed attempts to use the current sdk version vendored in to watch imagestreams and templates ... not just using "v1" as the api version failed (though with a basic could not find the resource error) ... the `k8sutil.RuntimeObjectFromUnstructured` is what bombs trying to convert the json payload into a go struct/obj.

### DIFF
--- a/cmd/cluster-samples-operator/main.go
+++ b/cmd/cluster-samples-operator/main.go
@@ -10,6 +10,9 @@ import (
 	k8sutil "github.com/operator-framework/operator-sdk/pkg/util/k8sutil"
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
 
+	imagev1 "github.com/openshift/api/image/v1"
+	templatev1 "github.com/openshift/api/template/v1"
+
 	"github.com/sirupsen/logrus"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 )
@@ -36,6 +39,11 @@ func main() {
 	sdk.Watch(resource, kind, "", resyncPeriod)
 	logrus.Infof("Watching secrets")
 	sdk.Watch("v1", "Secret", namespace, resyncPeriod)
+	sdk.Watch("v1", "Secret", "openshift", resyncPeriod)
+	k8sutil.AddToSDKScheme(imagev1.AddToScheme)
+	k8sutil.AddToSDKScheme(templatev1.AddToScheme)
+	sdk.Watch(imagev1.SchemeGroupVersion.String(), "ImageStream", "openshift", resyncPeriod)
+	sdk.Watch(templatev1.SchemeGroupVersion.String(), "Template", "openshift", resyncPeriod)
 	sdk.Handle(stub.NewHandler())
 	sdk.Run(context.TODO())
 }
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -21,6 +21,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilnet "k8s.io/apimachinery/pkg/util/net"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apimachinery/pkg/watch"
 
 	restclient "k8s.io/client-go/rest"
 
@@ -71,6 +72,9 @@ func NewHandler() sdk.Handler {
 	h.mutex = &sync.Mutex{}
 	h.CreateDefaultResourceIfNeeded()
 
+	h.imagestreamFile = make(map[string]string)
+	h.templateFile = make(map[string]string)
+
 	return &h
 }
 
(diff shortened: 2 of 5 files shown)


cmd/cluster-samples-operator/main.go
```diff
@@ -36,6 +36,41 @@ func main() {
...
+	sdk.Watch("v1", "Secret", "openshift", resyncPeriod)
+	// note, sdk watch does not work for openshift types like imagestream or templates
+	// attempts included both v1 and image.openshift.io/v1 for imagestreams and
+	// both v1 and template.openshift.io/v1 for templates
+	// using the go client for that
+	//sdk.Watch("image.openshift.io/v1", "ImageStream", "openshift", resyncPeriod)
+	//sdk.Watch("template.openshift.io/v1", "Template", "openshift", resyncPeriod)
+	/* For example:
+	panic: failed to decode json data with gvk(template.openshift.io/v1, Kind=Template): no kind "Template" is registered for version "template.openshift.io/v1" [recovered]
+		panic: failed to decode json data with gvk(template.openshift.io/v1, Kind=Template): no kind "Template" is registered for version "template.openshift.io/v1"
```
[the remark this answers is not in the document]
gabemontero: just be a bug in the sdk ... at least the version I'm on ? .... maybe
subtle nuanced differences between deploymentconfigs vs. imagestreams / templates? ... doubt it

will explore
- approximating the `SchemeGroupVersion.String()` usage the registry guys are doing ... see if it makes any difference
- either
--- upgrade the sdk (including following the migration doc)
--- open a bug on the sdk (though they'll most likely say to upgrade first)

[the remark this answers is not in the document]
bparees: Did you add the scheme? ala `k8sutil.AddToSDKScheme(appsapi.AddToScheme)` ?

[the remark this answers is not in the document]
gabemontero: ah I bet that is it (missed that previously) ... coding it up now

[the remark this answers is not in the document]
gabemontero: yep it works now ... thanks ... will push change shortly

pkg/stub/handler.go
```diff
@@ -862,6 +960,73 @@ func (h *Handler) processError(opcfg *v1alpha1.SamplesResource, ctype v1alpha1.S
 	return err
 }
 
+func (h *Handler) imageStream(imagestream, is *imagev1.ImageStream, opcfg *v1alpha1.SamplesResource) error {
```
bparees: this method name is not very enlightening....  also the two imagestream parameters named "imagestream" and "is" leave me unclear on the difference between the two.

pkg/stub/handler.go
```diff
@@ -862,6 +960,73 @@ func (h *Handler) processError(opcfg *v1alpha1.SamplesResource, ctype v1alpha1.S
...
+		} else {
+			imagestream.ResourceVersion = is.ResourceVersion
+			_, err := h.imageclientwrapper.Update("openshift", imagestream)
+			if err != nil {
+				return h.processError(opcfg, v1alpha1.SamplesExist, corev1.ConditionUnknown, err, "imagestream update error: %v")
+			}
+			logrus.Printf("updated imagestream %s", is.Name)
+		}
+	} else {
+		if is.Labels == nil {
```
bparees: `is` could still be nil here, presumably.

pkg/stub/handler.go
```diff
@@ -862,6 +960,73 @@ func (h *Handler) processError(opcfg *v1alpha1.SamplesResource, ctype v1alpha1.S
...
+		}
+		is.Labels[v1alpha1.SamplesManagedLabel] = "false"
+		h.imageclientwrapper.Update("openshift", is)
+		// if we get an error, we'll just try to remove the label next
+		// time; and we'll examine the skipped lists on delete
+	}
+	return nil
+}
+
+func (h *Handler) template(template, t *templatev1.Template, opcfg *v1alpha1.SamplesResource) error {
```
bparees: same naming issues here w/ the method name + param names.

pkg/stub/handler.go
```diff
@@ -862,6 +960,73 @@ func (h *Handler) processError(opcfg *v1alpha1.SamplesResource, ctype v1alpha1.S
...
+			}
+			logrus.Printf("updated template %s", t.Name)
+		}
+	} else {
+		t.Labels[v1alpha1.SamplesManagedLabel] = "false"
+		h.templateclientwrapper.Update("openshift", t)
+		// if we get an error, we'll just try to remove the label next
+		// time; and we'll examine the skipped lists on delete
+	}
+	return nil
```
bparees: idiomatic go (yeah i hate myself) for this structure (and the other method) should be:

```
if !skippedItem {
  do stuff
  return nil
}
// handle skippedItem logic
return nil
```

Or even better since the skippedItem logic is simpler, put that in the if block (ie minimize indentation).

pkg/stub/handler.go
```diff
@@ -99,11 +103,90 @@ type Handler struct {
...
+		logrus.Warningf("Problem accessing config map during image stream watch event processing: %#v", err)
+		return nil, "", false
+	}
+	if annotations != nil {
+		isv, ok := annotations[v1alpha1.SamplesVersionAnnotation]
+		if ok && isv == v1alpha1.CodeLevel {
+			return nil, "", false
+		}
+	}
+	filePath, ok := cm.Data[name]
```
bparees: this doesn't handle the possibility of a template + imagestream w/ the same name...you need to prefix your keys w/ the type.  

I also think you should prefix them with the namespace and make sure that the event you're processing is for a template/imagestream from the openshift namespace unless you've proven to yourself that the SDK watcher actually respects the namespace you asked to watch (i'm pretty sure it doesn't).

pkg/stub/handler.go
gabemontero: method names will be prefaced with `upsert`

parameters names will be suffixed by `InOperatorImage` and `InCluster`

pkg/stub/handler.go
gabemontero: method names will be prefaced with `upsert`

parameters names will be suffixed by `InOperatorImage` and `InCluster`

pkg/stub/handler.go
```diff
@@ -99,11 +103,90 @@ type Handler struct {
...
+	if err != nil {
+		logrus.Warningf("Problem accessing config map during image stream watch event processing: %#v", err)
+		return nil, "", false
+	}
+	if annotations != nil {
+		isv, ok := annotations[v1alpha1.SamplesVersionAnnotation]
+		if ok && isv == v1alpha1.CodeLevel {
+			return nil, "", false
+		}
+	}
```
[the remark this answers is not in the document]
gabemontero: good call ... part of next push

pkg/stub/handler.go
gabemontero: yep ... and same in the template one ... part of next push

pkg/stub/handler.go

gabemontero:
updates pushed @bparees

pkg/stub/handler.go
```diff
@@ -130,13 +136,7 @@ func (h *Handler) prepWatchEvent(name string, annotations map[string]string) (*v
...
 		return nil, "", false
 	}
-	if annotations != nil {
-		isv, ok := annotations[v1alpha1.SamplesVersionAnnotation]
-		if ok && isv == v1alpha1.CodeLevel {
-			return nil, "", false
-		}
-	}
-	filePath, ok := cm.Data[name]
+	filePath, ok := cm.Data[kind+name]
```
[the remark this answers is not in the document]
gabemontero: I did originally use a separator, but happened to pick one (`/`) that broke the ConfigMap.Data map functionality, where you couldn't find entries one had just put in the map

then got fed up with having to look up the rules, removed the separator character (the only change between runs) and it worked fine :-)

I'll search for the rules, add a separator that works, then squash and merge ... thanks @bparees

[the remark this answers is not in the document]
bparees: hm.  i'm surprised it's so restrictive but i guess it's probably using the dns-naming rules like everything else.  Which is too bad, I was hoping to pick a separator character that couldn't appear in the kind or name value.

Anyway `.` or `-` are probably your best bets if indeed it's dns naming rules.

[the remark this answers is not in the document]
gabemontero: went with hyphen

PR #47, opened 2018-11-07
use pull policy always

bparees:


bparees:
/assign @gabemontero 

this will make your life easier when doing local iterative dev and I can't think of a good reason not to use this policy in general.

PR #48, opened 2018-11-08
add progressing cluster status (imagestreams only)

gabemontero:

/assign @bparees 

NOTE - I did not add e2e tests yet because the "in between state" of clusteroperator wrt its groups etc. is causing odd behavior in my test cluster ... via debug, I'm confirming that creation/update of the object is happening, but "oc get clusteroperator" returns no resources

the samplesresource updates are working just fine, but I'd rather add the tests all at once

I'll refresh the installer and update deps here as the clusteroperator situation progresses, and add e2e's accordingly.

### DIFF
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -115,16 +115,18 @@ func (h *Handler) prepWatchEvent(kind, name string, annotations map[string]strin
 	if h.deleteInProgress {
 		return nil, "", false
 	}
-	if annotations != nil {
-		isv, ok := annotations[v1alpha1.SamplesVersionAnnotation]
-		if ok && isv == v1alpha1.CodeLevel {
-			return nil, "", false
-		}
-	}
 	srcfg, err := h.sdkwrapper.Get(v1alpha1.SamplesResourceName)
 	if srcfg == nil || err != nil {
 		logrus.Printf("Received watch event %s but ignoring since not have the SamplesResource yet: %#v %#v", kind+"/"+name, err, srcfg)
 		return nil, "", false
+	}
+	if annotations != nil {
+		isv, ok := annotations[v1alpha1.SamplesVersionAnnotation]
+		if ok && isv == v1alpha1.CodeLevel {
+			logrus.Debugf("ignoring %s/%s cause operator version matches", kind, name)
+			// but return srcfg to potentially toggle pending condition
+			return srcfg, "", false
+		}
 	}
 	proceed, err := h.ProcessManagementField(srcfg)
 	if !proceed || err != nil {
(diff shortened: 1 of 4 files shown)


bparees:
I have mixed feelings but i think i'm leaning towards a single condition for "imagestreamsimporting" instead of one per imagestream....

perhaps the reason field of the condition can list all the imagestreams that are still importing, when imagestreamimporting is true?

i'd rather not have N condition objects, it becomes a lot to look through.

pkg/stub/handler.go
```diff
@@ -145,45 +147,92 @@ func (h *Handler) prepWatchEvent(kind, name string, annotations map[string]strin
...
+			if len(is.Spec.Tags) == len(is.Status.Tags) {
+				for _, specTag := range is.Spec.Tags {
+					matched := false
+					for _, statusTag := range is.Status.Tags {
+						logrus.Debugf("checking spec tag %s against status tag %s with num items %d", specTag.Name, statusTag.Tag, len(statusTag.Items))
+						if specTag.Name == statusTag.Tag {
+							for _, event := range statusTag.Items {
+								logrus.Debugf("checking status tag %d against spec tag %#v", event.Generation, specTag.Generation)
+								if specTag.Generation != nil &&
+									*specTag.Generation <= event.Generation {
```
bparees: were you able to confirm this is a valid way to check the import status of a tag?  it was a bit of a guess on my part, it's been a while since i've dug through the imagestreamimport code flow.

gabemontero: Yeah it definitely looked valid.  There was a gradual progression of imagestreams getting to completion state as images were downloaded.  Working from home yesterday, with a slower network, made that progression even more deliberate and pronounced.  An eventually all the tags matched up.

I was able to watch specific image streams via `oc` as well as monitor the samplesresource obj via `oc` and run with debug in the samples operator pod, and line things up.  And as I mentioned in the description, while my pod debug showed the clusteroperator api calls flowing and completing without error, and the final updates setting progressing to false after it had been true when the imagestream updates started, there was no clusteroperator object to look at given the in between state that api is in based on what you and Oleg found.

Doing the `>=` check vs. `==` check was even spot on, as the `latest` tag often times was a tag ref vs. an image ref, so you would have cases where the spec generation was 1, but the status generation was 2, for example.

bparees:
I would normally expect the list to be empty (all images should be imported unless things are broken), so to me that motivates the single condition instead of many of them, because you're going to be paying the readability cost to see a bunch of "imagestreamimporting=false" most of the time.

manipulating the string to remove entries is a bit of a pain i agree, but nothing a split+map shouldn't be able to handle for you.

gabemontero:
OK pros/cons have been suitable registered here.  I'll switch to a single condition.

gabemontero:
@bparess the switch to a single condition is pushed ... ptal

bparees:
/lgtm

i'm ok w/ merging this and circling back to add the ClusterOperator status e2e when you make the change to use the new ClusterOperator group.

gabemontero:
I'm thinking now to add a slew of e2e tests once the clusteroperator situation is sorted out (realistically after this PR if it does not become contentious)

gabemontero:
doh ... now saw @bparees 's comment :-)

PR #51, opened 2018-11-11
Add a RHEL7 dockerfile and standarize format

smarterclayton:


gabemontero:
so the unit test failure has consistently been this problem after the actual test passed

success notification:

`ok  	github.com/openshift/cluster-samples-operator/pkg/stub	0.030s`

error:

`error waiting for operator deployment to exist: deployments.apps "cluster-samples-operator" is forbidden: User "system:serviceaccount:ci-op-kzpmvvj6:default" cannot get deployments.apps in the namespace "openshift-cluster-samples-operator": no RBAC policy matched`

@smarterclayton - of course the rbac's are injected into the image during its construction ... from my scan, it would appear your dockerfile changes are still doing that.  Do you think there is in fact something amiss with the dockerfile change, or is the lack of an RBAC indicative of an issue with the api.ci test image or CVO ?

thanks

bparees:
> error waiting for operator deployment to exist: deployments.apps "cluster-samples-operator" is forbidden: User "system:serviceaccount:ci-op-kzpmvvj6:default" cannot get deployments.apps in the namespace "openshift-cluster-samples-operator": no RBAC policy matched

@gabemontero that should not be run as part of our unit tests at all.  That looks like output from the e2e test that I added, but the unit test should not be executing that:

https://github.com/openshift/cluster-samples-operator/blob/master/Makefile#L14

bparees:
/lgtm

will deal w/ reordering these dockerfile lines in a followup.

bparees:
And the job definition is calling make test-unit:
https://github.com/openshift/release/blob/master/ci-operator/config/openshift/cluster-samples-operator/openshift-cluster-samples-operator-master.yaml#L22-L25

So I don't know what happened in https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cluster-samples-operator/51/pull-ci-openshift-cluster-samples-operator-master-unit/87/ that it appears to have tried to run the e2e tests as part of running the unit tests.

guess we'll have to keep an eye out in case it continues to happen.

PR #54, opened 2018-11-15
add e2e tests (along with fixes as a result)

gabemontero:
/assign @bparees 

Of course this is not the complete set of e2e's to do this sprint, but I uncovered enough code fix of relevance that I decided to break things up.

So here is the initial set

### DIFF
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -7,7 +7,6 @@ import (
 	"io/ioutil"
 	"os"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/operator-framework/operator-sdk/pkg/sdk"
@@ -54,9 +53,9 @@ func NewHandler() sdk.Handler {
 
 	h.sdkwrapper = &defaultSDKWrapper{h: &h}
 
-	h.fileimagegetter = &defaultImageStreamFromFileGetter{h: &h}
-	h.filetemplategetter = &defaultTemplateFromFileGetter{h: &h}
-	h.filefinder = &defaultResourceFileLister{h: &h}
+	h.Fileimagegetter = &DefaultImageStreamFromFileGetter{}
+	h.Filetemplategetter = &DefaultTemplateFromFileGetter{}
+	h.Filefinder = &DefaultResourceFileLister{}
 
 	h.imageclientwrapper = &defaultImageStreamClientWrapper{h: &h}
 	h.templateclientwrapper = &defaultTemplateClientWrapper{h: &h}
@@ -69,8 +68,7 @@ func NewHandler() sdk.Handler {
 	h.skippedImagestreams = make(map[string]bool)
 	h.skippedTemplates = make(map[string]bool)
 
-	h.mutex = &sync.Mutex{}
-	h.CreateDefaultResourceIfNeeded()
+	h.CreateDefaultResourceIfNeeded(nil)
 
 	h.imagestreamFile = make(map[string]string)
 	h.templateFile = make(map[string]string)
@@ -94,9 +92,9 @@ type Handler struct {
 	secretclientwrapper    SecretClientWrapper
 	configmapclientwrapper ConfigMapClientWrapper
 
-	fileimagegetter    ImageStreamFromFileGetter
-	filetemplategetter TemplateFromFileGetter
-	filefinder         ResourceFileLister
+	Fileimagegetter    ImageStreamFromFileGetter
+	Filetemplategetter TemplateFromFileGetter
+	Filefinder         ResourceFileLister
 
 	namespace string
 
@@ -105,145 +103,201 @@ type Handler struct {
 
 	imagestreamFile map[string]string
 	templateFile    map[string]string
-
-	deleteInProgress bool
-
-	mutex *sync.Mutex
 }
 
-func (h *Handler) prepWatchEvent(kind, name string, annotations map[string]string) (*v1alpha1.SamplesResource, string, bool) {
-	if h.deleteInProgress {
-		return nil, "", false
-	}
+// prepWatchEvent decides whether an upsert of the sample should be done, as well as data for either doing the upsert or checking the status of a prior upsert;
+// the return values:
+// - srcfg: return this if we want to check the status of a prior upsert
+// - filePath: used to the caller to look up the image content for the upsert
+// - doUpsert: whether to do the upsert of not ... not doing the upsert optionally triggers the need for checking status of prior upsert
+// - err: if a problem occurred getting the samplesresource, we return the error to bubble up and initiate a retry
+func (h *Handler) prepWatchEvent(kind, name string, annotations map[string]string) (*v1alpha1.SamplesResource, string, bool, error) {
 	srcfg, err := h.sdkwrapper.Get(v1alpha1.SamplesResourceName)
 	if srcfg == nil || err != nil {
-		logrus.Printf("Received watch event %s but ignoring since not have the SamplesResource yet: %#v %#v", kind+"/"+name, err, srcfg)
-		return nil, "", false
+		logrus.Printf("Received watch event %s but not upserting since not have the SamplesResource yet: %#v %#v", kind+"/"+name, err, srcfg)
+		return nil, "", false, err
 	}
+
+	if srcfg.ConditionFalse(v1alpha1.ImageChangesInProgress) {
+		// we do no return the srcfg in these cases because we do not want to bother with any progress tracking
+
+		if srcfg.DeletionTimestamp != nil {
+			logrus.Printf("Received watch event %s but not upserting since deletion of the SamplesResource is in progress and image changes are not in progress", kind+"/"+name)
+			return nil, "", false, nil
+		}
+		switch srcfg.Spec.ManagementState {
+		case operatorsv1alpha1api.Removed:
+			logrus.Debugf("Not upserting %s/%s event because operator is in removed state and image changes are not in progress", kind, name)
+			return nil, "", false, nil
+		case operatorsv1alpha1api.Unmanaged:
+			logrus.Debugf("Not upserting %s/%s event because operator is in unmanaged state and image changes are not in progress", kind, name)
+			return nil, "", false, nil
+		}
+	}
+
 	if annotations != nil {
 		isv, ok := annotations[v1alpha1.SamplesVersionAnnotation]
 		if ok && isv == v1alpha1.CodeLevel {
-			logrus.Debugf("ignoring %s/%s cause operator version matches", kind, name)
+			logrus.Debugf("Not upserting %s/%s cause operator version matches", kind, name)
 			// but return srcfg to potentially toggle pending condition
-			return srcfg, "", false
+			return srcfg, "", false, nil
 		}
 	}
-	proceed, err := h.ProcessManagementField(srcfg)
-	if !proceed || err != nil {
-		return nil, "", false
-	}
-
 	cm, err := h.configmapclientwrapper.Get(h.namespace, v1alpha1.SamplesResourceName)
 	if err != nil {
 		logrus.Warningf("Problem accessing config map during image stream watch event processing: %#v", err)
-		return nil, "", false
+		return nil, "", false, nil
 	}
 	filePath, ok := cm.Data[kind+"-"+name]
 	if !ok {
 		logrus.Printf("watch event %s not part of operators inventory", name)
-		return nil, "", false
+		return nil, "", false, nil
 	}
 
-	return srcfg, filePath, true
+	return srcfg, filePath, true, nil
 }
 
 func (h *Handler) processImageStreamWatchEvent(is *imagev1.ImageStream) error {
-	h.mutex.Lock()
-	defer h.mutex.Unlock()
+	// our pattern is the top most caller locks the mutex
 
-	srcfg, filePath, ok := h.prepWatchEvent("imagestream", is.Name, is.Annotations)
-	logrus.Debugf("prep watch event imgstr %s ok %v", is.Name, ok)
-	if !ok {
-		if srcfg != nil {
-			logrus.Debugf("checking tag spec/status for %s spec len %d status len %d", is.Name, len(is.Spec.Tags), len(is.Status.Tags))
-			// won't be updating imagestream this go around, which sets pending==true, so see if we should turn off pending
-			pending := false
-			if len(is.Spec.Tags) == len(is.Status.Tags) {
-				for _, specTag := range is.Spec.Tags {
-					matched := false
-					for _, statusTag := range is.Status.Tags {
-						logrus.Debugf("checking spec tag %s against status tag %s with num items %d", specTag.Name, statusTag.Tag, len(statusTag.Items))
-						if specTag.Name == statusTag.Tag {
-							for _, event := range statusTag.Items {
-								if specTag.Generation != nil {
-									logrus.Debugf("checking status tag %d against spec tag %d", event.Generation, *specTag.Generation)
-								}
-								if specTag.Generation != nil &&
-									*specTag.Generation <= event.Generation {
-									logrus.Debugf("got match")
+	srcfg, filePath, doUpsert, err := h.prepWatchEvent("imagestream", is.Name, is.Annotations)
+	logrus.Debugf("prep watch event imgstr %s ok %v", is.Name, doUpsert)
+	if !doUpsert {
+		if err != nil {
+			return err
+		}
+		if srcfg == nil {
+			return nil
+		}
+		processing := srcfg.Condition(v1alpha1.ImageChangesInProgress)
+		if processing.Status == corev1.ConditionFalse {
+			return nil
+		}
+
+		// so reaching this point means we have a prior upsert in progress, and we just want to track the status
+
+		logrus.Debugf("checking tag spec/status for %s spec len %d status len %d", is.Name, len(is.Spec.Tags), len(is.Status.Tags))
+		// won't be updating imagestream this go around, which sets pending==true, so see if we should turn off pending
+		pending := false
+		if len(is.Spec.Tags) == len(is.Status.Tags) {
+			for _, specTag := range is.Spec.Tags {
+				matched := false
+				for _, statusTag := range is.Status.Tags {
+					logrus.Debugf("checking spec tag %s against status tag %s with num items %d", specTag.Name, statusTag.Tag, len(statusTag.Items))
+					if specTag.Name == statusTag.Tag {
+						for _, event := range statusTag.Items {
+							if specTag.Generation != nil {
+								logrus.Debugf("checking status tag %d against spec tag %d", event.Generation, *specTag.Generation)
+							}
+							if specTag.Generation != nil &&
+								*specTag.Generation <= event.Generation {
+								logrus.Debugf("got match")
+								matched = true
+								break
+							}
+						}
+						if !matched {
+							// if an error occurred, let's give up as we are no longer "in progress"
+							// in that case as well
+							//TODO do we need yet another condition to report these errors separately from the
+							// in progress condition?
+							for _, condition := range statusTag.Conditions {
+								if condition.Status == corev1.ConditionFalse &&
+									len(condition.Message) > 0 &&
+									len(condition.Reason) > 0 {
+									logrus.Warningf("Image import for imagestream %s failed with reason %s and detailed message %s", is.Name, condition.Reason, condition.Message)
 									matched = true
 									break
 								}
 							}
+
 						}
 
-					}
-					if !matched {
-						pending = true
-						break
+						if matched {
+							break
+						}
 					}
 				}
-			}
-
-			logrus.Debugf("pending is %v for %s", pending, is.Name)
-
-			if !pending {
-				processing := srcfg.Condition(v1alpha1.ImageChangesInProgress)
-				now := kapis.Now()
-				// remove this imagestream name, including the space separator
-				logrus.Debugf("current reason %s ", processing.Reason)
-				processing.Reason = strings.Replace(processing.Reason, is.Name+" ", "", -1)
-				logrus.Debugf("reason now %s", processing.Reason)
-				if len(strings.TrimSpace(processing.Reason)) == 0 {
-					logrus.Debugf("reason empty setting to false")
-					processing.Status = corev1.ConditionFalse
-					processing.Reason = ""
+				if !matched {
+					pending = true
+					break
 				}
-				processing.LastTransitionTime = now
-				processing.LastUpdateTime = now
-				srcfg.ConditionUpdate(processing)
-				h.sdkwrapper.Update(srcfg)
-				return nil
 			}
-			// return error here so we retry the imagestream event
-			return fmt.Errorf("imagestream %s still in progress", is.Name)
+		}
+
+		logrus.Debugf("pending is %v for %s", pending, is.Name)
+
+		if !pending {
+			now := kapis.Now()
+			// remove this imagestream name, including the space separator
+			logrus.Debugf("current reason %s ", processing.Reason)
+			processing.Reason = strings.Replace(processing.Reason, is.Name+" ", "", -1)
+			logrus.Debugf("reason now %s", processing.Reason)
+			if len(strings.TrimSpace(processing.Reason)) == 0 {
+				logrus.Debugf("reason empty setting to false")
+				processing.Status = corev1.ConditionFalse
+				processing.Reason = ""
+			}
+			processing.LastTransitionTime = now
+			processing.LastUpdateTime = now
+			srcfg.ConditionUpdate(processing)
+			logrus.Debugf("SDKUPDATE no pending imgstr update")
+			return h.sdkwrapper.Update(srcfg)
 		}
 
 		return nil
 	}
 
-	imagestream, err := h.fileimagegetter.Get(filePath)
+	// prepWatchEvent has determined we actually need to do the upsert
+
+	imagestream, err := h.Fileimagegetter.Get(filePath)
 	if err != nil {
+		// still attempt to report error in status
 		h.processError(srcfg, v1alpha1.SamplesExist, corev1.ConditionUnknown, err, "%v error reading file %s", filePath)
+		logrus.Debugf("SDKUPDATE event img update err bad fs read")
+		h.sdkwrapper.Update(srcfg)
 		// if we get this, don't bother retrying
-		err = nil
+		return nil
 	} else {
 		err = h.upsertImageStream(imagestream, is, srcfg)
+		if err != nil {
+			h.processError(srcfg, v1alpha1.SamplesExist, corev1.ConditionUnknown, err, "%v error replacing imagestream %s", imagestream.Name)
+			logrus.Debugf("SDKUPDATE event img update err bad api obj update")
+			h.sdkwrapper.Update(srcfg)
+		}
+		return err
 	}
-	h.sdkwrapper.Update(srcfg)
-	return err
 
 }
 
 func (h *Handler) processTemplateWatchEvent(t *templatev1.Template) error {
-	h.mutex.Lock()
-	defer h.mutex.Unlock()
+	// our pattern is the top most caller locks the mutex
 
-	srcfg, filePath, ok := h.prepWatchEvent("template", t.Name, t.Annotations)
-	if !ok {
+	srcfg, filePath, doUpsert, err := h.prepWatchEvent("template", t.Name, t.Annotations)
+	if err != nil {
+		return err
+	}
+	if !doUpsert {
 		return nil
 	}
 
-	template, err := h.filetemplategetter.Get(filePath)
+	template, err := h.Filetemplategetter.Get(filePath)
 	if err != nil {
+		// still attempt to report error in status
 		h.processError(srcfg, v1alpha1.SamplesExist, corev1.ConditionUnknown, err, "%v error reading file %s", filePath)
+		logrus.Debugf("SDKUPDATE event temp udpate err")
+		h.sdkwrapper.Update(srcfg)
 		// if we get this, don't bother retrying
-		err = nil
+		return nil
 	} else {
 		err = h.upsertTemplate(template, t, srcfg)
+		if err != nil {
+			h.processError(srcfg, v1alpha1.SamplesExist, corev1.ConditionUnknown, err, "%v error replacing template %s", template.Name)
+			logrus.Debugf("SDKUPDATE event temp update err bad api obj update")
+			h.sdkwrapper.Update(srcfg)
+		}
+		return err
 	}
-	h.sdkwrapper.Update(srcfg)
-	return err
 
 }
 
@@ -322,7 +376,7 @@ func (h *Handler) VariableConfigChanged(srcfg *v1alpha1.SamplesResource, cm *cor
 		}
 	}
 
-	logrus.Printf("Incoming samplesresource unchanged from last processed version")
+	logrus.Debugf("Incoming samplesresource unchanged from last processed version")
 	return false, nil
 }
 
@@ -421,7 +475,7 @@ func (h *Handler) SpecValidation(srcfg *v1alpha1.SamplesResource) (*corev1.Confi
 
 	// if we have not had a valid SamplesResource processed, allow caller to try with
 	// the srcfg contents
-	if !srcfg.ConditionTrue(v1alpha1.SamplesExist) {
+	if !srcfg.ConditionTrue(v1alpha1.SamplesExist) && !srcfg.ConditionTrue(v1alpha1.ImageChangesInProgress) {
 		logrus.Println("Spec is valid because this operator has not processed a config yet")
 		return nil, nil
 	}
@@ -508,12 +562,10 @@ func (h *Handler) RemoveFinalizer(srcfg *v1alpha1.SamplesResource) {
 }
 
 func (h *Handler) NeedsFinalizing(srcfg *v1alpha1.SamplesResource) bool {
-	h.mutex.Lock()
-	defer h.mutex.Unlock()
-	if h.deleteInProgress {
+	if srcfg.ConditionFalse(v1alpha1.SamplesExist) {
 		return false
 	}
-	h.deleteInProgress = true
+
 	for _, f := range srcfg.Finalizers {
 		if f == v1alpha1.SamplesResourceFinalizer {
 			return true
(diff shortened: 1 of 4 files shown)


test/e2e/cluster_samples_operator_test.go
```diff
@@ -33,7 +37,41 @@ func TestClusterSamplesOperator(t *testing.T) {
...
+			return false, nil
+		}
+
+		return true, nil
+	})
+}
 
+func verifyImagesPresent(t *testing.T) {
+	// as the content will vary over time as openshift/library evolves
+	// sample a random set of imagestream related artifacts
```
bparees: you've got the content the operator is supposed to install in this repo, so you can reference it and verify it all gets installed.

test/e2e/cluster_samples_operator_test.go
```diff
@@ -53,4 +91,157 @@ func TestClusterSamplesOperator(t *testing.T) {
...
+		t.Errorf("creation times the same after delete: %v, %v, %#v", oldTime, sr.CreationTimestamp, sr)
+	}
+
+	err = verifyConditionsCompleteSamplesAdded(sr)
+	if err != nil {
+		t.Errorf("samples not re-established after delete %#v", sr)
+	}
+
+	verifyImagesPresent(t)
+	verifyTemplatesPresent(t)
```
bparees: should confirm the templates+imagestreams were also recreated since they should have been removed when you deleted the samples resource.

test/e2e/cluster_samples_operator_test.go
```diff
@@ -53,4 +91,157 @@ func TestClusterSamplesOperator(t *testing.T) {
...
+}
+
+func TestManagedState(t *testing.T) {
```
bparees: this is testing managed+removed states for the management state, so either rename appropriately(TestManagedAndRemovedState) or add testing for unmanaged state as well (and rename it to TestManagementState since you're testing the whole field)

bparees:
@gabemontero on further reflection last night i think the way you should do this is to add more .status fields to reflect your "creating content" state.

The flow would be:

1) operator sees it needs to create content
2) operator updates status to creating content
3) operator sees event indicated it status=creating content
4) operator ensures content is created
5) operator removes event indicating status=creating content

meanwhile if the resource is also being deleted/finalized, you can wait until the "creating content" status is removed before actually finalizing the resource (ie wait for everything to be created, and then immediately tear it down).

so to continue the flow/add steps in the middle:

3a) someone deletes the resource, so it is marked for deletion
3b) operator sees it is marked for deletion, but also sees status=creating content, so it does nothing
.....

6) operator sees status update event(or relist) such that creating content is not true and deletion is true, so it finalizes the resource.

This ensures that if the operator is restarted it can understand the state of the system and progress it correctly to the desired state w/o conflicting with other logic that is trying to create/delete resources.

gabemontero:
So the `e2e-aws-operator` run is the e2e tests here?

test/e2e/cluster_samples_operator_test.go
gabemontero: In my mind we were testing the `managementField` and varions valid values for it

Considering `TestSpecManagementStateField` given the prior name was unclear

And yes as I indicated in the description these tests in general are incomplete ... the needed delete processing changes I noted, which as I suspected has you rethinking the algorithm, resulted in my creating the PR now

pkg/stub/handler.go
```diff
@@ -898,6 +926,10 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
 			return err
 		}
 
+		// need to lock here some the imagestream/template watches don't cause
+		// conflicts on the samplesresource while we are deleting
+		h.mutex.Lock()
+		defer h.mutex.Unlock()
```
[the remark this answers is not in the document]
gabemontero: yeah the mutex reduces the need for conflict handling

the chicken / egg situation really stem from updating the samplesresource conditions when we get the secret/imagestream/templates events

i.e. a single controller/operator "managing" multiple api objects

As part of removing the mutex and delete* states, I'm beginning to lean toward bumping the relist interval down to 5 or 10 seconds, removing the template/imagestream/secret watches, and then on the relist of the samples resource selective list those resources and manage those changes that way.

Thoughts?

gabemontero:
@bparees - see my comment https://github.com/openshift/cluster-samples-operator/pull/54#discussion_r233946728 to respond/agree with/elaborate on your comment https://github.com/openshift/cluster-samples-operator/pull/54#issuecomment-439053232

pkg/stub/handler.go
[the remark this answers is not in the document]
bparees: > I'm beginning to lean toward bumping the relist interval down to 5 or 10 seconds, removing the template/imagestream/secret watches, and then on the relist of the samples resource selective list those resources and manage those changes that way.

i'm worried we'll get dinged on hammering the api server and have to revert...what's the motivation for that approach?

test/e2e/cluster_samples_operator_test.go
gabemontero: yeah your right ... just makes it more complicated :-)

I'll look at the load fixture utilities in the openshift/origin ext tests and see if that helps me figure out how to load the data via relative file lookups

pkg/stub/handler.go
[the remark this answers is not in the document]
gabemontero: motivation - what I noted above ... we currently have a multi write pattern ... a single watch creates a single write pattern

I hear the api server resiliency concern, but I also been told by those who would be concern they are not worried about lists .... agreed we'd need to consult again on this

I'll start with removing the mutex/delete* flags but leave the multiple watches and see how messed up it gets ...

test/e2e/cluster_samples_operator_test.go
gabemontero: Given the recreate is immediate in the delete case this proved difficult to do

Since the function which removes the samples is the same one used in the managementState: Removed scenario, I'm deferring revalidating here

pkg/stub/handler.go
```diff
@@ -751,43 +758,53 @@ func (h *Handler) CleanUpOpenshiftNamespaceOnDelete(srcfg *v1alpha1.SamplesResou
...
-			if err != nil && !kerrors.IsNotFound(err) {
-				logrus.Warnf("Problem deleteing openshift imagestream %s on SamplesResource delete: %#v", stream.Name, err)
+		if streamList.Items != nil {
+			for _, stream := range streamList.Items {
+				if _, ok := h.skippedImagestreams[stream.Name]; ok {
+					continue
+				}
+				err = h.imageclientwrapper.Delete("openshift", stream.Name, &metav1.DeleteOptions{})
+				if err != nil && !kerrors.IsNotFound(err) {
+					logrus.Warnf("Problem deleteing openshift imagestream %s on SamplesResource delete: %#v", stream.Name, err)
```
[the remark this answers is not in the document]
gabemontero: fixed in next push

gabemontero:
OK I've pushed a commit which removes the mutex and the associated changes which passed delete/remove testing, as well as some manual delete while creating (I'll add more automated tests in that vein later on)

Definitely more churn with error'ing on conflicts and getting events while still initially creating but the system stabilizes

Also a couple of minor updates per @bparees 's comments

Still have other pending comments from him ... but am next going to tacking removal of the delete* state in the operator and see about making various scenarios work ok

gabemontero:
net / pain

/retest

gabemontero:
OK I've been able to remove both the mutex and delete* state previously discussed @bparees 

I did not have to add additional conditions to do so

ptal when you get back from PTO

pending iteration on those changes, I'm going to circle back to the e2e's themselves and tackle the comments there as well as continue with the additions already planned ... if the additional tests surface another revisit of the operator implementation, I'll update this PR again even if the tests are not complete

test/e2e/cluster_samples_operator_test.go
gabemontero: yeah forgot for a sec about bindata in openshift/origin ... going to try to avoid that path, and do relative read

gabemontero:
OK pushed commit where I leverage the repo's content in order to build the list of templates and imagestreams to use during validation in the e2e's @bparees

gabemontero:
Still need e2e's for config settings/validation, mgmt disabled, status update with cluster operator (which requires of course pulling in the recent openshift/api updates ... so thinking that will be a separate PR)

pkg/stub/handler.go
```diff
@@ -105,21 +103,26 @@ type Handler struct {
...
-	if h.deleteInProgress {
-		return nil, "", false
-	}
 	srcfg, err := h.sdkwrapper.Get(v1alpha1.SamplesResourceName)
 	if srcfg == nil || err != nil {
 		logrus.Printf("Received watch event %s but ignoring since not have the SamplesResource yet: %#v %#v", kind+"/"+name, err, srcfg)
 		return nil, "", false
 	}
+	if srcfg.DeletionTimestamp != nil {
+		return nil, "", false
```
bparees: should probably log here too that you're ignoring the event.

pkg/stub/handler.go
```diff
@@ -186,64 +183,91 @@ func (h *Handler) processImageStreamWatchEvent(is *imagev1.ImageStream) error {
...
+			// the contition to true; so now, if in progress is false, but exists is also
+			// false, meaning this is part of the initial startup, we throw the error to
+			// so we process the pending condition after the startup completed
+			processing := srcfg.Condition(v1alpha1.ImageChangesInProgress)
+			exists := srcfg.Condition(v1alpha1.SamplesExist)
+			if !pending &&
+				processing.Status != corev1.ConditionTrue &&
+				exists.Status != corev1.ConditionTrue {
+				return fmt.Errorf("imagestream %s is no longer pending but creation loop still in progress", is.Name)
+			}
```
bparees: Can we simplify this down to "ignore imagestream events as long as SamplesExist is false"?

And when we start creating samples, we set ImageChangeInProgress to true?

pkg/stub/handler.go
```diff
@@ -561,23 +583,23 @@ func (h *Handler) IsRetryableAPIError(err error) bool {
 	return false
 }
 
-func (h *Handler) CreateDefaultResourceIfNeeded() (*v1alpha1.SamplesResource, error) {
+func (h *Handler) CreateDefaultResourceIfNeeded(srcfg *v1alpha1.SamplesResource) (*v1alpha1.SamplesResource, error) {
+	// assume the caller has call lock on the mutex .. out pattern is to have that as
+	// high up the stack as possible ... loc because need to
 	// coordinate with event handler processing
-	// where it will set h.sampleResource
+	// where it will set to false h.deleteInProgress
```
bparees: h.deleteInProgress isn't being set anymore

pkg/stub/handler.go
```diff
@@ -751,51 +773,61 @@ func (h *Handler) CleanUpOpenshiftNamespaceOnDelete(srcfg *v1alpha1.SamplesResou
...
 	} else {
-		for _, stream := range streamList.Items {
-			if _, ok := h.skippedImagestreams[stream.Name]; ok {
-				continue
-			}
-			err = h.imageclientwrapper.Delete("openshift", stream.Name, &metav1.DeleteOptions{})
-			if err != nil && !kerrors.IsNotFound(err) {
-				logrus.Warnf("Problem deleteing openshift imagestream %s on SamplesResource delete: %#v", stream.Name, err)
+		if streamList.Items != nil {
+			for _, stream := range streamList.Items {
```
bparees: we can't just delete every imagestream that's not in our skippedlist.  Users(admins) could have created additional imagestreams in the openshift namespace.  We need to check an annotation or something to confirm it was created/is-managed by us.

pkg/stub/handler.go
```diff
@@ -751,51 +773,61 @@ func (h *Handler) CleanUpOpenshiftNamespaceOnDelete(srcfg *v1alpha1.SamplesResou
...
-			}
-			err = h.templateclientwrapper.Delete("openshift", temp.Name, &metav1.DeleteOptions{})
-			if err != nil && !kerrors.IsNotFound(err) {
-				logrus.Warnf("Problem deleting openshift template %s on SamplesResource delete: %#v", temp.Name, err)
+		if tempList.Items != nil {
+			for _, temp := range tempList.Items {
+				if _, ok := h.skippedTemplates[temp.Name]; ok {
+					continue
+				}
+				err = h.templateclientwrapper.Delete("openshift", temp.Name, &metav1.DeleteOptions{})
```
bparees: same issue deleting templates..can only delete ones we own/manage/created

pkg/stub/handler.go
```diff
@@ -823,20 +855,20 @@ func (h *Handler) ProcessManagementField(srcfg *v1alpha1.SamplesResource) (bool,
...
 		}
-		return true, nil
+		return true, false, nil
 	case operatorsv1alpha1api.Unmanaged:
 		logrus.Println("management state set to unmanaged")
-		return false, nil
+		return false, false, nil
 	default:
-		return true, nil
+		return true, false, nil
```
bparees: seems like this should log+error since it's an unknown management state.

pkg/stub/handler.go
```diff
@@ -508,12 +532,10 @@ func (h *Handler) RemoveFinalizer(srcfg *v1alpha1.SamplesResource) {
 }
 
 func (h *Handler) NeedsFinalizing(srcfg *v1alpha1.SamplesResource) bool {
-	h.mutex.Lock()
-	defer h.mutex.Unlock()
-	if h.deleteInProgress {
+	if srcfg.ConditionFalse(v1alpha1.SamplesExist) {
```
bparees: samplesexist is set to true as soon as we start creating samples? or after all samples are created?

pkg/stub/handler.go
```diff
@@ -908,35 +943,67 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
...
-					h.CreateDefaultResourceIfNeeded()
+					h.CreateDefaultResourceIfNeeded(srcfg)
 				}()
-				return err
 			}
 			return nil
 		}
 
-		doit, err := h.ProcessManagementField(srcfg)
+		doit, sdkupdate, err := h.ProcessManagementField(srcfg)
```
bparees: sdkupdate doesn't seem like the right name for this...  srcfg update?

test/e2e/cluster_samples_operator_test.go
```diff
@@ -33,24 +46,296 @@ func TestClusterSamplesOperator(t *testing.T) {
...
 	})
 	if err != nil {
-		t.Errorf("error waiting for example imagestreamtag to appear: %v", err)
+		t.Errorf("creation times the same after delete: %v, %v, %#v", oldTime, sr.CreationTimestamp, sr)
+	}
+
+	err = verifyConditionsCompleteSamplesAdded(sr)
+	if err != nil {
+		t.Errorf("samples not re-established after delete %#v", sr)
 	}
```
bparees: how are you confirming the samples ever got removed?

pkg/stub/handler.go
bparees: I think it's a followup but i think there's still a window here either way.

The conditions you'd need to avoid a race are:

1) SampleCreationInProgress (we are iterating the imagestreams/templates and invoking create on each)
2) SamplesExist (we are done iterating them all, they are created now)

They should never both be true at the same time, and we should not attempt to finalize when (1) is true and (2) is not.

test/e2e/cluster_samples_operator_test.go
gabemontero: creation TS compare part of commit I just pushed

gabemontero:
OK I just pushed a commit for additions I had in flight today.  Specifics in the commit description.

I'll next circle back to @bparees 's comments for today Nov 19 and try to also include any comments he has on the commit I just pushed as they come in

When all those are sorted out, it is on to config settings/validations ... then that should be it for this PR

Next PR will be clusteroperator bump and tests

pkg/stub/handler.go
gabemontero: To answer you second question, yes, we set in progress to true when we start creating samples ... for example https://github.com/openshift/cluster-samples-operator/blob/master/pkg/stub/handler.go#L1065

To you first question, yeah I think you are right.  Change will be part of next push, unless something unexpected happens during testing.

pkg/stub/handler.go
gabemontero: The samplesresource with the samples exists condition set to true is persisted to etcd after the samples are created.
The flow either on startup, or after deletion, is:
- the create default logic creates the samplesresource with exists==false, calling the sdk to persist to etcd
- the watch event resulting from that creation comes in
- after the spec validation, mgmt state processing, skip filters creation, etc. we read the samples content from disk and upsert the samples to the openshift namespace
- if errors occur, exists become unknown, persist that to etc,
- is samples upsert successful, exists becomes true and in progress becomes true, and the sdk is called to update ... if conflict error, return error for sdk to relist/retry
- end of initial watch event
- the next watch event for the samples resource, as a result of setting exists,inprogress to true, comes in ... it is effectively ignored since it had no config changes
- the watch events for imagestreams and templates start coming in as soon as we create the samples ... they read in the sdk and will update if  the imagestream is complete (i.e. spec tag generations match status tag generations ... and they update the in progress condition accordingly ... if conflict, return error for retry

I don't agrees that exist==true and inprogress == true cannot occur at the same time

bparees: I was suggesting a new condition, so you'd have:

CreationInProgress (we are iterating the samples for creation)
SamplesExist (we have created everything we intend to create)
ChangesInProgress (we are waiting for imagestreams to import)

So SamplesExist and ChangesInProgress(ImportInProgress) could both be true.

But CreationInProgress and SamplesExist should be set exclusively (ie if at some point in the future we discover a sample is missing that we need to create, we'd set samplesexist back to false and creationinprogress back to true)

gabemontero: OK ... after our discussion and looking at the code, I realized that coupling the inprogress=true change and samplesexists=true change in the same samplesresource attempt like the code currently does is incorrect.

Not the first time I've learned the lesson and then had to re-learn it.

I need to set inprogress to true first, update the object, and then start creating the samples on the next watch event when inprogress is  true.

This essentially what @bparees referred to, but without creating the new condition.

Working on changes now ...

test/e2e/cluster_samples_operator_test.go
gabemontero: The call the `validateContent` with the `&now` param below calls `verifyImagesPresent` and `verifyTemplatesPresent` with that same timestamp ... those method test the CreationTime of the samples they find to confirm they were recreated

gabemontero:
@bparees - an intermediate update: ... I've been working off and on during PTO on the implementation changes around locking down concurrent create samples and (delete samples resource)/(mark samples resources mgmt state removed) we previously discussed.

In my testing to date, for the delete scenario, simply returning errors to initiate retry when conflicts occur has worked well.

However, in my testing to date for marking the mgmt state removed, simply returning errors when either the imagestream watch or samplesresource watch gets a conflict has not worked well.  Even after attempting to narrow the window on the samples resource fetch in the image stream watch, the timings have been, given the amount of imagestreams and presumably how quickly the update occurs from the samplesresource watch, that the system never settles down.  We may over many minutes get a few imagestreams removed from the reason field in the pending condition, but it is way too slow.

The timing on the api server side must be different enough between deleting the object and updating it ... seems plausible.

I believe I'm going to have to relent and add a new condition to help avert this.  Semantically, a "pending remove" condition of some sort when a remove is initiated in the middle of a creation cycle.  Once the creation cycle has completed and all the imagestreams have reported as being up to date, the changing of the pending condition to false can be seen by the samplesresource watch, and if "pending remove" is true, then initiate the remove.

I'll start coding this weekend, but of course chime in as desired if you don't like this approach, or want more elaboration on the details I noted above.

gabemontero:
OK the commit I just pushed should address @bparees 's last set of comments, my comment above from earlier this weekend ... includes the concurrent/delete timing windows ... it ended up being a refactor of some note ... I've got a detailed summary of changes in the commit text ... I'll post here as well:

- add remove on hold condition; 
- watch/delete/remove/in progress coordination fixes (do not delete while upserts in progress); 
- decouple samplesresrouce updates; 
- decouple filesystem read from samples creation; 
- init all conditions and finalizer on samplesresource create (finalizer on create helps with rapid/successive deletes); 
- filter out unmanaged samples on removes

Assuming we can get past major changes and just fine tune what I just pushed, I'll then 
get back to adding the additional tests 

Still view cluster operator status related stuff as separate PR

pkg/stub/handler.go
```diff
@@ -863,27 +962,29 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
...
-		defer h.mutex.Unlock()
 
 		srcfg, _ := h.sdkwrapper.Get(v1alpha1.SamplesResourceName)
 		if srcfg != nil {
-			doit, err := h.ProcessManagementField(srcfg)
-			if !doit || err != nil {
-				return err
+			switch srcfg.Spec.ManagementState {
+			case operatorsv1alpha1api.Removed:
+				logrus.Debugln("Ignoring secret event because operator is in removed state")
```
bparees: don't you still need to process the secret deleted event if the samples resource is in the removed state?  the earlier comments implied that was how the credentialsexist status would get updated.  (though you also need to ensure on a resync we can detect the credentials are missing and update the status, in case you miss the delete event)

(also it's the samples resource, not the operator, that is removed/unmanaged).

pkg/stub/handler.go
```diff
@@ -902,68 +1003,114 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
...
 			logrus.Info("A previous delete attempt has been successfully completed")
 			return nil
 		}
 		if srcfg.DeletionTimestamp != nil {
+			// before we kick off the delete cycle though, we make sure a prior creation
+			// cycle is not still in progress, because we don't want the create adding back
+			// in things we just deleted ... if a create is still in progress, return an
+			// error so we retry
+			if srcfg.ConditionTrue(v1alpha1.ImageChangesInProgress) {
+				return fmt.Errorf("Deletes will not be processed until outstanding creates are completed")
```
bparees: why do we need to retry?  won't we get another update event when the imagechangesinprogress status gets updated to false?

pkg/stub/handler.go
```diff
@@ -902,68 +1003,114 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
...
+				if !changed &&
+					srcfg.ConditionTrue(v1alpha1.SamplesExist) &&
+					srcfg.ConditionFalse(v1alpha1.ImageChangesInProgress) {
+					logrus.Debugf("Handle ignoring because config the same and exists is true and in progress false")
+					return nil
+				}
 			}
 		}
 
+		srcfg.Status.State = operatorsv1alpha1api.Managed
```
bparees: i don't see how this is guaranteed to be true?

pkg/stub/handler.go
```diff
@@ -751,92 +803,139 @@ func (h *Handler) CleanUpOpenshiftNamespaceOnDelete(srcfg *v1alpha1.SamplesResou
...
+		return err
 	}
 	return nil
 }
 
 // ProcessManagementField returns true if the operator should handle the SampleResource event
 // and false if it should not, as well as an err in case we want to bubble that up to
 // the controller level logic for retry
-func (h *Handler) ProcessManagementField(srcfg *v1alpha1.SamplesResource) (bool, error) {
+func (h *Handler) ProcessManagementField(srcfg *v1alpha1.SamplesResource) (processThisEvent, updateSamplesResource bool, err error) {
```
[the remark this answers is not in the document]
gabemontero: yeah i was just trying to clarify

pkg/stub/handler.go
```diff
@@ -187,63 +211,82 @@ func (h *Handler) processImageStreamWatchEvent(is *imagev1.ImageStream) error {
...
+					return h.sdkwrapper.Update(srcfg)
 				}
-				processing.LastTransitionTime = now
-				processing.LastUpdateTime = now
-				srcfg.ConditionUpdate(processing)
-				h.sdkwrapper.Update(srcfg)
-				return nil
+
+				// return error here so we retry the imagestream event
+				return fmt.Errorf("imagestream %s still in progress", is.Name)
```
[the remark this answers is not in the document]
gabemontero: it seemed to expedite / speed up processing but that was early on in all this ... I'll replace with returning nil and revisit

pkg/stub/handler.go
gabemontero: I took Removed as still being !Managed, or Unmanaged, but with additional effects.  Also, the removed processing removes the secret.  So if the user, after marking Removed, first adds the secret, and then deletes it, we would not have created the secret in the openshift namespace in the first place.  

Not sure what you exactly mean by "on a resync..."

Fixed debug string to say samples resource.

pkg/stub/handler.go
gabemontero: yeah I was hesitant to depend on the update from elsewhere .. but I believe you are correct

I'll try the change out, verify things still look OK

pkg/stub/handler.go
bparees: @gabemontero https://github.com/openshift/cluster-samples-operator/pull/54/files#diff-2d7ceed365f1ea47bba386dac808cb07R906

I interpret that comment as meaning the flow will be:

0) CredentialsExist=true, secret exists.
1) user sets spec.state=Removed
2) cleanupopenshiftnamespace gets run (this deletes the secret)
4) we will see the delete event for the secret
5) we will ignore it because src.spec.ManagementState is removed, so the CredentialsExist status never gets updated to false.

As for what I mean about on a resync, i mean suppose we miss the even in (4) (and assume you fix the logic in (5)).  If we miss the delete event, how does CredentialsExist ever get updated to false?

pkg/stub/handler.go
gabemontero: It corresponds with my interpretation that Removed is the same as Unmanaged and that we do not process any changes after this point.

The code above is always going to return before this point if we are Unmanaged or Removed.

pkg/stub/handler.go
gabemontero: OK ... @bparees and I talked 
- will fix comment
- will set cred exists condition to false when we delete since we are deleting both namespace secrets
- add a TODO for deciding what polling or other alternative is needed on missing secret delete event since we cannot add finalizer

bparees: we should mark the one in the operator namespace as to whether we created it or not (we will have created it if we created it by copying pull secrets from the installer).

and then we should delete it if we own it, otherwise leave it.

gabemontero:
OK pushed new commit for today's exchange @bparees

bparees:
new commit looks good, i'll take another full pass at it tomorrow.

gabemontero:
one more slight optimization while processing imagestream watch events and spec/status occurred to me last night ... should have additional commit up soon

gabemontero:
well the installer has decided to not come up this AM ... looks like I need to go through its refresh cycle in order to test my optimization

gabemontero:
ok installer back up after auth/scc's had me in the woods ... optimization pushed

gabemontero:
ok I think I've properly determined that our `samplesRegistry` implementation, aside form  the fact that https://github.com/openshift/cluster-samples-operator/pull/57 is needed to fix its serialization, is incorrect

It needs to account for the tag ref from kind before substituting

I'll be starting on a commit for that next ... it should be isolated from the other changes here so far, but adjust your scheduling you next full pass as you see fit given this @bparees 

*sigh*

I'm

bparees:
Please don't do it in this pr. Start another branch based off this one. It
will be easy to rebase that branch as needed if this one doesn't merge as
is and it'll reduce the churn on this pr.



Ben Parees | OpenShift

On Nov 27, 2018 4:09 PM, "Gabe Montero" <notifications@github.com> wrote:

ok I think I've properly determined that our samplesRegistry
implementation, aside form the fact that #57
<https://github.com/openshift/cluster-samples-operator/pull/57> is needed
to fix its serialization, is incorrect

It needs to account for the tag ref from kind before substituting

I'll be starting on a commit for that next ... it should be isolated from
the other changes here so far, but adjust your scheduling you next full
pass as you see fit given this @bparees <https://github.com/bparees>

*sigh*

I'm

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/openshift/cluster-samples-operator/pull/54#issuecomment-442217614>,
or mute the thread

gabemontero:
ok reverted samplesRegistry fix will stash / resubmit after this PR

pkg/stub/handler.go
```diff
@@ -105,145 +103,201 @@ type Handler struct {
...
+			logrus.Debugf("Not upserting %s/%s event because operator is in unmanaged state and image changes are not in progress", kind, name)
+			return nil, "", false, nil
+		}
+	}
+
 	if annotations != nil {
 		isv, ok := annotations[v1alpha1.SamplesVersionAnnotation]
 		if ok && isv == v1alpha1.CodeLevel {
-			logrus.Debugf("ignoring %s/%s cause operator version matches", kind, name)
+			logrus.Debugf("Not upserting %s/%s cause operator version matches", kind, name)
```
bparees: followup:  doesn't this mean i can modify the imagestream in any way i want, as long as i don't touch the SamplesVersionAnnotation?

test/e2e/cluster_samples_operator_test.go
```diff
@@ -33,24 +48,383 @@ func TestClusterSamplesOperator(t *testing.T) {
...
+			}
+
+			templates, _ := names[templatesKey]
+			templates[template.Name] = true
+		}
+	}
+	return names
+}
+
+func verifyImagesPresent(t *testing.T, content map[string]bool, timeToCompare *kapis.Time) {
```
bparees: ImageStreams

test/e2e/cluster_samples_operator_test.go
```diff
@@ -33,24 +48,383 @@ func TestClusterSamplesOperator(t *testing.T) {
...
+		t.Errorf("delete did not occur %v", err)
+	}
+	// now make sure it has not been recreated
+	err = wait.PollImmediate(1*time.Second, 30*time.Second, func() (bool, error) {
+		err := sdk.Get(is)
+		if err == nil {
+			return false, fmt.Errorf("jenkins imagestream recreated")
+		}
+		return false, nil
+	})
```
bparees: if your goal is to wait 30s and see if the imagestream got recreated, you might as well just do a time.Sleep(30*time.Second) and then check if it exists.

the wait.Poll isn't buying you anything if you're never going to exit early.

test/e2e/cluster_samples_operator_test.go
```diff
@@ -33,24 +48,383 @@ func TestClusterSamplesOperator(t *testing.T) {
...
+		t.Errorf("error updating samplesresource %v", err)
+	}
+
+	// wait for it to get into pending
+	err = wait.PollImmediate(1*time.Second, 30*time.Second, func() (bool, error) {
+		err := sdk.Get(sr)
+		if err != nil {
+			return false, err
+		}
+		if sr.ConditionTrue(samplesapi.ImageChangesInProgress) {
```
bparees: seems likely you could miss this condition if the imagestream gets imported fast enough. Seems likely to be a source of flakes.

bparees:
couple nits, one followup discussion, and one flaky test that needs to be addressed.

pkg/stub/handler.go
gabemontero: Yes that is correct.  This goes back to when I originally asked about checking for this, and asked if we wanted to compare the Spec structs, etc.  You said `no` to that, and suggested we employ the annotation.

test/e2e/cluster_samples_operator_test.go
gabemontero: part of next push

test/e2e/cluster_samples_operator_test.go
gabemontero: yep that is the goal ... will switch

test/e2e/cluster_samples_operator_test.go
gabemontero: Good catch.  I need to establish a timestamp before setting managed to true, and allow for either in progress being  true, or the last update time on the condition being greater than the pre-managed timestamp.

bparees:
@gabemontero squash and lets merge it.

gabemontero:
cool deal @bparees 

more e2e's to cover what was listed in the jira in a follow up PR

gabemontero:
Our e2e test ran up against the 10 minute limit:

`*** Test killed with quit: ran too long (10m0s).`

Hadn't seen that previously.  Would not think our squashing of commits would matter there.

And intermittent hang/error if the samples operator and/or tests?

Had to give up on libvirt today and just got a working aws cluster up.  I'll try some local runs to compare.

gabemontero:
The e2e test was waiting trying to verify that the sample templates were available in the openshift namespace when it got aborted.

gabemontero:
Hmmm ... running locally, I'm seeing complaints trying to create one of the templates:

```
time="2018-11-28T20:41:23Z" level=info msg="template update error: the server is currently unable to handle the request (put templates.template.openshift.io rails-postgresql-example)"
```

perhaps that is happening in e2e as well ....

gabemontero:
Local test against my aws cluster ran fine.  That said, it almost took 7 minutes, with all the waiting we do for image import.

The current 10 minute cap on e2e's might start becoming a barrier for our adding additional tests.  As well as intermittently delays in pulling images from docker hub.

bparees:
> Local test against my aws cluster ran fine. That said, it almost took 7 minutes, with all the waiting we do for image import.

is it really taking 7 minutes for the imports to complete, or is it taking 7 minutes for the operator to recognize it?   I would expect the image import to be very fast.  less than a minute for all the imagestreams to import.

> As well as intermittently delays in pulling images from docker hub.

nothing about importing requires pulling the images from docker hub

gabemontero:
On Wed, Nov 28, 2018 at 4:14 PM Ben Parees <notifications@github.com> wrote:

> Local test against my aws cluster ran fine. That said, it almost took 7
> minutes, with all the waiting we do for image import.
>
> is it really taking 7 minutes for the imports to complete, or is it taking
> 7 minutes for the operator to recognize it? I would expect the image import
> to be very fast. less than a minute for all the imagestreams to import.
>
I'd have to add debug in order the gauge the timing at those levels.  Will
add that to the list if my fear here comes to fruition.

> As well as intermittently delays in pulling images from docker hub.
>
> nothing about importing requires pulling the images from docker hub
>
Our openshift namesapce okd/centos imagestreams specs have DockerImage  tag
refs that point to docker hub content.  That is what I am referring to.
Does that not somewhere along the way involve interacting with docker hub?

bparees:
> Does that not somewhere along the way involve interacting with docker hub?

interacts yes, pulls the image, no.  it just grabs the metadata, which is significantly faster.

gabemontero:
Forgot ... the 10min number was the limit I had set on my poll, not the overall framework.  Most likely this was a template create hiccup like I saw once locally.

The test just passed and took 35min.

Need to figure out how to dump the pod logs from the current framework.  Took a quick stab last week but the log api from the sdk was not viable.  Maybe the sdk bump will help here ... otherwise, would need to bring in more k8s/origin deps to make it happen.

gabemontero:
`e2e-aws` had some networking, storage, and other unrelated failures by my interpretation @bparees 

manual merge?

PR #58, opened 2018-11-28
fix samplesRegistry processing; only update DockerImage kind tag refs

gabemontero:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1654402

QA is bumping into this now, so I want to get them this fix independent of the e2e PR @bparees


gabemontero:
same here @bparees re: e2e-aws and unrelated errors

PR #59, opened 2018-11-28
config validation tests

gabemontero:
fix to reset config valid to true when config corrected

add skipped lists tests


gabemontero:
e2e-aws failed test terraform setup not  route table pain

gabemontero:
Aside from the needed rebase, my local testing is uncovering some inconsistency with the skipped imagestream testing  ... some fixes in that space may be in the offing.

gabemontero:
As I suspected recently based on some local testing, occasionally random creates of templates fail:

`templates.template.openshift.io "dancer-mysql-example" not found`

Looks like I have an opportunity to examine possible retry in some fashion when this occurs, as well as adding more diagnostics in the conditions.

gabemontero:
My revisit of the code left me with believing the code should retry creation/update of any sample if there was an api level error.

So for now just pushed more diagnostics ... among other things, figured out how to dump the pod logs from the e2e.

bparees:
> My revisit of the code left me with believing the code should retry creation/update of any sample if there was an api level error.

though that's true, there also needs to be some more generic reconciliation that makes sure if we are missing an expected template, we'll attempt to create it again.  I thought we had that, but maybe because it's a 10 minute resync interval, it took too long for the test?

gabemontero:
We probably need to clarify what you mean by "reconciliation" @bparees 

My initial interpretation is that you mean "something periodically lists the samples in the openshift namespace and makes sure there is an entry for every sample we are carrying in the operator image based on configuration ... i.e. minus skips, ocp vs. okd, etc.".

If that is what you mean, then no, we don't have that.

If not that, please elaborate.

And certainly independent of the above, or in conjunction once we are at a common understanding, we can revisit the resync interval.

bparees:
> My initial interpretation is that you mean "something periodically lists the samples in the openshift namespace and makes sure there is an entry for every sample we are carrying in the operator image based on configuration ... i.e. minus skips, ocp vs. okd, etc.".
> If that is what you mean, then no, we don't have that.

yes, that's what i mean.  But there are really two cases where reconciliation matters:

1) periodically rechecking the state of things to ensure that if the user did something they weren't supposed to, or we missed an event, or something else, we can get the system back to the state it should be in.

2) ensuring that if something happens to our operator, the next time it comes up it is able to assess the state of the system, and get it to the intended state.

The two are obviously very very related, and if you have (1) you probably have (2) (but if you have (2), you may not have (1) unless you just kill the operator periodically).


What happens today if the operator crashes halfway through creating the templates?  Presumably SamplesExist is not set to true, and on startup the next operator instance will go through trying to create everything, ignoring/overwriting the ones that exist? 

What is happening when it gets an error creating one of the templates?

My concern with your proposal of adding a retry on a creation failure (though there is nothing wrong with it in terms of ensuring a quick retry in the face of temporary errors) is that:

1) we have to decide how long we retry for before giving up
2) once we give up, when do we ever try again?
3) if the operator crashes while we are in a retry state, how do we know to resume retrying when we come back up?

That's where a more full-fledged reconciliation model is needed, to answer (2) and (3).

gabemontero:
On Mon, Dec 3, 2018 at 3:38 PM Ben Parees <notifications@github.com> wrote:

> My initial interpretation is that you mean "something periodically lists
> the samples in the openshift namespace and makes sure there is an entry for
> every sample we are carrying in the operator image based on configuration
> ... i.e. minus skips, ocp vs. okd, etc.".
> If that is what you mean, then no, we don't have that.
>
> yes, that's what i mean. But there are really two cases where
> reconciliation matters:
>
ok thanks for confirming


>    1.
>
>    periodically rechecking the state of things to ensure that if the user
>    did something they weren't supposed to, or we missed an event, or something
>    else, we can get the system back to the state it should be in.
>    2.
>
>    ensuring that if something happens to our operator, the next time it
>    comes up it is able to assess the state of the system, and get it to the
>    intended state.
>
> The two are obviously very very related, and if you have (1) you probably
> have (2) (but if you have (2), you may not have (1) unless you just kill
> the operator periodically).
>
> What happens today if the operator crashes halfway through creating the
> templates? Presumably SamplesExist is not set to true, and on startup the
> next operator instance will go through trying to create everything,
> ignoring/overwriting the ones that exist?
>
yes

> What is happening when it gets an error creating one of the templates?
>
it returns the error to the operator sdk and the event should be requeued,
and a create/update samples retry should occur

> My concern with your proposal of adding a retry on a creation failure
> (though there is nothing wrong with it in terms of ensuring a quick retry
> in the face of temporary errors) is that:
>
>    1. we have to decide how long we retry for before giving up
>
> the sdk has an algorithm that will quit requeueing at some point

>
>    1. once we give up, when do we ever try again?
>
> on the resync, the condition state should be as such that we'll retry
again

we can certainly reduce the 10 minutes to say maybe 2 minutes.

>
>    1. if the operator crashes while we are in a retry state, how do we
>    know to resume retrying when we come back up?
>
>  upon reflection and revisiting the code I minimally need to fix the order
of creating the samples and setting in progress to true

after that change, I think we would still retry upon coming back up after a
crash in the retry state

that said, a reconciliation loop at first glance could fit in seamlessly
enough ... I'll look into it along with the change I noted above.

bparees:
> that said, a reconciliation loop at first glance could fit in seamlessly enough ... I'll look into it along with the change I noted above.

i'm not advocating an explicit reconciliation loop, reconciliation should be part of what we do on every resync interval.  ie we see the samples resource event (due to the resync/relist) and we confirm the system is in the desired state.

> we can certainly reduce the 10 minutes to say maybe 2 minutes.

not necessarily asking for that, i'm more concerned with ensuring we know we *will* get to where we want to be (or at least keep retrying forever because we recognize we aren't where we are supposed to be) than how often we do that.  I only mentioned the 10 mins to understand how the existing logic that should have solved this, might be interacting with your test which did not see it get fixed.

based on your answers above it sounds like we have everything we'd want in place, except for the ordering issue you mentioned.

gabemontero:
ok thanks for clarification on the explicit reconciliation piece

also, before I jump into the change I mentioned, I forgot to mention *why* I had the order like I did before ... creating the samples before setting in progress lead to some choppiness with the imagestream watches trying to update the samplesresource in progress as the samplesresrouce watch tried to set in progress to true

i think things have evolved such that we can avoid those conflicts when I switch this time around ... keep our fingers crossed.

gabemontero:
passing test now with the debug of course re: our intermittent missing template

gabemontero:
bootstrap failed

pkg/stub/handler.go
```diff
@@ -1117,6 +1127,13 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
 					logrus.Debugf("Handle ignoring because config the same and exists is true and in progress false")
 					return nil
 				}
+				// if config changed, but a prior config action is still in progress,
+				// return error so that we enact this config change against a steady
+				// state system ... think changes to skipped imagestreams and which
+				// imagestreams to account for in the in progress tracking
+				if changed && srcfg.ConditionTrue(v1alpha1.ImageChangesInProgress) {
+					return fmt.Errorf("this config change will not be operated on until in progress from the prior config setting becomes false")
```
gabemontero: rather than returning an error for retry, I'm considering setting in progress to false and clearing out the reason field ... then on the next event process the config change 

thoughts @bparees

gabemontero:
I've embedded one discussion point in the review comments @bparees ... ptal at these changes .. reflect recent fixes and today's discussion

pkg/stub/handler.go
bparees: I definitely prefer an approach where we don't have to queue up changes but rather can just apply the latest desired config to whatever state we are in.  you'd have to tell me, though, if that can potentially cause issues where we're in the middle of creating samples(or something) and then we process a new config that, say, changes the registry for all the samples.. will we end up in the right final state?

gabemontero:
ok ... clean tests runs ... ptal when you can @bparees

pkg/stub/handler.go
```diff
@@ -1092,13 +1116,19 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
...
 		if err != nil {
 			// flush status update
 			logrus.Debugf("SDKUPDATE bad spec validation update")
 			h.sdkwrapper.Update(srcfg)
 			return err
 		}
+		// if a bad config was corrected, update and return
+		if existingValidStatus != srcfg.Condition(v1alpha1.ConfigurationValid).Status {
+			logrus.Debugf("SDKUPDATE spec corrected")
+			return h.sdkwrapper.Update(srcfg)
```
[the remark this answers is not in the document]
gabemontero: yes that was the intent / that is why it is safe

in general, narrowing the amount of *change* with an event *transaction* has proven more stable in that it narrows conflict windows

gabemontero:
Yeah I'll squash ... kept separate commits to help me remember the motivation for particular changes and aid review

bparees:
/approve

feel free to self-lgm after squashing, assuming you can.

PR #61, opened 2018-11-29
Rebase clusteroperator status

gabemontero:
cluster operator status working again with new group and cluster scope

also moved from v1aplha1 to v1 api

as discussed on slack, the sdk could not properly update status on the clusteroperator status subresource, so using client-go there

note, the get *does* retrieve the entire clusteroperator obj ... so was able to add some basic verification in the e2e's


gabemontero:
well got a clean `e2e-aws-operator` run through ... `e2e-aws` is its usual wide range mess .. nothing I saw was related to the samples not being present.

bump @bparees after any change from your review IMO we green button this one

pkg/stub/handler.go
```diff
@@ -1624,6 +1629,24 @@ func (g *defaultConfigMapClientWrapper) Delete(namespace, name string) error {
...
 
+type defaultConfigClientWrapper struct {
+	h *Handler
+}
+
+func (g *defaultConfigClientWrapper) Get(name string) (*configv1.ClusterOperator, error) {
+	return g.h.configclient.ClusterOperators().Get(name, metav1.GetOptions{})
+}
+
+func (g *defaultConfigClientWrapper) Update(state *configv1.ClusterOperator) error {
```
bparees: UpdateStatus ?

test/e2e/cluster_samples_operator_test.go
```diff
@@ -80,6 +81,49 @@ func verifyConditionsCompleteSamplesRemoved(sr *samplesapi.SamplesResource) erro
...
 
+func verifyClusterOperatorConditionsComplete(t *testing.T) {
+	err := wait.PollImmediate(1*time.Second, 10*time.Minute, func() (bool, error) {
+		state := &configv1.ClusterOperator{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: configv1.SchemeGroupVersion.String(),
+				Kind:       "ClusterOperator",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name: samplesapi.SamplesResourceName,
```
bparees: looks like this should be named "openshift-cluster-samples-operator" if we want to align w/ the other ones that exist.

pkg/operatorstatus/operatorstatus.go
```diff
@@ -10,123 +10,133 @@ import (
...
+	return &handler
 }
 
 // OperatorStatusSDKWrapper is the interface that wrappers actuall access to github.com/operator-framework/operator-sdk/pkg/sdk
 type OperatorStatusSDKWrapper interface {
-	Get(name, namespace string) (*osapi.ClusterOperator, error)
-	Update(state *osapi.ClusterOperator) (err error)
-	Create(state *osapi.ClusterOperator) (err error)
+	Get(name string) (*configv1.ClusterOperator, error)
+	Update(state *configv1.ClusterOperator) (err error)
```
bparees: UpdateStatus ?

pkg/operatorstatus/operatorstatus.go
```diff
@@ -10,123 +10,133 @@ import (
...
 
 // NewCVOOperatorStatusHandler is the default initializer function for CVOOperatorStatusHandler
-func NewCVOOperatorStatusHandler(namespace string) *CVOOperatorStatusHandler {
-	return &CVOOperatorStatusHandler{SDKwrapper: &defaultOperatorStatusSDKWrapper{}, Namespace: namespace}
+func NewCVOOperatorStatusHandler(clientGoWrapper *OperatorStatusSDKWrapper) *CVOOperatorStatusHandler {
+	handler := CVOOperatorStatusHandler{}
+	if clientGoWrapper != nil {
+		handler.SDKwrapper = *clientGoWrapper
+	} else {
+		handler.SDKwrapper = &defaultOperatorStatusSDKWrapper{}
```
bparees: since this doesn't even work, when are we ever using it? can we delete this codepath?

pkg/stub/handler.go
```diff
@@ -1624,6 +1629,24 @@ func (g *defaultConfigMapClientWrapper) Delete(namespace, name string) error {
...
+	_, err := g.h.configclient.ClusterOperators().UpdateStatus(state)
+	return err
+}
+
+func (g *defaultConfigClientWrapper) Create(state *configv1.ClusterOperator) error {
+	_, err := g.h.configclient.ClusterOperators().Create(state)
+	return err
+}
```
bparees: seems like all of this should be moved up to where OperatorSDKHandler is defined.  (or that interface definition should move down here).

pkg/stub/handler.go
```diff
@@ -1624,6 +1629,24 @@ func (g *defaultConfigMapClientWrapper) Delete(namespace, name string) error {
 	return g.h.coreclient.ConfigMaps(namespace).Delete(name, &metav1.DeleteOptions{})
 }
 
+type defaultConfigClientWrapper struct {
```
bparees: should have ClusterOperator somewhere in the name since it only operators on ClusterOperator resources?

pkg/operatorstatus/operatorstatus.go
gabemontero: I kept it around with the reconsidering the sdk again when we upgraded the sdk in mind, and it also highlighted where I made the switch

but sure I can delete the codepath ... from the pointers from the network-edge team you got, the sdk api is different enough that keeping the old thing is less useful than I had anticipated

will delete

pkg/stub/handler.go
bparees: It can't be used generically because your function signatures aren't generic (ie "Get()" can only return a ClusterOperator).

pkg/stub/handler.go
gabemontero: I'll move the interface def from operatorstatus.go to handler.go

gabemontero: check that ... the other direction

pkg/stub/handler.go
gabemontero: Yeah but I can create new interfaces that leverage configclient but call something in place of `ClusterOperators()`

bparees: it's a struct, how are you going to create new interfaces?  If you want additional flavors of Get() you'd have to define another struct.

pkg/operatorstatus/operatorstatus.go
bparees: marked as resolved but i don't see a response or change?

pkg/stub/handler.go
gabemontero: I think I got hung up on the wrong line .... thought you were talking about the h.configclient and not the struct  (stupid forgetting my reading glasses today ;-)

I'll rename the struct

pkg/operatorstatus/operatorstatus.go
gabemontero: Sorry I'm updating this ... just forgot to post the comment

test/e2e/cluster_samples_operator_test.go
gabemontero: except for the machine api/config ones but yeah ... will rename

gabemontero:
updates based on your comments pushed @bparees

bparees:
squash and lgtm
/approve

gabemontero:
commits squashed ... adding lgtm

also, our e2e has had a few more instances of one of the template create/updates failing ...added some debug to the tests in the hope of getting the error message on the template operation if it occurs again

gabemontero:
passed this time

PR #64, opened 2018-12-05
add git based version information a la openshift/origin

gabemontero:
/assign @bparees

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM registry.svc.ci.openshift.org/openshift/release:golang-1.10 AS builder
 WORKDIR /go/src/github.com/openshift/cluster-samples-operator
 COPY . .
-RUN go build ./cmd/cluster-samples-operator
+RUN make build
 
 FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
 COPY --from=builder /go/src/github.com/openshift/cluster-samples-operator/cluster-samples-operator /usr/bin/
--- a/Dockerfile.rhel7
+++ b/Dockerfile.rhel7
@@ -1,7 +1,7 @@
 FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.10 AS builder
 WORKDIR /go/src/github.com/openshift/cluster-samples-operator
 COPY . .
-RUN go build ./cmd/cluster-samples-operator
+RUN make build
 
 FROM registry.svc.ci.openshift.org/ocp/4.0:base
 COPY --from=builder /go/src/github.com/openshift/cluster-samples-operator/cluster-samples-operator /usr/bin/
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ all: generate build build-image
 generate:
 	operator-sdk generate k8s
 build:
-	go build ./cmd/cluster-samples-operator
+	go build -ldflags '-X github.com/openshift/cluster-samples-operator/vendor/k8s.io/client-go/pkg/version.gitVersion=$(shell git describe) -X github.com/openshift/cluster-samples-operator/vendor/k8s.io/client-go/pkg/version.gitCommit=$(shell git rev-parse HEAD)' ./cmd/cluster-samples-operator
 build-image:
 	# save some time setting up the docker build context by deleting this first.
 	rm -f cluster-samples-operator        
--- a/pkg/apis/samplesoperator/v1alpha1/types.go
+++ b/pkg/apis/samplesoperator/v1alpha1/types.go
@@ -60,10 +60,10 @@ const (
 )
 
 var (
-	//TODO eventually we will want this off of the commit or something that correlates to versions in the code repo
 	// CodeLevel is a precise code version level that the operator will use to determine whether it ias
-	// been upgraded, and by extension, whether the samples should be updated.
-	CodeLevel = "v0.0.1"
+	// been upgraded, and by extension, whether the samples should be updated.  It will be set at the same
+	// time the ClusterOperator API Object for the samples operator is set.
+	CodeLevel string
 )
 
 type SamplesResourceSpec struct {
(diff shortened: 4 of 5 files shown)


Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM registry.svc.ci.openshift.org/openshift/release:golang-1.10 AS builder
 WORKDIR /go/src/github.com/openshift/cluster-samples-operator
 COPY . .
-RUN go build ./cmd/cluster-samples-operator
+RUN ./build-go.sh
```
bparees: i'm pretty sure, though not positive, that @smarterclayton doesn't want us customizing these dockerfiles

(though i do not know what that means for how we're supposed to inject git commit version info into our binaries)

bparees: (maybe customizing the first stage is ok and just not the second stage)

build-go.sh
```diff
@@ -0,0 +1,56 @@
...
+			else
+				OS_GIT_TREE_STATE="dirty"
+			fi
+		fi
+		# Use git describe to find the version based on annotated tags.
+		if [[ -n ${OS_GIT_VERSION-} ]] || OS_GIT_VERSION=$("${git[@]}" describe --long --tags --abbrev=7 --match 'v[0-9]*' "${OS_GIT_COMMIT}^{commit}" 2>/dev/null); then
+			# Try to match the "git describe" output to a regex to try to extract
+			# the "major" and "minor" versions and whether this is the exact tagged
+			# version or whether the tree is between two tagged versions.
+			if [[ "${OS_GIT_VERSION}" =~ ^v([0-9]+)\.([0-9]+)\.([0-9]+)(\.[0-9]+)*([-].*)?$ ]]; then
```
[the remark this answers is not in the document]
bparees: so these statements:
```
-X github.com/openshift/cluster-samples-operator/pkg/version.majorFromGit=${OS_GIT_MAJOR} \
	-X github.com/openshift/cluster-samples-operator/pkg/version.minorFromGit=${OS_GIT_MINOR} \
	-X github.com/openshift/cluster-samples-operator/pkg/version.versionFromGit=${OS_GIT_VERSION} \
```

do not error when those variables are undefined?  I guess maybe we used to use "set -u" in the origin script.

build-go.sh
```diff
@@ -0,0 +1,56 @@
...
+				# that is problematic since new untracked .go files affect the build,
+				# so use our idea of "dirty" from git status instead.
+				OS_GIT_VERSION+="-dirty"
+			fi
+		fi
+	fi
+
+	go build \
+	-ldflags "-X github.com/openshift/cluster-samples-operator/pkg/version.majorFromGit=${OS_GIT_MAJOR} \
+	-X github.com/openshift/cluster-samples-operator/pkg/version.majorFromGit=${OS_GIT_MAJOR} \
```
bparees: you're passing major twice

bparees: and not passing patch

pkg/version/version.go
```diff
@@ -0,0 +1,41 @@
...
+	// commitFromGit is a constant representing the source version that
+	// generated this build. It should be set during build via -ldflags.
+	commitFromGit string
+	// versionFromGit is a constant representing the version tag that
+	// generated this build. It should be set during build via -ldflags.
+	versionFromGit string
+	// major version
+	majorFromGit string
+	// minor version
+	minorFromGit string
```
bparees: should have a patch field too?

pkg/version/version.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// Get returns the overall codebase version. It's for detecting
+// what code a binary was built from.
+func Get() version.Info {
+	return version.Info{
+		Major:      majorFromGit,
+		Minor:      minorFromGit,
```
bparees: add patch

pkg/version/version.go
gabemontero: https://github.com/openshift/origin/blob/master/pkg/version/version.go and of the other version.go under origin/pkg didn't have a patch field

build-go.sh
gabemontero: so no setting of patch, or how to gather, in the openshift origin examples

removed duplicate major

build-go.sh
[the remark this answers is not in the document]
gabemontero: they did not error when I ran / tested

bparees:
ok, generally lgtm but want to hear from @smarterclayton to ensure this is the right pattern.

pkg/version/version.go
gabemontero: so yeah not inclined to add patch ... we can revisit if/when @smarterclayton chimes in and when you get back from PTO

In conjunction with not finding it in this space in openshift/origin, would have to research how to compile that information

gabemontero:
ah ... made my in line comment before seeing your response ^^

smarterclayton:
So there are two parts.

One, if you want your commit, set a version and tag (use describe ideally with a semantic up front version).

If you just need to know which payload you were part of (most people may want this as well) we’re going to add something there.

Technically we already know what commit you came from (it’s in the payload already).

Dockerfile
smarterclayton: It’s ok for customizing this line.  This line is yours

pkg/version/version.go
```diff
@@ -0,0 +1,41 @@
...
+	versionFromGit string
+	// major version
+	majorFromGit string
+	// minor version
+	minorFromGit string
+)
+
+// Get returns the overall codebase version. It's for detecting
+// what code a binary was built from.
+func Get() version.Info {
```
smarterclayton: Another option is to just set these into client-go and use that version.  I did that in the router and it simplifies behavior because client go already sends this info to the apiserver and you want that in audit.

See openshift/router makefile

bparees:
> One, if you want your commit, set a version and tag (use describe ideally with a semantic up front version).

what happens when we are being built out of dist-git?  (what will the git version/tags be?)

pkg/version/version.go
gabemontero: oooh the openshift/router way is cool .... switching to that

gabemontero:
pushed new commit that copies what @smarterclayton did for openshift/router

PR #65, opened 2018-12-06
Modify the secret name and some tiny typos

xiuwang:
The name of secret for openshift-cluster-samples-operator project should be '[samples-registry-credentials'](https://github.com/openshift/cluster-samples-operator/blob/master/pkg/apis/samplesoperator/v1alpha1/types.go#L40)

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1656689


gabemontero:
thanks @xiuwang !

PR #66, opened 2018-12-06
get centos to rhel transition working via managed to removed back to …

gabemontero:
…managed flow (with secret creation mixed in any step)

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1654626

Another opportunity for code deletion down the line if we end up only supporting the rhel install type.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ The samples resource offers the following configuration fields:
 - ManagementState
 -- Managed: the operator will update the samples as the configuration dictates
 -- Unmanaged: the operator will ignore updates to the samples resource object and any imagestreams or templates in the openshift namespace
--- Removed: the operator will remove the set of managed imagestreams and templates in the openshift namespace. It will ignore new samples created by the cluster admin or any samples in the skipped lists.  After the removals are complete, the operator will work like it is in the 'Unmanaged' state and ignore any watch events on the samples resources, secrets, imagestreams, or templates.  There are some caveats around concurrent creates and removal (see Change behaviors section).
+-- Removed: the operator will remove the set of managed imagestreams and templates in the openshift namespace. It will ignore new samples created by the cluster admin or any samples in the skipped lists.  After the removals are complete, the operator will work like it is in the 'Unmanaged' state and ignore any watch events on the sample resources, imagestreams, or templates.  It will operate on secrets to facilitate the 'centos' to 'rhel' switch.  There are some caveats around concurrent creates and removal (see Change behaviors section).
 - Samples Registry
 -- Override the registry that images are imported from
 - Install Type
@@ -46,8 +46,9 @@ The samples resource offers the following configuration fields:
 The install type and architectures are not allowed to be changed while in the 'Managed' state.
 
 In order to change the install type and architectures values, an administrator must:
-- Mark the ManagementState as 'Removed', saving the change
-- In a subsequent change, switch the Install Type or Architecture and change the ManagementState back to Managed
+- Mark the ManagementState as 'Removed', saving the change.
+- In a subsequent change, switch the install type or architecture and change the ManagementState back to Managed.
+- In the case of changing from 'centos' to 'rhel', the operator will still process Secrets while in Removed state.  You can create the secret either before switching to Removed, while in Removed state before switching to Managed state, or after switching to Managed state (though you'll see delays creating the samples until the secret event is processed if you create the secret after switching to Managed).
 
 ## Config behaviors
 
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -105,6 +105,7 @@ type Handler struct {
 	templateFile    map[string]string
 
 	imagestreamRetryCount map[string]int8
+	secretRetryCount      int8
 }
 
 // prepWatchEvent decides whether an upsert of the sample should be done, as well as data for either doing the upsert or checking the status of a prior upsert;
@@ -195,7 +196,7 @@ func (h *Handler) processImageStreamWatchEvent(is *imagev1.ImageStream) error {
 				retryCount = 0
 			}
 			retryCount++
-			if retryCount > 5 {
+			if retryCount > 3 {
 				h.imagestreamRetryCount[is.Name] = 0
 				// ignore this change of imagestream
 				logrus.Printf("retry count for imagestream event for %s exceeded; ignoring this change", is.Name)
@@ -606,9 +607,14 @@ func (h *Handler) CreateDefaultResourceIfNeeded(srcfg *v1alpha1.SamplesResource)
 	return srcfg, nil
 }
 
-func (h *Handler) WaitingForCredential(srcfg *v1alpha1.SamplesResource) error {
+// WaitingForCredential determines whether we should proceed with processing the sample resource event,
+// where we should *NOT* proceed if we are RHEL and using the default redhat registry;  The return from
+// this method is in 2 flavors:  1) if the first boolean is true, tell the caller to just return nil to the sdk;
+// 2) the second boolean being true means we've updated the samplesresource with cred exists == false and the caller should call
+// the sdk to update the object
+func (h *Handler) WaitingForCredential(srcfg *v1alpha1.SamplesResource) (bool, bool) {
 	if srcfg.ConditionTrue(v1alpha1.ImportCredentialsExist) {
-		return nil
+		return false, false
 	}
 
 	// if trying to do rhel to the default registry.redhat.io registry requires the secret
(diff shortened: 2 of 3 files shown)


gabemontero:
bump @bparees

pkg/stub/handler.go
```diff
@@ -812,6 +825,22 @@ func (h *Handler) ProcessManagementField(srcfg *v1alpha1.SamplesResource) (bool,
...
+					// initiate a secret event vs. doing the copy ourselves
+					go func() {
+						logrus.Println("updating operator namespace credential to initiate creation of credential in openshift namespace")
+						if secret.Annotations == nil {
+							secret.Annotations = map[string]string{}
+						}
+						// testing showed that we needed to change something for the update to actually go through
+						secret.Annotations[v1alpha1.SamplesRecreateCredentialAnnotation] = kapis.Now().String()
+						h.secretclientwrapper.Update(h.namespace, secret)
+					}()
```
bparees: why does this need to be done in a go routine?

pkg/stub/handler.go
```diff
@@ -862,13 +891,67 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
...
+				// - mark mgmt state removed
+				// - after that complete, edit again, mark install type to rhel and mgmt state to managed
+				// but what about the secret needed for rhel ... do we force the user to create the secret
+				// while still in managed/centos state?  Even with that, the "removed" action removes the
+				// secret the operator creates in the openshift namespace since it was owned/created by
+				// the operator
+				// So we allow the processing of the secret event while in removed state to
+				// facilitate the switch from centos to rhel, as necessitating use of removed as the means for
+				// changing from centos to rhel since  we allow changing the distribution once the samples have initially been created
+				logrus.Printf("processing secret watch event while in Removed state; deletion event: %v", event.Deleted)
```
bparees: should probably have a `case Managed:` and `default` for completeness (just logging what flow you went through)

pkg/stub/handler.go
```diff
@@ -606,23 +607,28 @@ func (h *Handler) CreateDefaultResourceIfNeeded(srcfg *v1alpha1.SamplesResource)
 	return srcfg, nil
 }
 
-func (h *Handler) WaitingForCredential(srcfg *v1alpha1.SamplesResource) error {
+func (h *Handler) WaitingForCredential(srcfg *v1alpha1.SamplesResource) (bool, error) {
```
bparees: godoc..i'm not clear on what the bool return value represents?  I assumed it meant "we are waiting for the credential"(true/false)" but you use it as "returnNil" below?

README.md
```diff
@@ -29,7 +29,7 @@ The samples resource offers the following configuration fields:
...
 -- Unmanaged: the operator will ignore updates to the samples resource object and any imagestreams or templates in the openshift namespace
--- Removed: the operator will remove the set of managed imagestreams and templates in the openshift namespace. It will ignore new samples created by the cluster admin or any samples in the skipped lists.  After the removals are complete, the operator will work like it is in the 'Unmanaged' state and ignore any watch events on the sample resources, secrets, imagestreams, or templates.  There are some caveats around concurrent creates and removal (see Change behaviors section).
+-- Removed: the operator will remove the set of managed imagestreams and templates in the openshift namespace. It will ignore new samples created by the cluster admin or any samples in the skipped lists.  After the removals are complete, the operator will work like it is in the 'Unmanaged' state and ignore any watch events on the sample resources, imagestreams, or templates.  It will operator on secrets to facilitate the 'centos' to 'rhel' switch.  There are some caveats around concurrent creates and removal (see Change behaviors section).
```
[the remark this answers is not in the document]
gabemontero: part of next push

pkg/stub/handler.go
```diff
@@ -606,23 +607,28 @@ func (h *Handler) CreateDefaultResourceIfNeeded(srcfg *v1alpha1.SamplesResource)
...
 	// if trying to do rhel to the default registry.redhat.io registry requires the secret
 	// be in place since registry.redhat.io requires auth to pull; since it is not ready
 	// log error state
 	if srcfg.Spec.InstallType == v1alpha1.RHELSamplesDistribution &&
 		(srcfg.Spec.SamplesRegistry == "" || srcfg.Spec.SamplesRegistry == "registry.redhat.io") {
+		cred := srcfg.Condition(v1alpha1.ImageChangesInProgress)
+		if len(cred.Message) > 0 {
+			// need for credential already registered
+			return true, nil
+		}
```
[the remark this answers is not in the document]
gabemontero: well the fact that there is a bug on line 620 does not help things :-) ... it should be looking at the message in the import credential condition, not the image in progress condition

In addition to fixing I'll expand on the comment along these lines:
- if import cred is false, and the message is emtpy, that means we have NOT registered the error, and need to do so
- if cred is false, and the message is there, we can just return nil to the sdk, to avoid additional churn;  once the secret comes in, it will update the samplesresource with cred == true, and then we'll get another samplesresource event that will trigger config processing

pkg/stub/handler.go
gabemontero: Yep will tie in with the comment I noted above

pkg/stub/handler.go
gabemontero: ok

pkg/stub/handler.go
gabemontero: it doesn't *HAVE* to be in a go routine, but there is not reason to gate returning to the sdk with this poking of the secret ... it is just a slight performance improvement

also I am augmenting the comment here as well with more details on why we are doing this

but if you really want it out of the go routine just let me know and I'll change it

bparees: Yeah my inclination would be to drop the go routine in favor of less complexity and more deterministic flows.

gabemontero:
ok pushed additional commit for @bparees 's first set of comments and rebased to pick up the README cleanup from QE

pkg/stub/handler.go
```diff
@@ -732,6 +749,11 @@ func (h *Handler) CleanUpOpenshiftNamespaceOnDelete(srcfg *v1alpha1.SamplesResou
 		}
 	}
 
+	if srcfg.ConditionTrue(v1alpha1.ImportCredentialsExist) {
+		logrus.Println("Operator is deleting the credential in the openshfift namespace that was previously created.")
```
bparees: s/openshfift/openshift/

pkg/stub/handler.go
```diff
@@ -606,23 +607,36 @@ func (h *Handler) CreateDefaultResourceIfNeeded(srcfg *v1alpha1.SamplesResource)
...
+		// - if cred is false, and the message is there, we can just return nil to the sdk, which "true" for the boolean return value indicates;
+		// not returning the same error multiple times to the sdk avoids additional churn; once the secret comes in, it will update the samplesresource
+		// with cred == true, and then we'll get another samplesresource event that will trigger config processing
+		if len(cred.Message) > 0 {
+			return true, nil
+		}
 		err := fmt.Errorf("Cannot create rhel imagestreams to registry.redhat.io without the credentials being available")
 		h.processError(srcfg, v1alpha1.ImportCredentialsExist, corev1.ConditionFalse, err, "%v")
-		return err
+		return false, err
```
bparees: make this `return true, err`

pkg/stub/handler.go
```diff
@@ -1005,7 +1101,10 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
 		// if trying to do rhel to the default registry.redhat.io registry requires the secret
 		// be in place since registry.redhat.io requires auth to pull; if it is not ready
 		// error state will be logged by WaitingForCredential
-		err = h.WaitingForCredential(srcfg)
+		returnNil, err := h.WaitingForCredential(srcfg)
+		if returnNil {
+			return nil
+		}
```
bparees: then make this logic:

```
waiting,err:=h.WaitingForCredential(srcfg)
if waiting && err!=nil {
      // if err!=nil, then srccfg was updated with an error condition, so we need to call sdk update
      logrus.Println("samplesresource update ignored since need the RHEL credential")
      return h.sdkwrapper.Update(srcfg)
}
if waiting {
  // we're waiting, but srcfg has not changed, so no-op
  return nil
}
// process
```

bparees: it seems easier to reason about than "returnNil"

bparees: you could even make waitingForCredential just return bool,bool  (waiting,srcfgModified) since you are not actually using the err object that's being returned other than checking for nil on it.

pkg/stub/handler.go

pkg/stub/handler.go
gabemontero: I assume this goes in conjunction with your comment below ...

gabemontero: ah yeah ... your next comment starts with `then make ...`

bparees:
squash and lgtm

PR #67, opened 2018-12-14
fix managed sample recreation after delete of said sample

gabemontero:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1658898

/assign @bparees


gabemontero:
image eco env fail

Issue #69, opened 2018-12-17
clusteroperator condition message improvements


gabemontero:
slack exchange:

```
i’d like to get some style cleanups in both the devex team operators on the progressing message
this is your primary communication channel and the intended receiver is the admin
right now it’s “The samples operator configuration is valid” and for registry it’s something slightly more machine oriented
i’d like the message to be something more affirmative like “The latest sample images and templates are installed” or something
or even shorter “All sample resources are up to date”
the style is sentence without punctuation
upper case leading, no punctution, human intended receipient
try to avoid kube-isms and overly technical explanations
Gabe Montero [1:41 PM]
the available condition is the one that says "Samples exist in the openshift project"
Clayton Coleman [1:42 PM]
Progressing is the generic condition
Gabe Montero [1:42 PM]
the failing condition is the one that says ""The samples operator configuration is valid""
Clayton Coleman [1:43 PM]
that’s where you put your human focused summary message
Progressing must represent a summary of the other conditions
so it’s fine to have the same message on progressing as available
(it’s fine to have the same messages on available)
but available and failed should explain the details as necessary, while progressing should explain to a human what the current state is
```

@bparees FYI in general

@dmage @legionus @coreydaley FYI re: @smarterclayton wants tweaks to the image-registry messages as well

smarterclayton:
For context see https://github.com/openshift/cluster-version-operator/blob/master/pkg/cvo/status.go which has comments and examples of how the CVO works which is what this pattern is modeled on.

smarterclayton:
https://github.com/openshift/cluster-version-operator/pull/72 is docs around what to do

PR #70, opened 2018-12-18
Adjust clusteroperator msgs to https://github.com/openshift/cluster-v…

gabemontero:
…ersion-operator/blob/master/docs/dev/clusteroperator.md#what-should-an-operator-report-with-clusteroperator-custom-resource

Fixes https://github.com/openshift/cluster-samples-operator/issues/69

@bparees @smarterclayton

### DIFF
--- a/pkg/apis/samplesoperator/v1alpha1/types.go
+++ b/pkg/apis/samplesoperator/v1alpha1/types.go
@@ -190,7 +190,7 @@ const (
 	// tag match.  The list of imagestreams that are still in progress will be stored
 	// in the Reason field of the condition.  The Reason field being empty corresponds
 	// with this condition being marked true.
-	ImageChangesInProgress SamplesResourceConditionType = "ChangesInProgress"
+	ImageChangesInProgress SamplesResourceConditionType = "ImageChangesInProgress"
 	// RemovedManagementStateOnHold represents whether the SamplesResource ManagementState
 	// has been set to Removed while a samples creation/update cycle is still in progress.  In other
 	// words, when ImageChangesInProgress is True.  We
@@ -338,16 +338,17 @@ func (s *SamplesResource) ClusterOperatorStatusAvailableCondition() (configv1.Co
 		return falseRC, falseMsg
 	}
 
-	// REMINDER: the intital config is always valid; only config changes after
+	// REMINDER: the intital config is always valid, as this operator generates it;
+	// only config changes after by a human cluster admin after
 	// the initial install result in ConfigurationValid == CondtitionFalse
 	// Next, if bad config is injected after installing at a certain level,
 	// the samples are still available at the old config setting; the
 	// config issues will be highlighted in the progressing/failing messages, per
 	// https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#conditions
 
-	// However, rhel and lack of creds is possible on intitial install, and if the
-	// creds are deleted after the initial install, it can prevent imagestream
-	// scheduled imports for example to fail ... the imagesstream "state" could
+	// However, rhel and lack of creds is possible on intitial install, as well
+	// as deleted after the initial install; either circumstance can prevent imagestream
+	// scheduled imports for example to fail ... the imagestream "state" could
 	// be sufficiently compromised, so we'll flag false there
 	if needCreds {
 		return falseRC, falseMsg
@@ -395,7 +396,7 @@ func (s *SamplesResource) ClusterOperatorStatusFailingCondition() (configv1.Cond
 	// condition, either api server interaction or file system interaction;
 	// Conversely, those errors result in a ConditionUnknown setting on one
 	// of the conditions;
-	// but if for some reason that ever changes, we'll need to adjust this
+	// If for some reason that ever changes, we'll need to adjust this
 	if s.AnyConditionUnknown() {
 		return trueRC, "bad API object operation", s.ConditionsMessages()
 	}
--- a/pkg/apis/samplesoperator/v1alpha1/types.go
+++ b/pkg/apis/samplesoperator/v1alpha1/types.go
@@ -1,9 +1,13 @@
 package v1alpha1
 
 import (
+	"fmt"
+
+	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/pkg/version"
 )
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
@@ -64,12 +68,21 @@ const (
 	SamplesRecreateCredentialAnnotation = GroupName + "/recreate"
 )
 
-var (
-	// CodeLevel is a precise code version level that the operator will use to determine whether it ias
-	// been upgraded, and by extension, whether the samples should be updated.  It will be set at the same
-	// time the ClusterOperator API Object for the samples operator is set.
-	CodeLevel string
-)
+func GitVersionString() string {
+	vinfo := version.Get()
+	versionString := "4.0.0-alpha1-"
+	switch {
+	case len(vinfo.GitVersion) > 0:
+		versionString = string(vinfo.GitVersion) + "-"
+		fallthrough
+	case len(vinfo.GitCommit) > 0:
+		c := string(vinfo.GitCommit)[0:9]
+		versionString = versionString + c
+	default:
+		versionString = "4.0.0-was-not-built-properly"
+	}
+	return versionString
+}
 
 type SamplesResourceSpec struct {
 	// ManagementState is top level on/off type of switch for all operators.
@@ -107,6 +120,11 @@ type SamplesResourceSpec struct {
 	// content but the operator will not recreate(or update) anything
 	// listed here.
 	SkippedTemplates []string `json:"skippedTemplates,omitempty" protobuf:"bytes,6,opt,name=skippedTemplates"`
+
+	// Version is the value of the operator's git based version indicator when the SamplesResource is being processed.
+	// The operator will use it to determine whether it has
+	// been upgraded, and by extension, whether the samples should be updated.
+	Version string `json:"version,omitempty" protobuf:"bytes,7,opt,name=version"`
 }
 type SamplesResourceStatus struct {
 	// operatorv1.ManagementState reflects the current operational status of the on/off switch for
@@ -146,6 +164,9 @@ type SamplesResourceStatus struct {
 	// content but the operator will not recreate(or update) anything
 	// listed here.
 	SkippedTemplates []string `json:"skippedTemplates,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,7,rep,name=skippedTemplates"`
+
+	// Version is the value of the operator's git based version indicator when it was last successfully processed
+	Version string `json:"version,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,8,rep,name=version"`
 }
 
 type SamplesResourceConditionType string
@@ -180,6 +201,15 @@ const (
 	// a retry when ManagementState was set to Removed lead to a prolonged, sometimes seemingly unresolved,
 	// period of circular contention
 	RemovedManagementStateOnHold SamplesResourceConditionType = "PendingRemove"
+	// MigrationInProgress represents the special case where the operator is running off of
+	// a new version of its image, and samples are deployed of a previous version.  This condition
+	// facilitates the maintenance of this operator's ClusterOperator object.
+	MigrationInProgress SamplesResourceConditionType = "MigrationInProgress"
+	// ImportImageErrorsExist registers any image import failures, separate from ImageChangeInProgress,
+	// so that we can a) indicate a problem to the ClusterOperator status, b) mark the current
+	// change cycle as complete in both ClusterOperator and SamplesResource; retry on import will
+	// occur by the next relist interval if it was an intermittent issue;
+	ImportImageErrorsExist SamplesResourceConditionType = "ImportImageErrorsExist"
 )
 
 // SamplesResourceCondition captures various conditions of the SamplesResource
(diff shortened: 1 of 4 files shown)


gabemontero:
getting ci setup errors so far, but have some changes coming based on local testing

/hold

gabemontero:
yeah @bparees @smarterclayton don't bother reviewing until I push my next commit and announce this is viable

pkg/operatorstatus/operatorstatus.go
```diff
@@ -59,40 +59,47 @@ type ClusterOperatorWrapper interface {
...
 	case sampleResource.ConditionTrue(v1alpha1.SamplesExist):
-		err = o.setOperatorStatus(configv1.OperatorAvailable, configv1.ConditionTrue, "Samples exist in the openshift project")
+		err = o.setOperatorStatus(configv1.OperatorAvailable, configv1.ConditionTrue, fmt.Sprintf("Samples installed at %s", v1alpha1.CodeLevel))
 	case sampleResource.ConditionFalse(v1alpha1.SamplesExist):
-		err = o.setOperatorStatus(configv1.OperatorAvailable, configv1.ConditionFalse, "Samples do not exist in the openshift project")
+	case sampleResource.ConditionUnknown(v1alpha1.SamplesExist):
+		err = o.setOperatorStatus(configv1.OperatorAvailable, configv1.ConditionFalse, fmt.Sprintf("Samples not installed at %s", v1alpha1.CodeLevel))
 	default:
-		err = o.setOperatorStatus(configv1.OperatorAvailable, configv1.ConditionUnknown, "The presence of the samples in the openshift project is unknown")
+		err = o.setOperatorStatus(configv1.OperatorAvailable, configv1.ConditionUnknown, fmt.Sprintf("Samples at %s unknown", v1alpha1.CodeLevel))
```
bparees: i don't love this one but i have nothing better to suggest at the moment.

pkg/operatorstatus/operatorstatus.go
```diff
@@ -59,40 +59,47 @@ type ClusterOperatorWrapper interface {
...
 	case sampleResource.ConditionFalse(v1alpha1.ConfigurationValid):
-		err = o.setOperatorStatus(configv1.OperatorFailing, configv1.ConditionTrue, "The samples operator configuration is invalid")
+	case sampleResource.ConditionUnknown(v1alpha1.ImageChangesInProgress):
+	case len(errorDetails) > 0:
+		err = o.setOperatorStatus(configv1.OperatorFailing, configv1.ConditionTrue, fmt.Sprintf("Unable to install samples at %s: %s", v1alpha1.CodeLevel, errorDetails))
+		if err != nil {
+			return err
+		}
+		// on errors we stop after this setting
+		return o.setOperatorStatus(configv1.OperatorProgressing, configv1.ConditionTrue, fmt.Sprintf("Samples unable to reach %s", v1alpha1.CodeLevel))
```
bparees: is it "unable to reach" or is it "Samples are updating to....."?  if it's "unable to reach" i'd expect us to be in failing=true and progressing=false, but this is setting progressing=true.

bparees: (did you mean to be setting the Failing condition here instead of Progressing?)

bparees:
> yeah @bparees @smarterclayton don't bother reviewing until I push my next commit and announce this is viable

sorry, didn't see this in time :)

pkg/operatorstatus/operatorstatus.go
gabemontero: Yeah in fact in the changes I noted in https://github.com/openshift/cluster-samples-operator/pull/70#issuecomment-448399599 I got rid of this ... but we can certainly revisit when I push the new commit

gabemontero:
On Tue, Dec 18, 2018 at 5:56 PM Ben Parees <notifications@github.com> wrote:

> yeah @bparees <https://github.com/bparees> @smarterclayton
> <https://github.com/smarterclayton> don't bother reviewing until I push
> my next commit and announce this is viable
>
> sorry, didn't see this in time :)
>

no worries ... I was a bit tardy in posting it ... and as a suggestion,
read
and cross reference with @smarterclayton 's examples at
https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#conditions

Hopefully it helps explain some of the choices made here (assuming you and
I interpret that doc the same way).

gabemontero:
OK the installer (both aws and libvirt) quit working for me between the office and home, so I couldn't finish testing, but I think I got far enough along to at least start reviewing.

Hopefully I can test tomorrow AM in conjunction with iterating over any review we get to over PTO (which starts tomorrow PM for me, but I plan on spending some cycles over PTO).

@bparees ^^

gabemontero:
got cluster install working again @bparees but have found some things that need to fix ... so please consider "paused" again until I update

gabemontero:
OK things look good in my local testing now.  Both initial startup and release upgrade flows.

@bparees ptal whenever you have the chance

pkg/operatorstatus/operatorstatus.go
```diff
@@ -59,44 +58,63 @@ type ClusterOperatorWrapper interface {
...
+	// errorDetails should have data if config valid false or progress unknown, but just in case
 	case sampleResource.ConditionFalse(v1alpha1.ConfigurationValid):
-		err = o.setOperatorStatus(configv1.OperatorFailing, configv1.ConditionTrue, "The samples operator configuration is invalid")
+	case sampleResource.ConditionUnknown(v1alpha1.ImageChangesInProgress):
+	case len(errorDetails) > 0:
+		err = o.setOperatorStatus(configv1.OperatorFailing, configv1.ConditionTrue, fmt.Sprintf("Unable to install samples at %s: %s", sampleResource.Spec.Version, errorDetails), "")
+		if err != nil {
+			return err
+		}
+		err = o.setOperatorStatus(configv1.OperatorAvailable, configv1.ConditionFalse, fmt.Sprintf("Unable to install samples at %s", sampleResource.Spec.Version), "")
```
bparees: is available really guaranteed to be false here?  

If configurationvalid==false but we previously had a valid configuration, then we would have (may have?) already installed the samples so things are "Available" even if we can't progress to the new desired state.  Basically if "samplesexist==true" && "imagechangesinprogress==false" then i think available should be true?


if imagechangesinprogress==true then i guess i agree, it's not available.

pkg/operatorstatus/operatorstatus.go
```diff
@@ -59,44 +58,63 @@ type ClusterOperatorWrapper interface {
...
+		}
+		briefStateDescription := ""
+		switch {
+		case sampleResource.ConditionFalse(v1alpha1.ConfigurationValid):
+			briefStateDescription = "invalid configuration"
+		case sampleResource.ConditionFalse(v1alpha1.ImportCredentialsExist) && sampleResource.Spec.InstallType == v1alpha1.RHELSamplesDistribution:
+			briefStateDescription = "image pull credentials needed"
+		}
+		// on errors we stop after this setting
+		return o.setOperatorStatus(configv1.OperatorProgressing, configv1.ConditionTrue, fmt.Sprintf("Unable to install samples at %s: %s", sampleResource.Spec.Version, briefStateDescription), "")
```
bparees: I also think this whole block confuses the issue a bit... you've got a block above that's solely responsible for setting Available, and a block below that's soley responsible for for setting Progressing, but then this block, which seems like it is mostly responsible for setting Failing, also contains logic for setting Available+Progressing in certain cases.

Can this be refactored so that all the logic for setting Progressing is in one place, Available in another, and Failing in a third, rather than intermingling it?

gabemontero: ah dammit ... I meant to put a //TODO comment explaining some things around this, but then forgot as I was rushing to get out of the office.

Yeah, while this functionally works, this is not the final form.  It was simply the form with the least amount change.  It leverages the current side effect of the implementation that the `Messages` field is only set on errors.

 There are a couple of probably better choices, but they involved a larger change, and I wanted to discuss and got consensus on which choice before making the investment on the change.

The 3 condition choices are `ConditionTrue`, `ConditionFalse`, and `ConditionUnknown`.  Per the godoc, "ConditionUnknown" means kubernetes can't decide if a resource is in the condition or not.  Whether "errors" should be `ConditionFalse` or `ConditionUnknown` has been a back and forth with me.  The current mixture of use 

One possible choice would be to consider *all* errors, whether we are unsure, or absolutely sure, on the errors, with a `ConditionUnknown` setting.  Using `ConditionFalse` muddies the waters, because distinguishing between error state and uninitialized or pre steady state becomes problematic.  But getting consistent on the condition value, the bloc you mentioned we be refactored along checking for `ConditionUnknown` lines.

Another choice is to explicitly add a "Failing" condition, and use it to map to the failing condition, and redo the block above along those lines.

Thoughts?

pkg/operatorstatus/operatorstatus.go
gabemontero: You are right on an upgrade scenario it could still be true at the older level.  I'll make a note of covering this with the next iteration choice we make below.

bparees: it's not just on the upgrade scenario.  I could also start out in a valid config, get everything created, and then change the config to something invalid, right?

pkg/operatorstatus/operatorstatus.go
bparees: I think ConditionUnknown and ConditionFalse are orthogonal to errors.  For us it is up to us and what our operator can determine about the state of the system.

If we are sure the sample exist, available=true.  if we're sure they don't exist (because we know we failed to create them for some reason or are blocked from creating them, or deleted them), available=false.  If we got an error part way into creating or deleting them, then we're in muddy waters where we should either be setting it to false (if we're confident we didn't manage to create everything) or unknown (if we really can't be sure if the creation/deletion went through).

We probably need to create a truth table that goes through the various samplesresource condition combinations, and then what the clusteroperator conditions should be given those samplesresource conditions.  (which is basically what you've constructed in your switch statement, but perhaps it we write it out formally somewhere things will become clearer).

> Whether "errors" should be ConditionFalse or ConditionUnknown has been a back and forth with me

just to elaborate on this point:  an error could result in a conditiontrue also (such as in the case of the Failing condition).  So i view the error as "additional information about why we are in the state we are reporting", the error itself does not determine what state we are in, it provides context to why we are reporting that state.

pkg/operatorstatus/operatorstatus.go
bparees: (at which point the samples created by the original valid config would still exist, so we'd be in available=true, progressing=false, failing=true)

gabemontero: yes I did not mean to imply the upgrade scenario was the exhaustive list

pkg/operatorstatus/operatorstatus.go
bparees: > Next, samples exists==true != available==true.

right, i forgot about the need to actually do the import/have the import succeed.  my bad. I don't think it alters my larger view on how to structure/determine the states though.


> I'm just don't think having the presence of a non-empty message on the existing conditions on the samplesresource as a good input in all cases for the truth table to determine things (it can in a case or two). Too fragile. 

agree.

> Solely using conditiontrue/false/unknown on the existing set of conditions in the samplesresource is insufficient as well, where our statements above capture that in my opinion.

hm, this one i'm not following.  Can you give me a specific example where a given set of conditions exists on the samplesresource, but those conditions (without the associated reason/message field) do not uniquely determine the appropriate set of clusteroperator condition values?  (setting aside what the reason/message should be on the clusteroperator conditions).  (You say it's above and it may be but i'm having trouble extracting it from the discussion at the moment).

Assuming their is such a case, i might be inclined to think that indicates a need for additional conditions on the samplesresource, but i guess i'll have to see what the scenario is first.

gabemontero:
OK I've pushed updates @bparees . 
- overall better organized and cleaner I think ... see if you agree
- Also a little more settled on the logic around determining state after re-review
- Added a few Conditions per discussion above to help manage state and errors
- I don't have a "truth table" documented in one place, but I think the comments around various decisions are a reasonable substitute ... see if you agree
- some minor fixes,
- and also a change to the imagestream retry between samples upserts and setting of the image import condition to true ... it worked better in local testing (started seeing fragility in the counter based retry), and the operator state being maintained can be lost with no ill consequence

gabemontero:
bump @bparees .... happy new year

pkg/apis/samplesoperator/v1alpha1/types.go
```diff
@@ -249,3 +310,117 @@ func (s *SamplesResource) Condition(c SamplesResourceConditionType) *SamplesReso
...
+	// the samples are still available at the old config setting; the
+	// config issues will be highlighted in the progressing/failing messages, per
+	// https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#conditions
+
+	// However, rhel and lack of creds is possible on intitial install, as well
+	// as deleted after the initial install; either circumstance can prevent imagestream
+	// scheduled imports for example to fail ... the imagestream "state" could
+	// be sufficiently compromised, so we'll flag false there
+	if needCreds {
+		return falseRC, falseMsg
```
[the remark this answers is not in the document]
gabemontero: Hmmmmmm ... so in a) taking in your point, and b) rereading (yet again) https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#conditions, I think there are these considerations here:

1) An API error during the initial install
2) We have installed at least once, are still at the installed version, but now have an error of some sort (could be bad api server interactions, needs creds for rhel, or the config is invalid
3) Item 2), but are migrating to a new version vs. staying at the installed version

For 1), I'll keep the empty message, and false ... details in the other two conditions

For 3), per the guideline, we need to set available to false, with the unable to reach <version> message

For 2), as you say, we should set available to true, with the ok at the version in question, where again the error will show up in the failing/progressing messages

To get to this behavior, I think we
- add spec.Version/status.Version comparison, accounting for not initially installed,  to determine if we are migrating versions
- if migrating, *ANY* error condition means we return false with the unable to reach message
- if not migrating,  we just leverage not any version, etc. as you note below

Thoughts? ... I'll work on the update if we agree here ...

thanks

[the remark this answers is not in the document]
bparees: per hallway discussion we'll revisit this after there is org-wide agreement on how failing/progressing/available should be used.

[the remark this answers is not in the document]
gabemontero: OK assuming https://github.com/openshift/cluster-version-operator/pull/73 holds up, scenario 3) goes away @bparees 

If so, I'll create a new PR to get rid of the apiError and needsCreds conditionals.

PR #71, opened 2019-01-04
move samples operator off of operator sdk; adjust e2e-test so it does…

gabemontero:
… not assume only the samples operator adds openshift namespace content

- the name changes are not in yet ... probably will do that in this PR

- a regen of the vendor tree to remove the sdk dependencies from there, as well as update some k8s deps in particular (there is one older vs. newer API marked with a TODO for example)


gabemontero:
FYI @bparees I have a couple of test/code fixes coming based on the e2e failure and some additional local testing

gabemontero:
Also potentially an issue I don't know how to fix yet ... the various factory informers closing the watches ... don't know yet how to learn of this and re-establish the watch:

```
ERROR: logging before flag.Parse: W0106 19:30:03.858407       1 reflector.go:341] github.com/openshift/cluster-samples-operator/vendor/github.com/openshift/client-go/image/informers/externalversions/factory.go:101: watch of *v1.ImageStream ended with: very short watch: github.com/openshift/cluster-samples-operator/vendor/github.com/openshift/client-go/image/informers/externalversions/factory.go:101: Unexpected watch close - watch lasted less than a second and no items received
time="2019-01-06T19:30:07Z" level=info msg="watch event cli not part of operators inventory"
ERROR: logging before flag.Parse: W0106 19:32:12.968719       1 reflector.go:341] github.com/openshift/cluster-samples-operator/vendor/github.com/openshift/client-go/template/informers/externalversions/factory.go:101: watch of *v1.Template ended with: very short watch: github.com/openshift/cluster-samples-operator/vendor/github.com/openshift/client-go/template/informers/externalversions/factory.go:101: Unexpected watch close - watch lasted less than a second and no items received
```

The watches did not re-establish on their own after several minutes.

bparees:
> don't know yet how to learn of this and re-establish the watch

as far as i know you shouldn't have to, the informer+lister should be handling all that for you.

gabemontero:
yeah it appears to be a red herring ... something else is going on ... still debugging ...

gabemontero:
Wondering if I am hitting rate limiting during imagestream image stream imports ....

gabemontero:
still curious about the precise numbers for the rate limiting (need to dig into that at some point) ... but found the bug in my new code re: the faililng e2e

bparees:
> still curious about the precise numbers for the rate limiting 

rate limiting where?

gabemontero:
fyi I have and updated vendor tree, based on `dep` vs. `glide`, a la the image registry op, functional on my laptop ... I'll push up separate commits based on it after it passes e2e locally ... should not impact any in progress reviews on the current commit here ... in addition to the commit for the vendor change, there is an update commit for the generated code, as well as a tweak in the controller to use the more recent k8s api for shared informer factory creation.

pkg/stub/handler.go
```diff
@@ -32,8 +30,10 @@ import (
 	templatev1client "github.com/openshift/client-go/template/clientset/versioned/typed/template/v1"
 
 	operatorsv1api "github.com/openshift/api/operator/v1"
-	"github.com/openshift/cluster-samples-operator/pkg/apis/samplesoperator/v1alpha1"
+	"github.com/openshift/cluster-samples-operator/pkg/apis/samplesresource/v1alpha1"
 	operatorstatus "github.com/openshift/cluster-samples-operator/pkg/operatorstatus"
+
+	sampopset "github.com/openshift/cluster-samples-operator/pkg/generated/clientset/versioned/typed/samplesresource/v1alpha1"
```
bparees: `sampleclientv1alpha1` would better follow our naming conventions

pkg/stub/handler.go
```diff
@@ -46,24 +46,32 @@ const (
...
-	h.sdkwrapper = &defaultSDKWrapper{h: &h}
+	crdWrapper := &generatedCRDWrapper{}
+	crdWrapper.h = h
+	client, err := sampopset.NewForConfig(kubeconfig)
+	if err != nil {
+		return nil, err
+	}
+	crdWrapper.c = client.SamplesResources()
+
+	h.crdwrapper = crdWrapper
```
bparees: handler contains crdwrapper which contains handler?  Seems like there's probably a better way to structure this?  Does the wrapper actually need the entire "handler" for something?

pkg/stub/handler.go
gabemontero: yeah i just copied img reg op naming convention

I'll change

pkg/stub/handler.go
gabemontero: A convention that has existed for a bit with most of these wrappers.

Now that the implementation has been sitting for a bit, I can replace the handler refs with the precise pieces within the handler that the wrapper needs.

gabemontero: OK just pushed a separate commit to address this

gabemontero:
ooooh no e2e-aws or e2e-aws-image-ecosystem test flakes ...

pkg/stub/handler.go
```diff
@@ -333,13 +342,17 @@ func (h *Handler) processImageStreamWatchEvent(is *imagev1.ImageStream, deleted
 	}
 	err = h.upsertImageStream(imagestream, is, srcfg)
 	if err != nil {
+		if kerrors.IsAlreadyExists(err) {
```
bparees: why would an upsert get an already exists error?  shouldn't it do an update if it already exists?

pkg/stub/handler.go
```diff
@@ -1763,101 +1786,96 @@ func (g *DefaultResourceFileLister) List(dir string) (files []os.FileInfo, err e
...
+type CRDWrapper interface {
 	Update(samplesResource *v1alpha1.SamplesResource) (err error)
 	Create(samplesResource *v1alpha1.SamplesResource) (err error)
 	Get(name string) (*v1alpha1.SamplesResource, error)
 }
 
-type defaultSDKWrapper struct {
+type generatedCRDWrapper struct {
 	h *Handler
+	c sampleclientv1alpha1.SamplesResourceInterface
```
bparees: s/c/client/ ?  (I'm not a fan of single letter variable names unless it's a temp/iterator var)

cmd/cluster-samples-operator/main.go
```diff
@@ -1,19 +1,12 @@
...
-	k8sutil "github.com/operator-framework/operator-sdk/pkg/util/k8sutil"
-	sdkVersion "github.com/operator-framework/operator-sdk/version"
-
-	configv1 "github.com/openshift/api/config/v1"
-	imagev1 "github.com/openshift/api/image/v1"
-	templatev1 "github.com/openshift/api/template/v1"
+	"github.com/openshift/cluster-samples-operator/pkg/operator"
+	"github.com/openshift/cluster-samples-operator/pkg/signals"
 
 	"github.com/sirupsen/logrus"
```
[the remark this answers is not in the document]
gabemontero: yeah i was trying to avoid this :-) .... I like logrus better

and yeah the auto generation uses glog :-(

pkg/stub/handler.go
gabemontero: concurrent activity can occur ... delete events for imagestreams that have been previously deleted as part of a samplesresource delete, but then then the new samplesresource creating the new item

pkg/stub/handler.go
```diff
@@ -1763,101 +1786,96 @@ func (g *DefaultResourceFileLister) List(dir string) (files []os.FileInfo, err e
...
-type SDKWrapper interface {
+type CRDWrapper interface {
 	Update(samplesResource *v1alpha1.SamplesResource) (err error)
 	Create(samplesResource *v1alpha1.SamplesResource) (err error)
 	Get(name string) (*v1alpha1.SamplesResource, error)
 }
 
-type defaultSDKWrapper struct {
+type generatedCRDWrapper struct {
 	h *Handler
```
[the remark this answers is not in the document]

pkg/stub/handler.go

gabemontero:
OK updates from the in depth pass pushed.

Shall I go ahead and do either / both:
- vendor changes to remove sdk deps, simplify
- name changes

In separate commits with this PR @bparees ?

bparees:
update the vendoring here.  name changes should be a separate PR entirely.

gabemontero:
10-4 thanks

gabemontero:
vendor commit pushed

PR #73, opened 2019-01-09
add message in cfg valid condition if override unsupported mgmt state

gabemontero:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1659351

### DIFF
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -992,6 +992,12 @@ func (h *Handler) ProcessManagementField(srcfg *v1alpha1.SamplesResource) (bool,
 	default:
 		// force it to Managed if they passed in something funky, including the empty string
 		logrus.Warningf("Unknown management state %s specified; switch to Managed", srcfg.Spec.ManagementState)
+		cfgvalid := srcfg.Condition(v1alpha1.ConfigurationValid)
+		cfgvalid.Message = fmt.Sprintf("Unexpected management state %v received, switching to %v", srcfg.Spec.ManagementState, operatorsv1api.Managed)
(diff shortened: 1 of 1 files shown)


pkg/stub/handler.go
```diff
@@ -958,6 +958,12 @@ func (h *Handler) ProcessManagementField(srcfg *v1alpha1.SamplesResource) (bool,
 	default:
 		// force it to Managed if they passed in something funky, including the empty string
 		logrus.Warningf("Unknown management state %s specified; switch to Managed", srcfg.Spec.ManagementState)
+		cfgvalid := srcfg.Condition(v1alpha1.ConfigurationValid)
+		cfgvalid.Message = fmt.Sprintf("Unexpected management state %v received, switching to %v", srcfg.Spec.ManagementState, operatorsv1api.Managed)
+		now := kapis.Now()
+		cfgvalid.LastTransitionTime = now
+		cfgvalid.LastUpdateTime = now
+		srcfg.ConditionUpdate(cfgvalid)
```
bparees: Thinking about this more, why not:

1) mark configurationvalid=false (with the message indicating that the management state is invalid)
2) leave the management state as whatever they set it to?

bparees: nm, I see we went through this discussion in the bug.  I still kinda prefer what i said above but i'm not gonna re-litigate it after signing off on this approach earlier.

gabemontero: I explicitly did not mark it false, as config validation will occur after this

IMM, the whole point was not to do 2)

That said, you lgtm'ed ... so perhaps moot

PR #77, opened 2019-01-11
Prevent usr upd status

gabemontero:
/assign @bparees 

this has commit from https://github.com/openshift/cluster-samples-operator/pull/76, which is still fighting aws flakes while on the merge queue

pls review https://github.com/openshift/cluster-samples-operator/pull/77/commits/7543e7d5085b07e8cbf60e47c512b9342656cdc9

will rebase, etc. as needed

### DIFF
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -331,8 +331,8 @@ func (h *Handler) processImageStreamWatchEvent(is *imagev1.ImageStream, deleted 
 			importError.LastTransitionTime = now
 			importError.LastUpdateTime = now
 			cfg.ConditionUpdate(importError)
-			logrus.Debugf("SDKUPDATE no migration / no pending / no error imgstr update")
-			return h.crdwrapper.Update(cfg)
+			logrus.Debugf("CRDUPDATE no migration / no pending / no error imgstr update")
+			return h.crdwrapper.UpdateStatus(cfg)
 		}
 
 		// clear out error for this stream if there were errors previously but no longer are
@@ -350,7 +350,7 @@ func (h *Handler) processImageStreamWatchEvent(is *imagev1.ImageStream, deleted 
 			importError.LastTransitionTime = now
 			importError.LastUpdateTime = now
 			cfg.ConditionUpdate(importError)
-			logrus.Debugf("SDKUPDATE no error imgstr update")
+			logrus.Debugf("CRDUPDATE no error imgstr update")
 		}
 
 		return nil
@@ -365,8 +365,8 @@ func (h *Handler) processImageStreamWatchEvent(is *imagev1.ImageStream, deleted 
 	if err != nil {
 		// still attempt to report error in status
 		h.processError(cfg, v1.SamplesExist, corev1.ConditionUnknown, err, "%v error reading file %s", filePath)
-		logrus.Debugf("SDKUPDATE event img update err bad fs read")
-		h.crdwrapper.Update(cfg)
+		logrus.Debugf("CRDUPDATE event img update err bad fs read")
+		h.crdwrapper.UpdateStatus(cfg)
 		// if we get this, don't bother retrying
 		return nil
 	}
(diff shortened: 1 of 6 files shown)


pkg/stub/handler.go
```diff
@@ -1873,7 +1854,7 @@ type generatedCRDWrapper struct {
 
 func (g *generatedCRDWrapper) Update(sr *v1.Config) error {
 	return wait.Poll(3*time.Second, 30*time.Second, func() (bool, error) {
-		_, err := g.client.Update(sr)
+		_, err := g.client.UpdateStatus(sr)
```
bparees: you're only ever updating status fields when calling crdwrapper.Update()?  What about removing the finalizer?

should probably rename the function to UpdateStatus() to make it very clear (and confirm all the places where you're calling it only expect to update .status fields)

gabemontero: OK ... the finalizer was *NOT* getting removed

I'll have to revisit the e2e's ... (unless I got running e2e for this PR confused with the other PRs I was workind concurrently yesterday).

I was having trouble updating the config.spec.version with client.Update yesterday (but I was trying to call that and then subsequently call UpdateStatus).  I'll revisit the img reg operator (or others as needed) if I need a cross ref after trying again with updating the metadata.

bparees: you can see the full test history for this PR here:
https://openshift-gce-devel.appspot.com/pr/openshift_cluster-samples-operator/77

(unless you mean running e2es locally)

gabemontero: yeah I meant running locally ... and the e2e is failing locally for me now too ... I must have had a mixup with manually updating the crd/clusterrole, etc. during initial local dev (I think there may have been a half done in one running cluster and half done on another, with running the e2e's in the second half).

gabemontero: OK Update is working for finalizer removal ... will update PR shortly

gabemontero: OK got update for finalizer working ... updates pushed ... ptal @bparees

pkg/stub/handler.go
```diff
@@ -735,6 +690,27 @@ func (h *Handler) CreateDefaultResourceIfNeeded(cfg *v1.Config) (*v1.Config, err
...
+	valid := cfg.Condition(v1.ConfigurationValid)
+	// our default config is valid; since Condition sets new conditions to false
+	// if we get false here this is the first pass through; invalid configs
+	// are caught above
+	if valid.Status != corev1.ConditionTrue {
+		valid.Status = corev1.ConditionTrue
+		valid.LastUpdateTime = now
+		valid.LastTransitionTime = now
+		cfg.ConditionUpdate(valid)
+	}
```
bparees: any reason not to just init the conditionvalid to false on the first pass(bootstrap?) and then when we see the CR we'll process the config and (assuming it's valid) update the configvalid to true at that point?

pkg/stub/handler.go
```diff
@@ -1338,9 +1315,14 @@ func (h *Handler) Handle(event v1.Event) error {
 			}
 			logrus.Debugf("Handle Reason field set to %s", progressing.Reason)
 			cfg.ConditionUpdate(progressing)
-			cfg.Spec.Version = v1.GitVersionString()
-			logrus.Debugf("SDKUPDATE progressing true update")
-			err = h.crdwrapper.Update(cfg)
+
+			// now that we employ status subresources, we can't populate
+			// the conditions on create; so we do initialize here, which is our "step 1"
+			// of the "make a change" flow in our state machine
```
bparees: hm. I kinda thought you can `create` an object w/ status fields populated, you just can't do an `update` against status fields.

bparees: (but if this works, it's fine)

gabemontero:
e2e-aws-operator failure terraform/aws flake

pkg/stub/handler.go
gabemontero: as you alluded to, this way saves us a round trip with the API server

this optimization is safe (IMO) since our initially created config is valid

bparees: k

PR #78, opened 2019-01-11
godoc/value consistency fixes for types via wewang/bparees

gabemontero:
built on https://github.com/openshift/cluster-samples-operator/pull/77 and https://github.com/openshift/cluster-samples-operator/pull/76 which are still fighting merge queue flakes


bparees:
doesn't look like it's built on top of anything (it's a single commit)

/lgtm

PR #79, opened 2019-01-12
fix import img err detect when len(status) < len(spec); minor fixes o…

gabemontero:
…nce error properly detected

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1663406

/assign @bparees


gabemontero:
Using `?w=1` on the github file diff greatly simplifies identifying the changes in this case

PR #81, opened 2019-01-16
force default to rhel and registry to old redhat one

gabemontero:
https://jira.coreos.com/browse/DEVEXP-251

/assign @bparees


gabemontero:
@bparees : looking at the image-ecosystem failures, some of them seem related to not finding the wildfly imagestream:

With `openshift-tests [image_ecosystem][Slow] openshift sample application repositories [image_ecosystem][nodejs] images with nodejs-ex repo Building nodejs app from new-app should build a nodejs image and run it in a pod [Suite:openshift] 46s`

It does a check on the imagestreams before continuing, and:

```
Jan 16 23:48:57.056: INFO: ImageStream Error: &errors.StatusError{ErrStatus:v1.Status{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ListMeta:v1.ListMeta{SelfLink:"", ResourceVersion:"", Continue:""}, Status:"Failure", Message:"imagestreams.image.openshift.io \"wildfly\" not found", Reason:"NotFound", Details:(*v1.StatusDetails)(0xc422232a20), Code:404}} 
```

`wildfly` is not included in the rhel/ocp version given all the EAP/JBoss content I would presume.

- Start updating the e2e's to not look for `wildfly`?
- Retag `wildfly` in openshift/library to be included in ocp?
- Special case the samples operator to include `wildfly` in rhel/ocp?

gabemontero:
on the operator e2e there were a bunch of apiserver comm errors and the samples did not stabilize within the tests' time expectations

bparees:
> Start updating the e2e's to not look for wildfly?

this one.

I don't think we have too many actual tests that rely on it (beyond checking for its presence), so hopefully it's not too painful.  If you do run into cases where we need to do a legitimate java-based s2i build or something, you should be able to switch the TC to create its own wildfly imagestream.

gabemontero:
OK ...shall I update https://github.com/openshift/origin/pull/21762 with those additional e2e tweaks, or do you prefer a separate PR?

bparees:
> OK ...shall I update openshift/origin#21762 with those additional e2e tweaks, or do you prefer a separate PR?

separate please

gabemontero:
Some api server comm hiccups, but also some image import failures going to registry.access.redhat.com prevented the e2e tests to get to the expected final state .....

... the switch to rhel might force us to implement image import retry within the operator (along with longer grace times to get the samples stable) @bparees :-(

or allow for image import failures wrt e2e validations and cluster operator status :-<

gabemontero:
fyi `aws-image-ecosystem` passed on my local AWS cluster with this PR's changes as well as https://github.com/openshift/origin/pull/21824

note, I had to manually retry an image import got get all the rhel imagestreams clean during that test (it was in an eap/jboss imagestream that was unrelated to image_ecosystem)

Per discussions with @bparees I'll start tackling the operator retrying image imports with this PR in the hopes of getting `e2e-aws-operator` happier

bparees:
these bits lgtm.

please put the retry logic in a separate commit for easy reviewing.

README.md
```diff
@@ -79,6 +79,12 @@ The samples resource maintains the following conditions in its status:
 -- True when an error has occurred
 -- The list of imagestreams with an error will be in the reason field
 -- The details of each error reported will be in the message field
+- ImageImportRetryInProgress
```
gabemontero: Ignore this - I removed it in the last commit

pkg/apis/samples/v1/types.go
```diff
@@ -205,6 +205,11 @@ const (
 	// change cycle as complete in both ClusterOperator and Config; retry on import will
 	// occur by the next relist interval if it was an intermittent issue;
 	ImportImageErrorsExist ConfigConditionType = "ImportImageErrorsExist"
+	// ImageImportRetryInProgress captures when the samples operator is retrying an image import
+	// and imagestream when it previously detected a problem occurred during the image import
+	// resulting from a prior create or update of the imagestream.  The list of imagestreams
+	// the samples operator is retrying imports for is in the Reason field of this condition.
+	ImageImportRetryInProgress ConfigConditionType = "ImageImportRetryInProgress"
```
gabemontero: Ignore this - I removed it in the last commit

pkg/stub/handler.go
```diff
@@ -321,6 +321,7 @@ func (h *Handler) initConditions(cfg *v1.Config) *v1.Config {
 	cfg.Condition(v1.RemovedManagementStateOnHold)
 	cfg.Condition(v1.MigrationInProgress)
 	cfg.Condition(v1.ImportImageErrorsExist)
+	cfg.Condition(v1.ImageImportRetryInProgress)
```
gabemontero: Ignore this - I removed it in the last commit

pkg/stub/imagestreams.go
```diff
@@ -358,3 +276,128 @@ func (h *Handler) clearStreamFromImportError(name string, importError *v1.Config
...
+					// add this imagestream to the Reason field
+					if !strings.Contains(importError.Reason, is.Name+" ") {
+						now := kapis.Now()
+						importError.Reason = importError.Reason + is.Name + " "
+						importError.Message = importError.Message + "<imagestream/" + is.Name + ">" + message + "<imagestream/" + is.Name + ">"
+						importError.Status = corev1.ConditionTrue
+						importError.LastTransitionTime = now
+						importError.LastUpdateTime = now
+						cfg.ConditionUpdate(importError)
+						// initiate a retry if needed
```
gabemontero: Ignore retry block - I removed it in the last commit

gabemontero:
image eco is passing now

gabemontero:
the e2e-operator test suffered from various api server conn timeouts and enough import image errors to registry.access.redhat.com `(Client.Timeout exceeded while awaiting headers)` where the retries also failed such that the tests did not observe their expected results in time.

gabemontero:
e2e-aws-operator cluster failed during startup

Attempts to parse all the new diagnostic data:
1) machine-config-operator was in failing state with `"error syncing: timed out waiting for the condition during syncRequiredMachineConfigPools: error pool master is not ready. status: (total: 3, updated: 3, unavailable: 1)",`

2) openshift-api-server was still progressing with `"daemonset/apiserver.openshift-operator: observed generation is 1, desired generation is 2.\nopenshiftapiserveroperatorconfigs/instance: observed generation is 0, desired generation is 2."`

3) connections refused going to kube-system namespace

pkg/stub/imagestreams.go
```diff
@@ -16,160 +17,90 @@ func (h *Handler) processImageStreamWatchEvent(is *imagev1.ImageStream, deleted
...
+
+		// if we initiated the change, it will be in our cache, and we want to process
+		// the in progress, import error conditions
+		beforeErrorReason := cfg.Condition(v1.ImportImageErrorsExist).Reason
+		beforeProgressReason := cfg.Condition(v1.ImageChangesInProgress).Reason
+		beforeInReasons := strings.Contains(beforeProgressReason, is.Name+" ") ||
+			strings.Contains(beforeErrorReason, is.Name+"")
+		if cache.UpsertsAmount() > 0 {
+			cache.AddReceivedEventFromUpsert(is)
+			if !cfg.ConditionTrue(v1.ImageChangesInProgress) || !cfg.ConditionTrue(v1.SamplesExist) {
```
bparees: should this be `if cfg.ConditionTrue(v1.ImageChangesInProgress) || !cfg.ConditionTrue(v1.SamplesExist)`  ?

(ie "if changesinprogress OR notsamplesexist")

pkg/stub/imagestreams.go
```diff
@@ -354,3 +293,148 @@ func (h *Handler) clearStreamFromImportError(name string, importError *v1.Config
...
+				logrus.Printf("There are no more image imports in flight for imagestream %s", is.Name)
+			}
+		}
+	}
+
+	// clear out error for this stream if there were errors previously but no longer are
+	// think a scheduled import failing then recovering
+	if strings.Contains(importError.Reason, is.Name) && !anyErrors {
+		logrus.Printf("There are no image import errors for %s", is.Name)
+		importError = h.clearStreamFromImportError(is.Name, importError, cfg)
```
bparees: you don't appear to be doing anything w/ this return value

pkg/stub/retry.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	// set the target item to import
+	if multiple {
+		tag = finalTag
+	}
+
+	// reset the generation
+	zero := int64(0)
+	existing.Generation = &zero
```
bparees: it doesn't look like existing.Generation is used after this point, is this needed?

pkg/stub/retry.go
```diff
@@ -0,0 +1,147 @@
...
+		})
+	}
+	return isi
+}
+
+func newImageStreamImport(stream *imagev1.ImageStream) *imagev1.ImageStreamImport {
+	isi := &imagev1.ImageStreamImport{
+		ObjectMeta: kapis.ObjectMeta{
+			Name:            stream.Name,
+			Namespace:       "openshift",
```
bparees: do we not have a constant for this("samples installation namespace") somewhere?

bparees: alternatively can't you get the namespace from stream.Namespace?

gabemontero:
looks like there is a problem with delete during startup commit https://github.com/openshift/cluster-samples-operator/pull/81/commits/bf407debd1ab3fb49c5061be785f743ab3dfe277

investigating.....

gabemontero:
ok sorted it out ... commit updated ... will move to @bparees 's comments shortly

pkg/operator/controller.go
```diff
@@ -263,8 +268,12 @@ func (c *Controller) handleWork(getter runtimeObjectGetter, o interface{}) error
 		obj, err := getter.Get(c, key)
 		if err != nil {
 			// see if this is a operator bootstrap scenario
-			if kerrors.IsNotFound(err) && key == sampopapi.ConfigName {
-				return c.Bootstrap()
+			if kerrors.IsNotFound(err) {
+				if key == sampopapi.ConfigName {
+					return c.Bootstrap()
+				}
+				logrus.Printf("handleWork resource %s has since been deleted, ignore update event", key)
```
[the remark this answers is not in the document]
gabemontero: It means it in my mind ... not sure what you mean "a name we don't care about"

Perhaps if I detailed the background, it would help us get to common ground:
- this path is the update/create path
- in various testing runs I would see instances where we update a sample, and then a delete it very soon after
- this code path would get the update event, and then when we refetch, the delete had gone through, and we get the non found
- requeueing the update on a not found error serves no purpose, since it is truly gone, so we got in a always repeating requeue cycle ... this prevented it

Alternatives to the wording:
- `this upsert event for resource %s will be ignored since the resource has been deleted since this event was generated`

Or does this elaboration result in an alternative you want to provide?

thanks

pkg/stub/imagestreams.go
gabemontero: No.

in progress is not actually set to true until *AFTER* the samples are upserted ...so we can get events for those upserts before we set in progress to true, especially since cycling through upserting all the resources can take a bit

In case you don't recall, the "whether we set in progress to true before or after we start the actual samples upsert cycle" is a discussion we had with a few iterations earlier on.  Setting *AFTER* was deemed safest.

I generally go with !cfg.Condition... to count for the Unknown condition, which can occur when we have trouble communicating with the api server.

pkg/stub/imagestreams.go
gabemontero: yeah a vestige of an earlier iteration which ultimately was not git pushed ... will clean up

pkg/stub/retry.go
gabemontero: yep some leftover from the borrowed code  ... will remove

pkg/stub/retry.go
gabemontero: switched to `stream.Namespace`

gabemontero:
first pass updates from you Jan23 comments pushed @bparees

pkg/operator/controller.go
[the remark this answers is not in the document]
bparees: this specific block deals w/ handling an event for an update to the samples operator resource (not an imagestream or template), right?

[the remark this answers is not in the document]
gabemontero: the `getter` is a locally defined interface that has impls for all 4 api types the operator watches

gabemontero:
the image-eco failure was a dancer-mysql build taking 11 minutes to complete  ... test gave up

pkg/operator/controller.go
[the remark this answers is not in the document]
bparees: so if i name an imagestream "instance" and then delete it, i can potentially send the system through c.Bootstrap()?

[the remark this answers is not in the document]
gabemontero: dah .. good point ... I'll update to check to getter type as well

pkg/stub/imagestreams.go
bparees: thanks, yeah i can never keep that one in my head

pkg/operator/controller.go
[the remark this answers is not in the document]
gabemontero: new commit pushed for that check

pkg/stub/config.go
```diff
@@ -170,34 +170,43 @@ func (h *Handler) VariableConfigChanged(cfg *v1.Config) (bool, bool, bool) {
...
-	newTempMap := make(map[string]bool)
-	for _, st := range opcfg.Spec.SkippedTemplates {
-		newTempMap[st] = true
-	}
-	for _, si := range opcfg.Spec.SkippedImagestreams {
-		newStreamMap[si] = true
+	h.mapsMutex.Lock()
+	defer h.mapsMutex.Unlock()
+	if len(opcfg.Spec.SkippedImagestreams) != len(h.skippedImagestreams) ||
+		len(opcfg.Spec.SkippedTemplates) != len(h.skippedTemplates) {
```
bparees: don't you also have to care if the values changed within the arrays (ie one value was replaced by another, to the lengths stayed the same?)

pkg/stub/handler.go
```diff
@@ -626,8 +623,14 @@ func (h *Handler) Handle(event v1.Event) error {
...
 				cfg.ConditionFalse(v1.ImageChangesInProgress) &&
 				v1.GitVersionString() == cfg.Status.Version {
-				logrus.Debugf("Handle ignoring because config the same and exists is true, in progress false, and version correct")
-				return nil
+				logrus.Debugf("At steady state: config the same and exists is true, in progress false, and version correct")
+
+				// in case this is a bring up after initial install, we take a pass
+				// and see if any samples were deleted while samples operator was down
+				h.buildFileMaps(cfg)
+				h.buildSkipFilters(cfg)
```
bparees: didn't the skipfilters already get built up on line 606?

gabemontero:
on e2e-aws-operator failure, flaky registry.access.redhat.com

pkg/stub/handler.go
gabemontero: yep missed that ... removing this call

pkg/stub/config.go
gabemontero: yeah with the mutex and rechecking the remaining calls, removing these checks and rebuilding is fine

will just simplify that way

gabemontero:
no flakes with the last run

gabemontero:
after that last successful set of runs, let's see how reliable registry.access.redhat.com and ci flakes are for us today

/retest

bparees:
@gabemontero i don't think i have anything else, squash to what you feel is appropriate and lgtm.

gabemontero:
ok thanks @bparees 

outside of the squashing, I'm going to execute the operator e2e throughout the day just to get a few more data points on various flakiness (most notable RH registry, along with any tweaks in the operator or e2e's that might jump out)

pending any shattering discoveries, I'll label lgtm by EOB

gabemontero:
yeah I'm now seeing an intermittent flake between the end of `TestArchitectureConfigChangeValidation` and the beginning of `TestSkippedProcessing`

I think I need to add some retry on conflicts when I'm resetting the Config object for the next test.

gabemontero:
ok e2e is passing locally for me ... let's see if my e2e fixes translate to `e2e-aws-operator`

gabemontero:
yay `e2e-aws-operator` passed !!

gabemontero:
image import errors

gabemontero:
@bparees I'm beginning to think https://github.com/openshift/origin/blob/master/test/extended/util/framework.go#L54-L91 is going to need some tweaking for moving to the rhel samples ...that code is only giving the imagestreams 30 seconds to complete ...depending on what runs first, and if the controller/registry/node is trying to get all those images downloaded and ready at once, that isn't enough time.

Even with my optimizations, it takes the entire stream suite between 1 and 2 minutes to report all ready.

In the test debug, I'm not seeing import errors .... there simply are not any attempts yet in the imagestream status it import the tags in the spec.

That test has passed sometime, but only intermittently.

bparees:
@gabemontero that check should be getting run well after the cluster is installed and the samplesoperator has reported available...  (no tests should be getting until until the samples operator has reported available).

gabemontero:
@bparees hmm .... without the presence of import errors, but the fact that the streams exist (meaning the samples operator created them), then something else would be amiss (image registry/controller/crio ?!?! ... just speculating there)

I'll see about adding a dump of the samples operator CR and its pod logs into the current debug.  Start eliminating some possibilities.

gabemontero:
`e2e-aws-operator` has passed consistently several times since the last commit 

adding merge label back in

will retest image-eco after https://github.com/openshift/origin/pull/21908 merges

gabemontero:
temporarily removing the lgtm again

- https://github.com/openshift/cluster-samples-operator/pull/88 will include a fix for reconciling `jenkins-agent-nodejs` vs. `nodejs` in the in progress reason code 
- merging https://github.com/openshift/cluster-samples-operator/pull/88 before this also will allow us to remove the forcing of jenkins to the centos version

gabemontero:
undoing the forcing of jenkins to centos pushed

@bparees ptal

pkg/stub/imagestreams.go
```diff
@@ -207,8 +207,11 @@ func (h *Handler) upsertImageStream(imagestreamInOperatorImage, imagestreamInClu
 }
 
 func (h *Handler) updateDockerPullSpec(oldies []string, imagestream *imagev1.ImageStream, opcfg *v1.Config) {
-	//TODO remove when TBR creds sorted out
-	if imagestream.Name == "jenkins" {
+	//TODO remove when TBR creds sorted out and we do not explicitly set SamplesRegistry to registry.access.redhat.com
+	// by default we want to leave the jenkins images as using the payload set to the IMAGE* env's
+	// but if the customer overrides (after we switch to TBR and don't set SamplesRegistry by default), we'll let them
+	// change jenkins related images
+	if strings.HasPrefix(imagestream.Name, "jenkins") {
```
bparees: probably better to call out the 3(?) we explicitly provide, just in case someone upstream contributes a "jenkins-foo" imagestream that we want to install later, but isn't part of the payload.

gabemontero: update pushed

gabemontero:
several image import failure torpedoed the e2e-aws-operator run

per https://coreos.slack.com/archives/CE2HALN2W/p1548965702099700 from @bparees we might have to opportunity to ignore image import failures which bypass such e2e failures

this PR and the switch to RHEL might add extra motivation for that (and hence perhaps add that here)

gabemontero:
@bparees preliminarily / for the moment I've pushed a commit up that nukes the setting of failing to true if there are image import errors

the change is trivial and easily reverted, even if the commit is squashed

bparees:
> @bparees preliminarily / for the moment I've pushed a commit up that nukes the setting of failing to true if there are image import errors

ok let's see what happens.  my primary concern at the moment is whether it starts causing e2e failures because tests that expect the imagestreams are imported before they start running, won't necessarily have that expectation met.

gabemontero:
roger that .... image-ecosystem should be a good indicator for whether that happens IMO

gabemontero:
ok now it is working :-)

gabemontero:
ok @bparees I've pushed as a separate commit (after successfully testing beforehand this time :-) ) the removal of OKD/centos

so reversible if the pending decision here goes unexpectedly 

ptal

bparees:
@gabemontero let's keep that commit out of this PR for now since we're still waiting for PM sign off, and the rest of this PR should be good to go in now(right?)

gabemontero:
@bparees - I *may* have some evidence that the extended tests are running before the samples operator has finished:

- see the failure at https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cluster-samples-operator/81/pull-ci-openshift-cluster-samples-operator-master-e2e-aws-image-ecosystem/149 : 

```
openshift namespace image streams corrupted 

Feb  1 20:08:37.583: INFO: Running 'oc get --config=/tmp/admin.kubeconfig --namespace=e2e-test-s2i-perl-n6wsb is -n openshift -o yaml --config /tmp/admin.kubeconfig'
Feb  1 20:08:38.008: INFO: 
  imagestreams in openshift namespace: 
```

- then, note the cluster operator status of the samples operator from https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-samples-operator/81/pull-ci-openshift-cluster-samples-operator-master-e2e-aws-image-ecosystem/149/artifacts/e2e-aws-image-ecosystem/clusteroperators.json :

```
        {
            "apiVersion": "config.openshift.io/v1",
            "kind": "ClusterOperator",
            "metadata": {
                "creationTimestamp": "2019-02-01T20:08:19Z",
                "generation": 1,
                "name": "openshift-cluster-samples-operator",
                "resourceVersion": "17376",
                "selfLink": "/apis/config.openshift.io/v1/clusteroperators/openshift-cluster-samples-operator",
                "uid": "1e958ca0-265d-11e9-9096-0e67d4cdc0ea"
            },
            "spec": {},
            "status": {
                "conditions": [
                    {
                        "lastTransitionTime": "2019-02-01T20:09:57Z",
                        "message": "Samples installation successful at 4.0.0-alpha1-59dae7a7c",
                        "status": "True",
                        "type": "Available"
                    },

```

it wasn't even created until 19 seconds before the error log from the test

I found the samples operator pod logs and they fall in line with this analysis.

- first, the samples were not "ready" until 20:09:51:

```
time="2019-02-01T20:09:51Z" level=info msg="The last in progress imagestream has completed"
time="2019-02-01T20:09:51Z" level=info msg="There are no more image imports in flight for imagestream jboss-webserver30-tomcat8-openshift"
time="2019-02-01T20:09:51Z" level=info msg="CRDUPDATE updating progress/error condition after results for ruby"
time="2019-02-01T20:09:54Z" level=info msg="An image import error occurred applying the latest configuration on version 4.0.0-alpha1-59dae7a7c; this operator will periodically retry the import, or an administrator can investigate and remedy manually"
time="2019-02-01T20:09:54Z" level=info msg="The samples are now at version 4.0.0-alpha1-59dae7a7c"
```

- and it did not even start until a few seconds before the e2e test complained about corruption:

```
time="2019-02-01T20:08:16Z" level=info msg="creating default Config"
time="2019-02-01T20:08:19Z" level=info msg="waiting for informer caches to sync"
time="2019-02-01T20:08:19Z" level=info msg="started events processor"
time="2019-02-01T20:08:19Z" level=info msg="management state set to managed"
time="2019-02-01T20:08:19Z" level=info msg="Spec is valid because this operator has not processed a config yet"
```

So unless clock skew is a think in our env, and the test pod is not remotely in sync with the sample operator pod, the appearance that the e2e's are running "too soon" would seem to hold.

Solutions / plans of attack:

- We could put check in the these e2e methods to Poll the samples operator cluster status and not start until it is available.

- circle back to openshift/release and the usual suspects there and confirm a few things / perhaps get them to "adjust" a few things

gabemontero:
the "starting too soon" situation appears relevant to the e2e-aws-operator failure as well

The samples operator did not initially reach steady state until about 10 minutes into the test

@bparees is sorting out the need for a "starter" clusteroperator obj in the deploy dir of the operator with the install team to help the installer gate on samples operator availability as I type

gabemontero:
ok @bparees ptal

2 commits
1) the clusteroperator stub yaml
2) cluster operator available / progressing only gated by CR samples exists

left old form present but commented out (for reference in case we switch back) and added some comments on rationale

gabemontero:
ok I'll see about looking into that over the weekend in detail and update the commit / comment as needed in prep for talking on Monday

gabemontero:
Couldn't help it ... some more thought embedded below, and I will be
changing some

On Fri, Feb 1, 2019 at 6:24 PM Ben Parees <notifications@github.com> wrote:

> Some of the condition logic looks off to me, let's talk through it on
> monday.
>
> What i'd naively expect:
>
> available=true - content created (doesn't mean the import happened yet).
> We have no choice about this to avoid blocking the installer
>
> progressing=true - ideally: we are in the process of creating/updating the
> content. maybe also we are in the process of importing content (import
> hasn't finished yet). *maaaaaaaaybe* also "some content is currently
> failing to import but we're periodically retrying it"
>
My interpretation is already waffling ... my first today read on
https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#conditions
was that both progressing and available being true as only pertaining to a
release migration (4.0.0 to 4.0.1) in his example

But your comment prompted me to re-review .... even on the same level, we
are progressing to some goal.

I'm going to revert that to the old form.

> failing=true - we couldn't create content for some reason. I'm also ok w/
> the idea that failing=true means an import failed (which is what this
> currently does I think).
>
No import failed meaning failing is true is commented out / that was removed
But I could revert that if we like.  But my thought was the CVO at some
point could "block" on failing==true

But yeah we can talk on Monday

> However right now it looks like when we have import errors, failing will
> be set to true, and when failing is set to true, we'll report
> progressing=true:
>
> a4f6417#diff-b62740dde2d9512574a75d6568b42fefR425
> <https://github.com/openshift/cluster-samples-operator/commit/a4f641700aade35127cd8672cb3c8d9628c4bbc6#diff-b62740dde2d9512574a75d6568b42fefR425>
>
> which..might be ok? but it seems weird that upon creating an imagestream
> we might go from progressing=false, to progressing=true if the import fails.
>
I based that (even prior to this change) off of Clayton's last example
under
https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#conditions

Progressing will be true, but it will have a negative sounding message
things going wrong.

I think the idea is that if you have not reached your "goal", you are still
"progressing", even if you may not be actively doing anything else,
though we will for image import on the sync/relist interval.

gabemontero:
there we go @bparees  ... between the e2e test change and adding the operator stub here we've got clean image-eco as well

going to retest again to see how consistent

gabemontero:
image-eco failed, but it was after the verification of the imagestreams being present that bit us before; this time, it was a problem pulling an image from the RH registry during a build:

```
Feb  3 22:18:27.156: INFO: Running 'oc logs --config=/tmp/admin.kubeconfig --namespace=e2e-test-s2i-php-jw8mm pod/cakephp-mysql-example-1-build -c sti-build -n e2e-test-s2i-php-jw8mm'
Feb  3 22:18:27.590: INFO: Log for pod "cakephp-mysql-example-1-build"/"sti-build"
---->
Caching blobs under "/var/cache/blobs".
Pulling docker://image-registry.openshift-image-registry.svc:5000/openshift/php@sha256:3895d8c39906fb07578ad5eb6dbdfb91471d2ebba570c67b2a1fccdf56c40c20
error: build error: 1 error occurred:
	* Error determining manifest MIME type for docker://image-registry.openshift-image-registry.svc:5000/openshift/php@sha256:3895d8c39906fb07578ad5eb6dbdfb91471d2ebba570c67b2a1fccdf56c40c20: Error reading manifest sha256:3895d8c39906fb07578ad5eb6dbdfb91471d2ebba570c67b2a1fccdf56c40c20 in image-registry.openshift-image-registry.svc:5000/openshift/php: unknown: unable to pull manifest from registry.access.redhat.com/rhscl/php-71-rhel7:latest: Get https://registry.access.redhat.com/v2/: net/http: request canceled (Client.Timeout exceeded while awaiting headers)
<----end of log for "cakephp-mysql-example-1-build"/"sti-build"
```

gabemontero:
slow registry

bparees:
/test e2e-aws
/hold

per clayton's email, let's make sure this passes e2e-aws a few times so we know the new clusteroperator object doesn't cause us to block the install due to flakes/failures in our operator.

bparees:
unrelated flakes.
/test e2e-aws

gabemontero:
networking flakes during initial install bringup of api server

/test e2e-aws

PR #86, opened 2019-01-29
update to kube 1.12 and update to new clusteroperator version logic

bparees:


pkg/operatorstatus/operatorstatus.go
```diff
@@ -133,11 +133,18 @@ func (o *ClusterOperatorHandler) setOperatorStatus(condtype configv1.ClusterStat
 			Message:            msg,
 			LastTransitionTime: metaapi.Now(),
 		})
-		if !modified && state.Status.Version == currentVersion {
+		if !modified && len(state.Status.Versions) > 0 && state.Status.Versions[0].Version == currentVersion {
```
bparees: all this logic is very much a short term hack that assumes we only ever put a single entry in Status.Versions.  If we ever need more, this will need revisiting.

gabemontero:
otherwise i'm gtm

bparees:
> Assuming things compile, I still wonder if cross referencing with openshift/origin to see if they were also updated.

they did compile as is, but i'll double check what origin is using these days.

bparees:
@gabemontero updated to align w/ what origin is locking these days.

PR #95, opened 2019-02-08
rebuild filesystem map on config change

gabemontero:


gabemontero:
the latest e2e-aws-operator run uncovered some more short comings in the e2e test ... it minimally needs to wait for the skip list change to be fully processed by the operator ... in the last run, it became apparent it was not

will be adding fixes to this PR

gabemontero:
image-eco:  another incredibly slow dancer-mysql build wrt downloading dancer-mysql dependencies from various github and other internet locations

gabemontero:
and the first clean aws-operator in a bit

gabemontero:
running operator e2e again

/test e2e-aws-operator

gabemontero:
aws-operator passes again

image-eco flakes again with the same dancer slowness ....

@bparees ptal at the changes

I'll initiate more test runs, but if these slowness flakes persist, having to dive into them and possible disabling tests or increasing timeouts of these tests might be needed .... I think the delays are upstream internet issues, but a second or third set of eyes may not hurt if it comes to that

pkg/stub/interfaces.go
```diff
@@ -141,13 +142,19 @@ func (g *DefaultImageStreamFromFileGetter) Get(fullFilePath string) (is *imagev1
...
 	}
-	switch {
-	case imagestream.Name == "jenkins":
-		imagestream = tagInPayload("2", "IMAGE_JENKINS", imagestream)
-	case imagestream.Name == "jenkins-agent-maven":
-		imagestream = tagInPayload("v4.0", "IMAGE_AGENT_MAVEN", imagestream)
-	case imagestream.Name == "jenkins-agent-nodejs":
-		imagestream = tagInPayload("v4.0", "IMAGE_AGENT_NODEJS", imagestream)
+	if g.h != nil {
+		// if we are not skipping jenkins imagestream, tag in images from payload
```
bparees: why does this need to be special cased?

pkg/stub/config.go
```diff
@@ -174,21 +174,21 @@ func (h *Handler) buildSkipFilters(opcfg *v1.Config) {
...
+	for _, si := range opcfg.Status.SkippedImagestreams {
 		newStreamMap[si] = true
 	}
 	h.skippedImagestreams = newStreamMap
 	h.skippedTemplates = newTempMap
 
 }
 
-func (h *Handler) buildFileMaps(cfg *v1.Config) error {
+func (h *Handler) buildFileMaps(cfg *v1.Config, cfgChange bool) error {
```
bparees: maybe rename this to forceRebuild ?  That seems to be its intent.

pkg/stub/interfaces.go
gabemontero: file processing can occur regardless of actual imagestream upserting

pkg/stub/interfaces.go
```diff
@@ -141,13 +142,19 @@ func (g *DefaultImageStreamFromFileGetter) Get(fullFilePath string) (is *imagev1
...
+		if _, ok := g.h.imagestreamFile["jenkins"]; ok {
```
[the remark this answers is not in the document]
gabemontero: I was inclined to think that customers would not be looking to manage the agent streams since they don't come from known samples content from openshift/library

But sure perhaps that is an over reaching assumption?

pkg/stub/config.go
gabemontero: Yeah I went with the "why" vs. "intent" etc.

But there may be new reasons to force in the future ... I'll change

pkg/stub/interfaces.go
gabemontero: I'll add a comment

bparees: but why not just always do this?

pkg/stub/interfaces.go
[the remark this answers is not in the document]
bparees: if we allow people to skip an imagestream by name i think we have to assume they might use that to skip the jenkins agent imagestreams in addition to the master imagestream.  (and conversely, just because they skip jenkins doesn't mean they want to skip the agents)

pkg/stub/interfaces.go
bparees: (ie always update the imagestream w/ the payload pullspec, even if the user is "skipping" the imagestream).

If anything it seems safer than returning an un-mutated imagestream that contains content we know is wrong.

gabemontero: I think there will be cases when users want to import their own jenkins images ... at least at the beginning of 4.0 dev @jupierce relayed to me his team would need this

The agent streams yeah maybe it just makes sense to always mutate

gabemontero: though per the other comment thread maybe we do allow manage separately, and allow to skip :-)

bparees: What I was trying to get at is that I see calling this function as orthogonal to whether the imagestream is being skipped or not.

This function's job(as i understand it) is to return an imagestream for a given filepath.  (and in the case of jenkins, handle mutating the imagestream as needed, though maybe this isn't the right place to do the mutation since presumably the other imagestream mutations, such as to change the registry hostname, are done elsewhere).

It shouldn't care whether a given imagestream is in the skipped list or not, that should be up to the caller to filter out.

So making this function aware of the skiplist, specifically to handle jenkins mutation, seems awkward.

What would happen if this function always mutated the jenkins imagestream before returning it, regardless of the skiplist?

Alternatively, can we put the mutation in the same place other imagestream mutation is done?

gabemontero: Admittedly the path of least resistance was here.  I'll revisit isolating the mutation, placing it elsewhere, and we can compare the awkwardness, etc.

pkg/stub/config.go
gabemontero: part of next push

pkg/stub/interfaces.go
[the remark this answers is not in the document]
gabemontero: this will go away in next push

pkg/stub/interfaces.go
gabemontero: yeah it will be cleaner ... part of next push

gabemontero:
updates pushed @bparees  ... ptal

gabemontero:
cool thx @bparees 

and the "ben 1" commit will live in posterity :-)

bparees:
doh.  i swear github was only showing me one commit. oh well.

PR #98, opened 2019-02-11
copy pull images secret form kube-system by default

gabemontero:
/assign @bparees

### DIFF
--- a/manifests/08-kube-system-rbac.yaml
+++ b/manifests/08-kube-system-rbac.yaml
@@ -1,0 +1,13 @@
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: cluster-samples-operator-kube-system-edit
+  namespace: kube-system
+subjects:
+- kind: ServiceAccount
+  name: cluster-samples-operator
+  namespace: openshift-cluster-samples-operator
+roleRef:
+  kind: ClusterRole
+  name: edit
+  apiGroup: rbac.authorization.k8s.io
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -115,6 +115,8 @@ type Handler struct {
 	mapsMutex sync.Mutex
 
 	secretRetryCount int8
+
+	copyKubeSysPullSecretErr error
 }
 
 // prepSamplesWatchEvent decides whether an upsert of the sample should be done, as well as data for either doing the upsert or checking the status of a prior upsert;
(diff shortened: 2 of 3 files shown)


bparees:
for now the copy logic looks sufficient to get us through the immediate situation.

as a followup (not in this PR) i think we should consider implementing a copy strategy where we watch the kube-system secret and continue to sync it into the openshift namespace if it changes.

If users want to use a different secret for importing, they might as well just create that secret themselves in the openshift namespace (it'll automatically get used).  In retrospect I don't think there's much value in having the samples operator copy a secret from its own namespace into the openshift namespace.

gabemontero:
On Sun, Feb 10, 2019 at 10:02 PM Ben Parees <notifications@github.com>
wrote:

> *@bparees* requested changes on this pull request.
>
> In manifests/08-kube-system-rbac.yaml
> <https://github.com/openshift/cluster-samples-operator/pull/98#discussion_r255373115>
> :
>
> > @@ -0,0 +1,13 @@
> +kind: RoleBinding
> +apiVersion: rbac.authorization.k8s.io/v1
> +metadata:
> +  name: cluster-samples-operator-kube-system-edit
> +  namespace: kube-system
> +subjects:
> +- kind: ServiceAccount
> +  name: cluster-samples-operator
> +  namespace: openshift-cluster-samples-operator
> +roleRef:
> +  kind: ClusterRole
> +  name: edit
>
> i guess view doesn't get access to secrets, but i wonder if it would be
> better to just create a new role that explicitly allows reading
> secrets(assuming there isn't one already) and then granting that. I don't
> really want us to have edit on the kube-system namespace if we can't help
> it.
>
Correct.  I tried view and it did not work.

I can look into doing that Monday either as part of this PR or the follow
up PR you mentioned below.  Just let me know.

gabemontero:
On Sun, Feb 10, 2019 at 10:06 PM Ben Parees <notifications@github.com>
wrote:

> for now the copy logic looks sufficient to get us through the immediate
> situation.
>
> as a followup (not in this PR) i think we should consider implementing a
> copy strategy where we watch the kube-system secret and continue to sync it
> into the openshift namespace if it changes.
>
> If users want to use a different secret for importing, they might as well
> just create that secret themselves in the openshift namespace (it'll
> automatically get used). In retrospect I don't think there's much value in
> having the samples operator copy a secret from its own namespace into the
> openshift namespace.
>
The value such as it arose with the delete scenario with rhel/TBR content.

The change to monitor kube-system vs. operator namespace would be trivial.
And I think everything else could stay the same.

bparees:
@gabemontero ok i'm going to lgtm this to unblock OCP releases, but please do the immediate followup on the RBAC and then we can discuss the creds management.

/lgtm

PR #99, opened 2019-02-11
Kube system cleanup

gabemontero:
/assign @bparees

### DIFF
--- a/pkg/stub/secrets.go
+++ b/pkg/stub/secrets.go
@@ -3,7 +3,9 @@ package stub
 import (
 	"fmt"
 
+	operatorsv1api "github.com/openshift/api/operator/v1"
 	"github.com/openshift/cluster-samples-operator/pkg/apis/samples/v1"
+	"github.com/openshift/cluster-samples-operator/pkg/cache"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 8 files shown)


manifests/08-kube-system-rbac.yaml
```diff
@@ -1,13 +1,30 @@
...
+  - list
+  - watch
+
+---
+
 kind: RoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: cluster-samples-operator-kube-system-edit
+  name: cluster-samples-operator-kube-system-view
```
bparees: cluster-samples-operator-kube-system-secret-reader ?

gabemontero:
oops didn't run make test-unit after getting rid of h.namespace

pkg/stub/secrets.go
```diff
@@ -35,22 +35,25 @@ func (h *Handler) copyDefaultClusterPullSecret() error {
...
+		secret, err = h.secretclientwrapper.Get(coreosPullSecretNamespace, coreosPullSecretName)
 		if err != nil && kerrors.IsNotFound(err) {
 			return nil
 		}
 		if err != nil {
 			return err
 		}
 	}
-	if secret.Name != v1.SamplesRegistryCredentials {
+	if secret.Name != coreosPullSecretName && secret.Name != v1.SamplesRegistryCredentials {
```
bparees: shouldn't this only be the coreosPullSecretName ?

manifests/08-kube-system-rbac.yaml
gabemontero: sure ... consistency

part of next push

pkg/stub/secrets.go
gabemontero: my thought was if they delete the secret in the openshift namespace, we want to recreate, no?

that said, after putting that marker, I forgot to tweak the delete processing below to do that

i'll update and push

bparees: ok.  if this function really can do something intelligent with both the kube-system secret and the openshift secret, this makes more sense (though you should probably be checking the namespace too, then...  otherwise this function risks taking action on a "coreos pull secret" that exists in the "openshift" namespace, for example.

gabemontero: actually I forgot the recreate on delete is handled elsewhere

sorting through some of the flows now with the unit tests

it may in fact turn out we only process coreosPullSecretName in this path

will clarify on repush

gabemontero:
@bparees ptal

I ended up with a small refactor to better account for the differences in copying from kube-system vs. the admin placing the secret in the operator namespace

gabemontero:
OK I picked the reorder fix into this PR

gabemontero:
network operator pain torpedoed install completion

gabemontero:
operator e2e passed @bparees .... more "tal" worthy

gabemontero:
router e2e flake on e2e-aws

gabemontero:
another router flake in e2e-aws

PR #101, opened 2019-02-12
re-order manifests for kube-system rbac is created before operator de…

gabemontero:
…ployment; Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1676438

/assign @bparees


bparees:
the operator deployment has to be created before the clusteroperator resource.

gabemontero:
deployment and clusteroperator swapped @bparees

PR #103, opened 2019-02-13
remove okd/centos

gabemontero:
not 100% sure if removing okd/centos has been officially agreed upon but thought it close enough to get it pushed to start reviewing

/assign @bparees 

ptal

/hold


gabemontero:
it is mainly file deletions and removal of sections of code except for the unit tests and the script for pulling openshift/library content @bparees

gabemontero:
independent of the okd/centos removal, another flake in the operator e2e has been uncovered as the result of an intermittent conflict on config processing, where
a) the e2e update of the config may be fine, but it needs to also check status to make sure the operator has dealt with conflicts and processed
b)  some discipline wrt Status vs. Spec checks is possibly warranted

thinking of handling this in a separate PR and then rebasing this one

@bparees fyi

gabemontero:
on the image ecosystem e2e the machine set operator complained on startup

pkg/apis/samples/v1/types.go
```diff
@@ -440,9 +427,7 @@ func (s *Config) ClusterNeedsCreds() bool {
 		s.Spec.ManagementState == operatorv1.Unmanaged {
 		return false
 	}
-	return s.Spec.InstallType == RHELSamplesDistribution &&
-		s.ConditionFalse(ImportCredentialsExist) &&
-		(s.Spec.SamplesRegistry == "" || s.Spec.SamplesRegistry == "registry.redhat.io")
+	return s.ConditionFalse(ImportCredentialsExist) && (s.Spec.SamplesRegistry == "" || s.Spec.SamplesRegistry == "registry.redhat.io")
```
bparees: Since we now always use the installer pullsecret as the initial creds, i think it's worth revisiting whether we get any value out of the "ImportCredentialsExist" condition anymore?

In addition, it was important to have that condition when we didn't retry imports because if we created the imagestream before the creds existed, the import would fail and never get retried once the creds did exist. Now that we have retries for image imports, it's less critical that we get the ordering right.  (If anything we should just rerun the imports when we update the creds)

Doesn't have to be in this PR, but (assuming you agree) i think it would allow us to remove a lot of complexity from the operator.

pkg/stub/config.go
```diff
@@ -303,8 +285,7 @@ func (h *Handler) ProcessManagementField(cfg *v1.Config) (bool, bool, error) {
 	case operatorsv1api.Managed:
 		if cfg.Spec.ManagementState != cfg.Status.ManagementState {
 			logrus.Println("management state set to managed")
-			if cfg.Spec.InstallType == v1.RHELSamplesDistribution &&
-				cfg.ConditionFalse(v1.ImportCredentialsExist) {
+			if cfg.ConditionFalse(v1.ImportCredentialsExist) {
 				h.copyDefaultClusterPullSecret(nil)
```
bparees: Along the lines of my earlier comment about the CredsExist condition, seems like we should just always do this.  But the copy has to be smart enough not to update the openshift namespace secret if it's unchanged.

pkg/apis/samples/v1/types.go
gabemontero: ImportCredExists provides value if there are problems copying the secret, and facilitates retry of that copy if it failed previously... the recent clean up PR for the kube-system changes brought that element in 

Yes, the motivation for it has changed ... but I'm still inclined to say it has value, and thus we should keep it.

pkg/stub/config.go
gabemontero: Per above, I say we keep the condition, and then the copy can be simple.

pkg/apis/samples/v1/types.go
bparees: I can see the value in using it to indicate issues copying the secret.

I don't agree with using it to facilitate retrying the copy.  That should be handled by the standard relist/resync logic as well as requeueing events.  (ie if we hit an error copying it, we requeue the watch event, and every 10 mins we relist the kube-system secrets and update the openshift secret with its content if it's different).

For the bootstrap itself (where there is no watch event to requeue), doing a few retries on the copy is reasonable, after which if it still fails we'd set the credsexist=false w/ error message and hope that during a resync/relist we succeed in copying.

gabemontero: i'm fine with removing the extra copy attempt on the config watch path if you want, as well as adding a few retries on bootsrap

Separate PR still though, right?

pkg/stub/config.go
gabemontero: Perhaps we just *NEVER* do this here on the config watch path, based on your other response, and just wait for the secret relist/resync to copy

pkg/apis/samples/v1/types.go
bparees: yeah, separate PR.

pkg/stub/config.go
bparees: > Perhaps we just _NEVER_ do this here on the config watch path, based on your other response, and just wait for the secret relist/resync to copy

This one is a bit different since you're going from unmanaged to managed, so it's possible the secret doesn't currently exist in the openshift namespace.  We don't want to wait 10 mins to hit a resync, so something needs to explicitly try to get us back to a working state (similar to bootstrap).

but we can definitely tackle this in a separate PR.  I just think we want to revisit the whole importcredentials/cred management flow given the use of the installer pullsecret.

gabemontero: ok sounds good 

I'll remove the /hold cancel once https://github.com/openshift/cluster-samples-operator/pull/104 merges so we get better operator e2e stability

Feel free to tag this gtm then at your discretion @bparees

gabemontero: I picked https://github.com/openshift/cluster-samples-operator/pull/104 into here to get test runs restarted earlier

gabemontero:
@bparees are you gtm or do you want to take another pass ?

bparees:
@gabemontero the changes lgtm, but i'm giving PM 24 hours to respond with a NAK on my last email.

gabemontero:
10-4 @bparees

gabemontero:
yeah I'm seeing those image-eco scl message missing errors here too

gabemontero:
though they appear intermittent as it finally passed in https://github.com/openshift/cluster-samples-operator/pull/104

bparees:
@gabemontero i'm seeing a variety of failures in your image-eco runs but they all look indicative of an unhealthy cluster (specifically openshift apiserver in some cases), nothing to do w/ the actual scl images or the tests, e.g.:

https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cluster-samples-operator/103/pull-ci-openshift-cluster-samples-operator-master-e2e-aws-image-ecosystem/207

```
E0214 02:08:44.627163     916 memcache.go:140] couldn't get resource list for apps.openshift.io/v1: the server could not find the requested resource
E0214 02:08:44.661156     916 memcache.go:140] couldn't get resource list for build.openshift.io/v1: the server could not find the requested resource
E0214 02:08:44.676213     916 memcache.go:140] couldn't get resource list for image.openshift.io/v1: the server could not find the requested resource
E0214 02:08:44.749355     916 memcache.go:140] couldn't get resource list for route.openshift.io/v1: the server could not find the requested resource
E0214 02:08:44.764242     916 memcache.go:140] couldn't get resource list for security.openshift.io/v1: the server could not find the requested resource
E0214 02:08:44.783221     916 memcache.go:140] couldn't get resource list for template.openshift.io/v1: the server could not find the requested resource
E0214 02:08:44.801226     916 memcache.go:140] couldn't get resource list for user.openshift.io/v1: the server could not find the requested resource
E0214 02:08:44.816238     916 memcache.go:140] couldn't get resource list for packages.apps.redhat.com/v1alpha1: the server could not find the requested resource
```

that points to an openshift apiserver problem.  

/cc @openshift/sig-master

gabemontero:
oooh ... we have a passing image eco @bparees ... and 4 PM is getting close :-)

gabemontero:
rebase cause rename PR merge completed ... readded lgtm label

gabemontero:
@xiuwang unless you prefer Jira, I'll open a bugzilla to track you picking up this change, updating test cases as needed

Issue #100, opened 2019-02-12
operator status Progressing=True long after the cluster reaches idle


sjenning:
I observe the operator stay in `Progressing=True` for 6-8m after the last pod in the cluster (typically telemeter-client) is `Running`.  The installers waits on all operators to be `Progressing=False` before reporting completion.  Happened on 2 of 3 install runs for me.

gabemontero:
Bad registry download performance *could* lead to that @sjenning 

Per https://github.com/openshift/cluster-samples-operator#troubleshooting
running `oc get configs.samples.operator.openshift.io instance -o yaml` would be helpful.

@bparees perhaps the first signs of the ramifications of our switch to rhel and adding the samples operator 's stub cluster status obj into the loop

sjenning:
I'll try to recreate and get this data

gabemontero:
thanks @sjenning

bparees:
There is on-going discussion about what the installer should be waiting for.  Last time I brought this up I was told it was waiting for available=true (no mention of progressing=false), so we updated this operator to change what "available" means so it can claim to be available sooner.  I'm not inclined to do the same with progressing..  the operator is progressing because it is importing the samples it is supposed to import.  Changing that behavior just makes the status reporting less useful.

And there's a new condition ("upgradeable") being proposed that it should be waiting for instead of any of those things.

So before we go making changes to the samples operator to work around this (the only option for us would be to lie about progressing status, which makes our status useless), can we finalize a rule about what the CVO should actually be waiting for so that operators which have lengthy startup processes have a way to not block the install but also not lie about their status?

/cc @deads2k @abhinavdahiya

gabemontero:
I may be seeing this in PR testing from https://github.com/openshift/cluster-samples-operator/pull/99 ... investigating the latest run

gabemontero:
In my PR https://github.com/openshift/cluster-samples-operator/pull/99, it is hitting the manifest ordering issue getting fixed with https://github.com/openshift/cluster-samples-operator/pull/101

I'll have to pick the 101 fix into the 99 PR.

That bug though would always cause an error.  If @sjenning at least had some success, we are back to perf and @bparees 's points / questions at https://github.com/openshift/cluster-samples-operator/issues/100#issuecomment-462824850

[...]

PR #105, opened 2019-02-13
bump cache window to account for timing data from initial installs

gabemontero:
Based on the investigation from https://github.com/openshift/cluster-samples-operator/issues/100 increasing the time we hold onto the startup cache during the initial install could help reduce some of the api churn

Saw a decent amount of `have not received an upsert event for...` messages in my analysis of the startup runs today.

/assign @bparees

### DIFF
--- a/pkg/cache/upserts.go
+++ b/pkg/cache/upserts.go
@@ -38,11 +38,11 @@ func AllUpsertEventsArrived() bool {
 	now := kapis.Now()
 	for key, stream := range upserts {
 		if stream == nil {
-			if now.Sub(timeout.Time) > 5*time.Minute {
+			if now.Sub(timeout.Time) > 15*time.Minute {
 				// if we have not retrieved all the events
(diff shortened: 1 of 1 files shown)


gabemontero:
@bparees - as an fyi, so examining the  run at https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cluster-samples-operator/105/pull-ci-openshift-cluster-samples-operator-master-e2e-aws/375

the change here did reduce the api churn induced by samples operator, but in the end, that did not improve end to end install time

Overall time was 15 min
Parsing the pod logs:
- it took 5 minutes just to be able to create the samples, as api server recycling caused hiccups ... the same messages we saw before in the diagnosis in https://github.com/openshift/cluster-samples-operator/issues/100
- then there was a noticeable pause of another 5 minutes before we started getting watch events on the imagestreams
- then it took 5 minutes for all the imagestreams to report ready

PR #106, opened 2019-02-14
change operator ID from 'instance' to 'cluster'

gabemontero:
/assign @bparees 


gabemontero:
fyi @xiuwang - per request from @bparees the ID for the samples operator config is changing with this PR

PR #108, opened 2019-02-14
Fix mng label when skipped

gabemontero:
this PR also has the remove okd/centos commit to minimize rebase pain

the last 2 commits are the ones really introduced

the first of those is the meets min for fixing the noted bugzilla 

the second is a perf improvement (perhaps the tip of the iceberg ... we can discussion as you deem fit)

/assigng @bparees


bparees:
yeah i'd like to noodle on the optimization a bit more, can you move that to a separate PR and we can merge the first two commits to land the centos-drop and fix the BZ?

gabemontero:
optimization removed; just the bz fix ... and the remove centos/okd in case this merges before https://github.com/openshift/cluster-samples-operator/pull/103

gabemontero:
will post the separate optimization PR in a bit

gabemontero:
rebase cause rename PR merge completed ... readded lgtm label

PR #109, opened 2019-02-15
Upsert opts

gabemontero:
as before, the first 2 commits are other PRs we are trying to merge separately, but touch the same files

/assign @bparees

### DIFF
--- a/pkg/stub/config.go
+++ b/pkg/stub/config.go
@@ -67,94 +67,91 @@ func (h *Handler) SpecValidation(cfg *v1.Config) error {
 // second boolean: does the config change require a samples upsert; for example, simply adding
 // to a skip list does not require a samples upsert
 // third boolean: even if an upsert is not needed, update the config instance to clear out image import errors
-func (h *Handler) VariableConfigChanged(cfg *v1.Config) (bool, bool, bool) {
+// fourth boolean: if the registry changed, that means a) we have to update all imagestreams regardless of skip lists;
+// and b) we don't have to update the templates if that is the only change
+// first map: imagestreams that were unskipped (imagestream updates can be expensive, so an optimization)
+// second map: templates that were unskipped
+func (h *Handler) VariableConfigChanged(cfg *v1.Config) (bool, bool, bool, bool, map[string]bool, map[string]bool) {
+	configChangeAtAll := false
+	configChangeRequireUpsert := false
+	clearImageImportErrors := false
+	registryChange := false
+
+	logrus.Debugf("the before skipped templates %#v", cfg.Status.SkippedTemplates)
+	logrus.Debugf("the after skipped templates %#v and associated map %#v", cfg.Spec.SkippedTemplates, h.skippedTemplates)
+	logrus.Debugf("the before skipped streams %#v", cfg.Status.SkippedImagestreams)
+	logrus.Debugf("the after skipped streams %#v and associated map %#v", cfg.Spec.SkippedImagestreams, h.skippedImagestreams)
+
+	unskippedTemplates := map[string]bool{}
+	// capture additions/subtractions from skipped list in general change boolean
+	if len(cfg.Spec.SkippedTemplates) != len(cfg.Status.SkippedTemplates) {
+		configChangeAtAll = true
+	}
+
+	// build the list of skipped templates; assumes buildSkipFilters called beforehand
+	for _, tpl := range cfg.Status.SkippedTemplates {
+		// if something that was skipped (in Status list) is no longer skipped, build list/capture that fact
+		// this should capture reductions in the skipped list, as well as changing entries, but the number of skips
+		// staying the same
+		if _, ok := h.skippedTemplates[tpl]; !ok {
+			unskippedTemplates[tpl] = true
+			configChangeAtAll = true // set to true in case sizes are equal, but contents have changed
+			configChangeRequireUpsert = true
+		}
+	}
+
+	if configChangeAtAll {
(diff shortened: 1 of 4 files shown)


gabemontero:
actually I might redo this one .... the perf commit has some elements of the bug fix still ...

gabemontero:
ok *NOW* the perf commit has the right set of deltas 

@bparees ptal as time permits (of course next week make sense if necessary)

gabemontero:
rebase for recent merges pushed

gabemontero:
I've got some template related optimizations in mind as well wrt skipped processing like what is here for imagestreams @bparees .... holding off until we've at least officially starting iterating over what was done so far

gabemontero:
actually I have IMO some more improvements to the config changed detection in general as well ...

gabemontero:
ok, unskipped optimization for both streams and templates pushed
also massaged the confg chg detection some ... I think I simplified it 

@bparees ptal

gabemontero:
I believe I've uncovered some in memory concurrency issues updating the Reason field of the import image in progress condition:

```
time="2019-02-19T19:20:02Z" level=info msg="There are no more image imports in flight for imagestream nodejs"
time="2019-02-19T19:20:02Z" level=info msg="imagestream nodejs still not finished with its image imports, including "
```

gabemontero:
Ignore the concurrence comment I just deleted ... I found another spot where I was not checking the existance of a stream in the in progress reason field correctly

gabemontero:
ok @bparees I've pushed 
- fixes to imagestream watch event cache processing that incorrectly left in progress == true until the next resync interval (it was this vs. missed watches that were occurring re: my scrum status this AM)
- additional info around why and imagestream was still in progress
- now always print an indication of any imagestream watch event coming in while we are in progress == true

ptal

pkg/stub/config.go
```diff
@@ -67,94 +67,93 @@ func (h *Handler) SpecValidation(cfg *v1.Config) error {
...
-			if _, ok := h.skippedTemplates[tpl]; !ok {
-				return true, true, false
-			}
+	// see if need to update the cfg from the main loop to clear out any prior image import
+	// errors for skipped streams
+	for _, stream := range cfg.Spec.SkippedImagestreams {
+		importErrors := cfg.Condition(v1.ImportImageErrorsExist)
+		beforeError := cfg.NameInReason(importErrors.Reason, stream)
+		importErrors = h.clearStreamFromImportError(stream, cfg.Condition(v1.ImportImageErrorsExist), cfg)
+		afterError := cfg.NameInReason(importErrors.Reason, stream)
```
bparees: when would it still be there after the clear?  Can it appear multiple times?

pkg/stub/config.go
```diff
@@ -67,94 +67,93 @@ func (h *Handler) SpecValidation(cfg *v1.Config) error {
...
-			logrus.Printf("SkippedTemplates changed in content from %s to %s", cfg.Status.SkippedTemplates[i], skip)
-			return true, true, false
-		}
+	if cfg.Spec.SamplesRegistry != cfg.Status.SamplesRegistry {
+		logrus.Printf("SamplesRegistry changed from %s to %s", cfg.Status.SamplesRegistry, cfg.Spec.SamplesRegistry)
+		configChangeAtAll = true
+		configChangeRequireUpsert = true
+		// skip/unskip imagestream does not matter, reset that list
+		registryChange = true
+		unskippedStreams = map[string]bool{}
```
bparees: could do this up front and immediately return, to avoid all the skipped/unskipped processing, right?

bparees: (presumably we also need to reset any import errors when that happens)

pkg/stub/config.go
gabemontero: it is possible that the skipped streams are NOT in the import error list ... if so, I don't want to bother with an update to the config object ... that is why we track unskips and this separately

bparees: right but in that case beforeError would be false.  When would afterError ever be true, given that you just cleared this imagestream from the errorlist (if i understand the code correctly)

pkg/stub/config.go
gabemontero: no I changed this and moved it from up top because I realized that the registry change does not affect templates

that said, I could do this after the template skip processing and before the imagestream skip processing, 
but then I liked the idea of
- logging the skip changes for dbg
- having only one return statement in this function

But I don't feel super strongly on those 2 points.  Would you like me to move this between the template skips and imagestream skips ?

pkg/stub/config.go
gabemontero: ah you want me to change from 

`if beforeError && !afterError {`

to 

`if beforeError {`

If so, yep, I can do that.

pkg/stub/config.go
gabemontero: I went ahead and made this change while making the one above

gabemontero:
updates pushed @bparees ... also pushed one minor tweak from me

pkg/stub/config.go
bparees: yeah..and there's no need to actually set/retrieve afterError right? (since it's always going to be false)

gabemontero:
bump @bparees

bparees:
@gabemontero squash+lgtm but let's make sure the TBR change merges first, as that's the critical one.

/hold

gabemontero:
ooh ... passing jobs .... can we unhold/merge this one @bparees since api.ci creds aren't coming today?

bparees:
/hold cancel


Ben Parees | OpenShift

PR #110, opened 2019-02-15
Remove ppc

gabemontero:
/assign @bparees 

again included pick from other in flight PRs to minimize rebase pain

last commit is the one ... also there were some openshift/library updates; picked those up while testing the change to the library sync script

@xiuwang FYI ... another cfg change has come down ... leaving the arch field in for future use, but removing the ppc choice since that is now *NOT* on the immediate horizon (we'll see what happens in a year after the acquisition)


gabemontero:
@xiuwang unless you prefer Jira, I'll open a bugzilla to track you picking up this change, updating test cases as needed

gabemontero:
rebase after remove-okd-centos pushed

PR #111, opened 2019-02-20
Goto tbr

gabemontero:
/hold

- this currently includes un squashed https://github.com/openshift/cluster-samples-operator/pull/109 to minimize rebase pain
- for CI to pass we either need to get a TBR cred into CI, or somehow set registry to registry.access.redhat.com during CI
- and ideally we'd like to wait until try.openshift.com's prod deploy has the TBR cred in the install pull secret

/assign @bparees 

but we can iterator on the code changes

### DIFF
--- a/pkg/stub/imagestreams.go
+++ b/pkg/stub/imagestreams.go
@@ -221,10 +221,7 @@ func (h *Handler) upsertImageStream(imagestreamInOperatorImage, imagestreamInClu
 }
 
 func (h *Handler) updateDockerPullSpec(oldies []string, imagestream *imagev1.ImageStream, opcfg *v1.Config) {
-	//TODO remove when TBR creds sorted out and we do not explicitly set SamplesRegistry to registry.access.redhat.com
-	// by default we want to leave the jenkins images as using the payload set to the IMAGE* env's
-	// but if the customer overrides (after we switch to TBR and don't set SamplesRegistry by default), we'll let them
-	// change jenkins related images
+	// we always want to leave the jenkins images as using the payload set to the IMAGE* env's;
 	switch imagestream.Name {
 	case "jenkins":
 		return
@@ -233,29 +230,32 @@ func (h *Handler) updateDockerPullSpec(oldies []string, imagestream *imagev1.Ima
 	case "jenkins-agent-maven":
 		return
 	}
-	if len(opcfg.Spec.SamplesRegistry) > 0 {
-		logrus.Debugf("updateDockerPullSpec stream %s has repo %s", imagestream.Name, imagestream.Spec.DockerImageRepository)
-		// don't mess with deprecated field unless it is actually set with something
-		if len(imagestream.Spec.DockerImageRepository) > 0 &&
-			!strings.HasPrefix(imagestream.Spec.DockerImageRepository, opcfg.Spec.SamplesRegistry) {
-			// if not one of our 4 defaults ...
-			imagestream.Spec.DockerImageRepository = h.coreUpdateDockerPullSpec(imagestream.Spec.DockerImageRepository,
-				opcfg.Spec.SamplesRegistry,
-				oldies)
-		}
 
-		for _, tagref := range imagestream.Spec.Tags {
-			logrus.Debugf("updateDockerPullSpec stream %s and tag %s has from %#v", imagestream.Name, tagref.Name, tagref.From)
-			if tagref.From != nil {
-				switch tagref.From.Kind {
-				// ImageStreamTag and ImageStreamImage will ultimately point to a DockerImage From object reference
-				// we are only updating the actual registry pull specs
-				case "DockerImage":
-					if !strings.HasPrefix(tagref.From.Name, opcfg.Spec.SamplesRegistry) {
-						tagref.From.Name = h.coreUpdateDockerPullSpec(tagref.From.Name,
-							opcfg.Spec.SamplesRegistry,
-							oldies)
-					}
+	if len(opcfg.Spec.SamplesRegistry) == 0 {
+		return
+	}
+
+	logrus.Debugf("updateDockerPullSpec stream %s has repo %s", imagestream.Name, imagestream.Spec.DockerImageRepository)
+	// don't mess with deprecated field unless it is actually set with something
+	if len(imagestream.Spec.DockerImageRepository) > 0 &&
+		!strings.HasPrefix(imagestream.Spec.DockerImageRepository, opcfg.Spec.SamplesRegistry) {
+		// if not one of our 4 defaults ...
+		imagestream.Spec.DockerImageRepository = h.coreUpdateDockerPullSpec(imagestream.Spec.DockerImageRepository,
+			opcfg.Spec.SamplesRegistry,
+			oldies)
+	}
+
+	for _, tagref := range imagestream.Spec.Tags {
+		logrus.Debugf("updateDockerPullSpec stream %s and tag %s has from %#v", imagestream.Name, tagref.Name, tagref.From)
+		if tagref.From != nil {
+			switch tagref.From.Kind {
+			// ImageStreamTag and ImageStreamImage will ultimately point to a DockerImage From object reference
+			// we are only updating the actual registry pull specs
+			case "DockerImage":
+				if !strings.HasPrefix(tagref.From.Name, opcfg.Spec.SamplesRegistry) {
+					tagref.From.Name = h.coreUpdateDockerPullSpec(tagref.From.Name,
+						opcfg.Spec.SamplesRegistry,
+						oldies)
 				}
 			}
 		}
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ The samples resource offers the following configuration fields:
 - ManagementState
 -- Managed: the operator will update the samples as the configuration dictates
 -- Unmanaged: the operator will ignore updates to the samples resource object and any imagestreams or templates in the openshift namespace
--- Removed: the operator will remove the set of managed imagestreams and templates in the openshift namespace. It will ignore new samples created by the cluster admin or any samples in the skipped lists.  After the removals are complete, the operator will work like it is in the 'Unmanaged' state and ignore any watch events on the sample resources, imagestreams, or templates.  It will operate on secrets to facilitate the 'centos' to 'rhel' switch.  There are some caveats around concurrent creates and removal (see Change behaviors section).
+-- Removed: the operator will remove the set of managed imagestreams and templates in the openshift namespace. It will ignore new samples created by the cluster admin or any samples in the skipped lists.  After the removals are complete, the operator will work like it is in the 'Unmanaged' state and ignore any watch events on the sample resources, imagestreams, or templates.  There are some caveats around concurrent creates and removal (see Config behaviors section).
 - Samples Registry
 -- Override the registry that images are imported from
 - Architecture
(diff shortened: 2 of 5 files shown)


gabemontero:
Just got an update ... tbr cred in production try.openshift ... awaiting dev productivity update on obtaining cred for CI

gabemontero:
some in parallel unit testing isn't looking right either ... we'll keep hold regardless of CI changes until I sort out ... most likely new commit coming

pkg/stub/imagestreams.go
```diff
@@ -186,7 +186,7 @@ func (h *Handler) upsertImageStream(imagestreamInOperatorImage, imagestreamInClu
 		return nil
 	}
 
-	h.updateDockerPullSpec([]string{"docker.io", "registry.redhat.io", "registry.access.redhat.com", "quay.io"}, imagestreamInOperatorImage, opcfg)
+	h.updateDockerPullSpec([]string{"docker.io", "registry.access.redhat.com", "quay.io"}, imagestreamInOperatorImage, opcfg)
```
bparees: i don't think this should be changing?

pkg/stub/imagestreams.go
```diff
@@ -221,41 +221,44 @@ func (h *Handler) upsertImageStream(imagestreamInOperatorImage, imagestreamInClu
...
-	if len(opcfg.Spec.SamplesRegistry) > 0 {
-		logrus.Debugf("updateDockerPullSpec stream %s has repo %s", imagestream.Name, imagestream.Spec.DockerImageRepository)
-		// don't mess with deprecated field unless it is actually set with something
-		if len(imagestream.Spec.DockerImageRepository) > 0 &&
-			!strings.HasPrefix(imagestream.Spec.DockerImageRepository, opcfg.Spec.SamplesRegistry) {
-			// if not one of our 4 defaults ...
-			imagestream.Spec.DockerImageRepository = h.coreUpdateDockerPullSpec(imagestream.Spec.DockerImageRepository,
-				opcfg.Spec.SamplesRegistry,
-				oldies)
+	registryReplacement := "registry.redhat.io"
```
bparees: isn't this going to default into replacing everything with registry.redhat.io even if they specified no replacement?  why do we want that?

pkg/stub/imagestreams.go
```diff
@@ -221,41 +221,44 @@ func (h *Handler) upsertImageStream(imagestreamInOperatorImage, imagestreamInClu
...
+		// in our default case we want to leave the jenkins images as using the payload set to the IMAGE* env's;
+		// but if they explicitly set SamplesRegistry to registry.redhat.io, we'll let them explicity point to there
+		// in case the payload ever pointed to something different (though that is not the expectation for 4.0)
+		switch imagestream.Name {
+		case "jenkins":
+			return
+		case "jenkins-agent-nodejs":
+			return
+		case "jenkins-agent-maven":
+			return
```
bparees: this doesn't seem to do what the comment says.... it looks like it'll only leave the jenkins imagestreams alone if no registry was specified.  But it says the intent is to leave the jenkins imagestreams alone unless the registry is explicitly set to registry.redhat.io.

In any case my personal take is we should just never touch the jenkins imagestreams.  They're part of the payload, so the users don't get to control where it should be imported from.  If we ever realize that's a mistake, we can introduce new behavior in an upgrade :)

pkg/stub/imagestreams.go
gabemontero: this is how it defaults to registry.redhat.io

pkg/stub/imagestreams.go
gabemontero: this was suppose to be an optimization with the default being registry.redhat.io now but you are right / I forgot if they override the registry we want to replace those too

that array string is things we want to replace

pkg/stub/imagestreams.go
gabemontero: yeah ... and the can always mark them in the skip list if they want to change them ... will change

pkg/stub/imagestreams.go
gabemontero: it will now look like:

```
	switch imagestream.Name {
	case "jenkins":
		return
	case "jenkins-agent-nodejs":
		return
	case "jenkins-agent-maven":
		return
	}

	registryReplacement := "registry.redhat.io"
	if len(opcfg.Spec.SamplesRegistry) != 0 {
		registryReplacement = opcfg.Spec.SamplesRegistry
	}
```

As a reminder, back in the early days, we did this as a way to catch any stray imagestreams still pointing to docker.io or registry.access.redhat.com, etc.

pkg/stub/imagestreams.go
gabemontero: part of next push

gabemontero:
updates pushed @bparees 

including the fix needed from that parallel unit testing I mentioned earlier

pkg/stub/imagestreams.go
bparees: I don't think we want it to default to anything.  If the user didn't provide an alternate registry value, we should leave the imagestreams alone.  If they did provide one, we should use the value they provided to replace our canonical list, except for jenkins.

> As a reminder, back in the early days, we did this as a way to catch any stray imagestreams still pointing to docker.io or registry.access.redhat.com, etc.

hm, I don't recall that and i'm not sure why we we'd want to change them?  If they're pointing there, we should fix the imagestream content.  Otherwise, if it points to docker.io there may well be a good reason (it may not be on registry.redhat.io).

gabemontero: the user doesn't provide these ... this comes from the openshift/library content we pull in

unless the user overrides the registry, he expects it to be correct

bparees: > the user doesn't provide these ... this comes from the openshift/library content we pull in

right, hence my question above of why we'd want to change them, absent an explicit user-provided alternate registry to use?

> unless the user overrides the registry, he expects it to be correct

Exactly, and the samples operator should also expect them to be correct.  We don't own the imagestreams, we get them from elsewhere, so if someone provides us with an imagestream and says "this comes from docker.io" (or quay.io, or something else) we should be respecting that value, absent an explicit override from the user.  Defaulting the override registry hostname to "registry.redhat.io" (or any other value) doesn't make sense to me...

gabemontero: OK, yeah there was originally a concern with openshift/library OCP content providing non TBR content and we should catch that.  And the code was doing that previously.

I'm fine with moving off of that now as part of this PR.  I'll push a change in a bit.

bparees: > OK, yeah there was originally a concern with openshift/library OCP content providing non TBR content and we should catch that. And the code was doing that previously.

yeah i can't remember that convo or the context, but let's go with "i've come to a new understanding and changed my mind" :)

gabemontero:
update pushed @bparees ... ptal

gabemontero:
I also removed the commits from https://github.com/openshift/cluster-samples-operator/pull/109 since we are looking more like we would want to merge this before wrapping up that one ... if so will rebase 109 as needed then

gabemontero:
tracking tbr creds in api.ci via https://coreos.slack.com/messages/CBN38N3MW/convo/CBN38N3MW-1550767187.131600/

pkg/stub/imagestreams.go
```diff
@@ -222,29 +219,32 @@ func (h *Handler) updateDockerPullSpec(oldies []string, imagestream *imagev1.Ima
...
+		logrus.Debugf("updateDockerPullSpec stream %s and tag %s has from %#v", imagestream.Name, tagref.Name, tagref.From)
+		if tagref.From != nil {
+			switch tagref.From.Kind {
+			// ImageStreamTag and ImageStreamImage will ultimately point to a DockerImage From object reference
+			// we are only updating the actual registry pull specs
+			case "DockerImage":
+				if !strings.HasPrefix(tagref.From.Name, opcfg.Spec.SamplesRegistry) {
+					tagref.From.Name = h.coreUpdateDockerPullSpec(tagref.From.Name,
+						opcfg.Spec.SamplesRegistry,
+						oldies)
```
bparees: future improvement:  parse the pullspec and replace the registryhostname section, rather than searching+replacing specific strings.  That would future proof this against imagestreams that reference additional registries beyond what "oldies" knows about.

same for the logic above.

gabemontero: I'll make a note to add a TODO in perf/fix PR

gabemontero:
ok thx @bparees 

I'll remove the hold 
a) when I get confirmation from @stevekuznetsov that the cred is in CI
b) I send out the notice on aos-devel/openshift-sme/slack that folks should go to try.openshift.com and update their pull secret as part of this PR merging

sound good?

bparees:
> a) when I get confirmation from @stevekuznetsov that the cred is in CI

presumably this won't pass CI until that happens anyway, but yeah..good call.  Certainly not reason to let CI beat its head against a wall until we expect it to work.

> b) I send out the notice on aos-devel/openshift-sme/slack that folks should go to try.openshift.com and update their pull secret as part of this PR merging

also a good call.  I was thinking about the need to tell people about this last night and forgot about it.  They might as well go update their secret now, no reason to wait.

gabemontero:
I have a draft of the email ready, but want to do some tests with this PRs changes and the TBR creds not present to confirm the behavior.

We've still have failing not going to true on image stream import failures, so the install should still at least complete, but I want to verify that and include those details in the email.

gabemontero:
Confirmed failing still stays false even with the image import errors

total install time went from 1.5 minutes typically to 2 minutes in a couple of my trial runs

will send email shortly @bparees

bparees:
sounds good, thanks for doing the extra testing

gabemontero:
the PR's prow/tide responsiveness seems hosed (whereas my other one is OK) ... going through the various remedies ...

gabemontero:
there we go ... tests are running

gabemontero:
OK we got cluster coming up and both e2e-aws and e2e-image-ecosystem showed failures as a result of the imagestreams *NOT* importing

So validation of understanding

We'll see once happens once they get the bandwidth to add the cred to api.ci

gabemontero:
rebased after other PRs merged

bparees:
@gabemontero you're not supposed to have to cherrypick, they are fastforwarding from master.

gabemontero:
yeah @bparees but I can always close the PR it opens ... this is just in case they stop fast forwarding before this merges

gabemontero:
e2e-aws failure typical aws could not get off the ground:

```
level=error msg="Error: Error applying plan:"
level=error
level=error msg="2 errors occurred:"
level=error msg="\t* module.vpc.aws_route_table_association.private_routing[3]: 1 error occurred:"
level=error msg="\t* aws_route_table_association.private_routing.3: timeout while waiting for state to become 'success' (timeout: 5m0s)"
level=error
level=error
level=error msg="\t* module.vpc.aws_route_table_association.private_routing[2]: 1 error occurred:"
level=error msg="\t* aws_route_table_association.private_routing.2: timeout while waiting for state to become 'success' (timeout: 5m0s)"
level=error
level=error
level=error
level=error
level=error
level=error msg="Terraform does not automatically rollback in the face of errors."
level=error msg="Instead, your Terraform state file has been partially updated with"
level=error msg="any resources that successfully completed. Please address the error"
level=error msg="above and apply again to incrementally change your infrastructure."
```

image eco passed !!!! thanks @bparees @stevekuznetsov @bbguimaraes

PR #113, opened 2019-02-26
use pumped up rate limiter, shave 30 seconds from startup creates

gabemontero:
/assign @bparees 

@smarterclayton fyi / ptal if you have bandwidth


gabemontero:
fyi @bparees does not conflict with tbr PR

PR #115, opened 2019-02-27
Cvo tweaks

gabemontero:
/assign @bparees 

in addition to the fix discussed in slack, I'm addressing one of Seth's findings re cvo status (testing that lead me to discover the startup regression ... confirmed first commit did not cause regressions)

### DIFF
--- a/pkg/apis/samples/v1/types.go
+++ b/pkg/apis/samples/v1/types.go
@@ -365,6 +365,10 @@ func (s *Config) ClusterOperatorStatusAvailableCondition() (configv1.ConditionSt
 // 2) the first string is the succinct text to apply to the Progressing condition on failure
 // 3) the second string is the fully detailed text to apply the the Failing condition
 func (s *Config) ClusterOperatorStatusFailingCondition() (configv1.ConditionStatus, string, string) {
+	if len(s.Status.Conditions) == 0 {
+		// first event, have not processed default config yet
+		return configv1.ConditionFalse, "", ""
+	}
 	// the ordering here is not random; an invalid config will be caught first;
 	// the lack of credenitials will be caught second; any hiccups manipulating API objects
 	// will be potentially anywhere in the process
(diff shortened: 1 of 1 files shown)


pkg/apis/samples/v1/types.go
```diff
@@ -424,9 +428,23 @@ func (s *Config) ClusterNeedsCreds() bool {
 		s.Spec.ManagementState == operatorv1.Unmanaged {
 		return false
 	}
-	if s.Status.Conditions == nil {
+	if s.Status.Conditions == nil || len(s.Status.Conditions) == 0 {
```
bparees: the length check is redundant since your for loop below will no-op if the length is zero.

pkg/apis/samples/v1/types.go
```diff
@@ -424,9 +428,23 @@ func (s *Config) ClusterNeedsCreds() bool {
...
+	// so explicitly check it the import creds condition is even there yet
+	foundImportCred := false
+	for _, rc := range s.Status.Conditions {
+		if rc.Type == ImportCredentialsExist {
+			foundImportCred = true
+			break
+		}
+	}
+	if !foundImportCred {
+		return true
```
bparees: why doesn't this return value care what the samples registry is?

pkg/apis/samples/v1/types.go
```diff
@@ -424,9 +428,23 @@ func (s *Config) ClusterNeedsCreds() bool {
 		s.Spec.ManagementState == operatorv1.Unmanaged {
 		return false
 	}
-	if s.Status.Conditions == nil {
+	if s.Status.Conditions == nil || len(s.Status.Conditions) == 0 {
 		return true
```
bparees: why doesn't this return value care what the samples registry is?

pkg/apis/samples/v1/types.go
gabemontero: update pushed

pkg/apis/samples/v1/types.go
bparees: (seems like you can add an early check that checks what the samples registry even is.  if it's not registry.redhat.io you can short circuit return false)

pkg/apis/samples/v1/types.go
gabemontero: addressed in new push

pkg/apis/samples/v1/types.go
gabemontero: addressed in new push

gabemontero:
install did not get off the ground, failing in the same fashion, for all three e2e's last go around... samples operator deployment not even started

PR #119, opened 2019-02-27
publish operator version from the payload

bparees:
aligning with https://github.com/openshift/cluster-image-registry-operator/pull/219

### DIFF
--- a/pkg/operatorstatus/operatorstatus.go
+++ b/pkg/operatorstatus/operatorstatus.go
@@ -2,6 +2,7 @@ package operator
 
 import (
 	"fmt"
+	"reflect"
 
 	"github.com/sirupsen/logrus"
 
@@ -58,18 +59,19 @@ type ClusterOperatorWrapper interface {
 }
 
 func (o *ClusterOperatorHandler) UpdateOperatorStatus(cfg *v1.Config) error {
-
 	var err error
-	failing, stateForPending, msgForFailing := cfg.ClusterOperatorStatusFailingCondition()
+	failing, msgForProgressing, msgForFailing := cfg.ClusterOperatorStatusFailingCondition()
 	err = o.setOperatorStatus(configv1.OperatorFailing,
 		failing,
 		msgForFailing,
-		v1.GitVersionString())
+		"")
 	if err != nil {
 		return err
 	}
 
 	available, msgForAvailable := cfg.ClusterOperatorStatusAvailableCondition()
+	// if we're setting the operator status to available, also set the operator version
+	// to the current version.
 	err = o.setOperatorStatus(configv1.OperatorAvailable,
 		available,
 		msgForAvailable,
@@ -78,11 +80,11 @@ func (o *ClusterOperatorHandler) UpdateOperatorStatus(cfg *v1.Config) error {
 		return err
 	}
 
-	progressing, msgForProgressing := cfg.ClusterOperatorStatusProgressingCondition(stateForPending, available)
+	progressing, msgForProgressing := cfg.ClusterOperatorStatusProgressingCondition(msgForProgressing, available)
 	err = o.setOperatorStatus(configv1.OperatorProgressing,
 		progressing,
 		msgForProgressing,
-		cfg.Status.Version)
+		"")
 	if err != nil {
 		return nil
 	}
@@ -90,7 +92,7 @@ func (o *ClusterOperatorHandler) UpdateOperatorStatus(cfg *v1.Config) error {
 }
 
 func (o *ClusterOperatorHandler) setOperatorStatus(condtype configv1.ClusterStatusConditionType, status configv1.ConditionStatus, msg, currentVersion string) error {
-	logrus.Debugf("setting clusteroperator status condition %s to %s", condtype, status)
+	logrus.Debugf("setting clusteroperator status condition %s to %s with version %s", condtype, status, currentVersion)
 	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
 		state, err := o.ClusterOperatorWrapper.Get(ClusterOperatorName)
 		if err != nil {
@@ -98,6 +100,7 @@ func (o *ClusterOperatorHandler) setOperatorStatus(condtype configv1.ClusterStat
 				return fmt.Errorf("failed to get cluster operator resource %s/%s: %s", state.ObjectMeta.Namespace, state.ObjectMeta.Name, err)
 			}
 
+			state = &configv1.ClusterOperator{}
 			state.Name = ClusterOperatorName
 
 			state.Status.Conditions = []configv1.ClusterOperatorStatusCondition{
@@ -118,6 +121,15 @@ func (o *ClusterOperatorHandler) setOperatorStatus(condtype configv1.ClusterStat
 				},
 			}
 
+			if len(currentVersion) > 0 {
+				state.Status.Versions = []configv1.OperandVersion{
+					{
+						Name:    "operator",
+						Version: currentVersion,
+					},
+				}
+			}
+
 			o.updateOperatorCondition(state, &configv1.ClusterOperatorStatusCondition{
 				Type:               condtype,
 				Status:             status,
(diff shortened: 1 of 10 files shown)


bparees:
/assign @gabemontero 

I didn't end up changing anything significant about the clusteroperator condition reporting right now.  We should probably talk through that (again).

pkg/operatorstatus/operatorstatus.go
```diff
@@ -58,18 +59,19 @@ type ClusterOperatorWrapper interface {
...
 func (o *ClusterOperatorHandler) UpdateOperatorStatus(cfg *v1.Config) error {
-
 	var err error
-	failing, stateForPending, msgForFailing := cfg.ClusterOperatorStatusFailingCondition()
+	failing, msgForProgressing, msgForFailing := cfg.ClusterOperatorStatusFailingCondition()
 	err = o.setOperatorStatus(configv1.OperatorFailing,
 		failing,
 		msgForFailing,
-		v1.GitVersionString())
+		"")
```
gabemontero: I know this has been a moving target, but at one point, the direction was to include the version we failed to get to ... I'll see if the doc we reference before has changed, but was this a conscious choice?

pkg/operatorstatus/operatorstatus.go
```diff
@@ -78,26 +80,27 @@ func (o *ClusterOperatorHandler) UpdateOperatorStatus(cfg *v1.Config) error {
 		return err
 	}
 
-	progressing, msgForProgressing := cfg.ClusterOperatorStatusProgressingCondition(stateForPending, available)
+	progressing, msgForProgressing := cfg.ClusterOperatorStatusProgressingCondition(msgForProgressing, available)
 	err = o.setOperatorStatus(configv1.OperatorProgressing,
 		progressing,
 		msgForProgressing,
-		cfg.Status.Version)
+		"")
```
gabemontero: similarly we aren't saying we are progressing to given version any more?

pkg/stub/handler.go
```diff
@@ -77,6 +78,7 @@ func NewSamplesOperatorHandler(kubeconfig *restclient.Config) (*Handler, error)
 	h.templateFile = make(map[string]string)
 
 	h.mapsMutex = sync.Mutex{}
+	h.version = os.Getenv("RELEASE_VERSION")
```
gabemontero: is caching this way really safe .... I had done it before with the prior version and found problems (can't remember the specifics off hand ... will have to go back and look.

Also, actually running through the flow probably would answer this question for me as well.

bparees:
> Ahhh ... not that I'm surprised to see this come down, but the git commit based stuff had the advantage of only inducing upsert churn when we actually changed the content

Yeah, that is a valid concern with this and it's going to affect all the operators/operands (basically any time a new release payload is created, all the images will be new, meaning everything gets redeployed even if it had no changes).  So it's worth discussing, but it's a global problem.  

In this case i think all it means is we'll upsert all the content(even if it didn't actually change).  Worst case that might trigger an imagestream import, which is basically a no-op unless there is actually a new image available for an existing tag.  If there is, it will update the tag to point to the new image sha and potentially trigger deployments/builds that were watching that tag.  But of course that could have happened if we delivered a new samples operator due to a code fix too.


> I still have not seen the recipe for testing upgrades this way. Can you specify @bparees ?

Since the "version" comes from an env var on the operator, you can test an upgrade by taking the operator out of CVO management and changing the version env var on the operator deployment.

That will make the operator restart (to pick up the env var change), see itself as a new version, and see all the existing imagestreams as not having the current version, thus going through all the upsert logic.  That said, you know that logic better than I do, so we should definitely talk through the expectations there.

pkg/operatorstatus/operatorstatus.go
bparees: That should still happen.  This argument is only used to set the clusteroperator.status.version field, it's not used to generate the failing message, which is where you put the "failed to achieve version X" message, right?

The change here is we only set clusteroperator.status.version to the operator's current version when we are setting available=true.

Until available=true, we are considered to be at the previous version.

pkg/operatorstatus/operatorstatus.go
bparees: same answer..the reason for being in progressing should still contain the "trying to achieve version $currentVersion" message.

pkg/stub/handler.go
bparees: You'd have a  better idea than I, but I can't see how it wouldn't be...  we only create the handler once, when the operator/controller starts up, right?  And the env variable value can't change unless the operator pod gets redeployed w/ a new env value (at which point we'd be restarting and pick up the new value).

bparees:
(If we talk through this and you're on board that it's workable, we can add an e2e test that actually does set the env var to some new version value and confirms that all the conditions get reported the way we'd expect, imagestreams get upserted w/ the new version annotation, etc)

pkg/stub/handler.go
gabemontero: based on how my local testing goes, if need be, I'll revisit my PRs/commits in this space and see

bparees:
> ok cool re: e2e tests using the env ... that piece is better/easier than before
> an additional comment occurred to me ... the Makefile could be updated to remove the obtaining of the git version information, since that is no longer used

Can I ask you to pick this up/push it over the hump (adding the e2e test, makefile cleanup)? I need to pivot onto the image registry operator which has much more significant challenges around status reporting right now.

pkg/operatorstatus/operatorstatus.go
gabemontero: My testing of the PR is not showing what you say:

```
status:
  conditions:
  - lastTransitionTime: 2019-02-28T18:00:17Z
    message: 'Samples installation successful at '
    status: "True"
    type: Available
  - lastTransitionTime: 2019-02-28T18:01:10Z
    message: 'Samples installation successful at '
    status: "False"
    type: Progressing
  - lastTransitionTime: 2019-02-28T18:00:14Z
    status: "False"
    type: Failing
  extension: null
  relatedObjects: null
  versions: null
```

My recollection was in fact correct.  You need to pass the version in that parameter.

pkg/operatorstatus/operatorstatus.go
gabemontero: same response as above

gabemontero:
OK testing the PR now @bparees 

The first few samples of the cvo status cited config errors:
```
apiVersion: config.openshift.io/v1
kind: ClusterOperator
metadata:
  creationTimestamp: 2019-02-28T17:59:52Z
  generation: 1
  name: openshift-samples
  resourceVersion: "122551"
  selfLink: /apis/config.openshift.io/v1/clusteroperators/openshift-samples
  uid: a5c7b7b1-3b82-11e9-885f-06195b77e04c
spec: {}
status:
  conditions:
  - lastTransitionTime: 2019-02-28T17:59:52Z
    status: "False"
    type: Available
  - lastTransitionTime: 2019-02-28T17:59:55Z
    message: 'Samples installation in error at : invalid configuration'
    status: "True"
    type: Progressing
  - lastTransitionTime: 2019-02-28T17:59:55Z
    message: 'Samples installation in error at : '
    status: "True"
    type: Failing
  extension: null
  relatedObjects: null
  versions: null
apiVersion: config.openshift.io/v1
kind: ClusterOperator
metadata:
  creationTimestamp: 2019-02-28T17:59:52Z
  generation: 1
  name: openshift-samples
  resourceVersion: "122949"
  selfLink: /apis/config.openshift.io/v1/clusteroperators/openshift-samples
  uid: a5c7b7b1-3b82-11e9-885f-06195b77e04c
spec: {}
status:
  conditions:
  - lastTransitionTime: 2019-02-28T17:59:52Z
    status: "False"
    type: Available
  - lastTransitionTime: 2019-02-28T17:59:55Z
    message: 'Samples installation in error at : invalid configuration'
    status: "True"
    type: Progressing
  - lastTransitionTime: 2019-02-28T18:00:14Z
    status: "False"
    type: Failing
  extension: null
  relatedObjects: null
  versions: null
```

gabemontero:
Also, in addition to the available/progressing messages looking weird because of the missing version, the versions field itself is null when everything is done:

`versions: null`

was that intentional ?

pkg/operatorstatus/operatorstatus.go
bparees: I still don't think so.  That argument is not used to set the message.  It wasn't before and it's not now.

gabemontero: Nope I was wrong ... needed to update my deployment

pkg/operatorstatus/operatorstatus.go
gabemontero: Nope I was wrong ... needed to update my deployment

gabemontero:
needed to update my deployment re: versions

gabemontero:
ok my basic testing looks good ... @bparees and I decided to close this PR ... I've got his commit in my own branch, and I'll update Makefile, address cfg invalid at start, add e2e's, do more testing, etc.

PR #122, opened 2019-03-04
BUG 1685228: switch pull policy to if not present per clayton

gabemontero:
/hold

will reword commit once bugzilla comes in

but getting testing started in the interim


gabemontero:
fyi did get clean test runs prior to adding in bug number to commit msg / pr title

PR #127, opened 2019-03-07
enable must-gather object refs

gabemontero:
Fixes https://github.com/openshift/cluster-samples-operator/issues/124

/assign @bparees 
/assign @adambkaplan 

### DIFF
--- a/pkg/operatorstatus/operatorstatus.go
+++ b/pkg/operatorstatus/operatorstatus.go
@@ -103,6 +103,12 @@ func (o *ClusterOperatorHandler) setOperatorStatus(condtype configv1.ClusterStat
 			state = &configv1.ClusterOperator{}
 			state.Name = ClusterOperatorName
 
+			state.Status.RelatedObjects = []configv1.ObjectReference{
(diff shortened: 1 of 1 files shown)


pkg/operatorstatus/operatorstatus.go
```diff
@@ -103,6 +103,11 @@ func (o *ClusterOperatorHandler) setOperatorStatus(condtype configv1.ClusterStat
 			state = &configv1.ClusterOperator{}
 			state.Name = ClusterOperatorName
 
+			state.Status.RelatedObjects = []configv1.ObjectReference{
+				{Group: v1.GroupName, Resource: "configs", Name: "cluster"},
+				{Resource: "namespaces", Name: "openshift-cluster-samples-operator"},
```
bparees: you probably want to gather the openshift namespace too.

gabemontero: good call

gabemontero: update pushed @bparees

PR #130, opened 2019-04-09
BUG 1697644: operator: use openshift-config/pull-secret to fetch the pull secret

abhinavdahiya:
The operator should be using the openshift-config/pull-secret as the kube-system/coreos-pull-secret has been marked deprecated.

### DIFF
--- a/manifests/05-kube-system-rbac.yaml
+++ b/manifests/05-kube-system-rbac.yaml
@@ -2,7 +2,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
   name: coreos-pull-secret-reader
-  namespace: kube-system
+  namespace: openshift-config
 rules:
 - apiGroups:
   - ""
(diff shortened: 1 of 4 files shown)


gabemontero:
Ah the move I've longer heard rumoured is actually happening

gabemontero:
@bmcelvee  - I checked the readme and we had not direct refs to the kube-system namespace.  I can't seem the find the link to the official docs.  Can you refresh my memory, so we make sure we don't reference kube-system there as well ?

bmcelvee:
> @bmcelvee - I checked the readme and we had not direct refs to the kube-system namespace. I can't seem the find the link to the official docs. Can you refresh my memory, so we make sure we don't reference kube-system there as well ?

@gabemontero we don't reference "kube-system" in the Samples Operator docs, but it's referenced in other places in the documentation.

gabemontero:
@bmcelvee ok thanks for the confirmation ... wrt this PR, I am only worried about the sample operator doc ... but presumably the larger change that prompted @abhinavdahiya here will drive the clean up elsewhere

if anything else hopefully the doc's team is now "on alert" wrt this change

abhinavdahiya:
With https://github.com/openshift/cluster-samples-operator/compare/91b81a381442b44f0c2343145e1c0327b55832ad...e0623f70c4d5c8b56d456618f585e3b40a200312 updated the rbac in manifests to allow operator to read secrets from openshift-config

gabemontero:
dammit should have caught that

gabemontero:
the image ecosystem run failed on cluster startup... machine-config and network operators unhappy:

```
machine-config                                              Available=False, Failing=True, Progressing=True
network                                                     Available=False, Failing=False, Progressing=True
```

PR #137, opened 2019-04-22
simplify, be explicit, on imagestream update reporting; don't miss up…

gabemontero:
…dating of progressing condition until relist

address delay in setting progressing=false during seen in upgrade testing at https://openshift-gce-devel.appspot.com/build/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-upgrade-4.0/1077

### DIFF
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -30,7 +30,7 @@ import (
 	templatev1client "github.com/openshift/client-go/template/clientset/versioned/typed/template/v1"
 
 	operatorsv1api "github.com/openshift/api/operator/v1"
-	"github.com/openshift/cluster-samples-operator/pkg/apis/samples/v1"
+	v1 "github.com/openshift/cluster-samples-operator/pkg/apis/samples/v1"
 	"github.com/openshift/cluster-samples-operator/pkg/cache"
 	operatorstatus "github.com/openshift/cluster-samples-operator/pkg/operatorstatus"
 
@@ -726,6 +726,8 @@ func (h *Handler) Handle(event v1.Event) error {
 		// cycle through them here now
 		if cache.AllUpsertEventsArrived() && cfg.ConditionTrue(v1.ImageChangesInProgress) {
 			keysToClear := []string{}
+			anyChange := false
+			ac := false
 			for key, is := range cache.GetUpsertImageStreams() {
 				if is == nil {
 					// never got update, refetch
@@ -736,7 +738,8 @@ func (h *Handler) Handle(event v1.Event) error {
 						continue
 					}
 				}
-				cfg, _ = h.processImportStatus(is, cfg)
+				cfg, _, ac = h.processImportStatus(is, cfg)
+				anyChange = anyChange || ac
 			}
 			for _, key := range keysToClear {
 				cache.RemoveUpsert(key)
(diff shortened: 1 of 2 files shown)


gabemontero:
`Unauthorized` and other flakes with e2e-aws/e2e-aws-image-ecosystem

/retest

gabemontero:
/assign @adambkaplan 

So the regression run here looked good wrt marking image stream imports complete:

First a specific imagestream event:
```
time="2019-04-22T20:50:31Z" level=info msg="Imagestream fis-karaf-openshift watch event do upsert false; no errors in prep true,  possibly update operator conditions true"
```

Followed eventually by the completion of that event:

```
time="2019-04-22T20:50:31Z" level=info msg="The last in progress imagestream has completed (import status check)"
time="2019-04-22T20:50:31Z" level=info msg="There are no more image imports in flight for imagestream jboss-fuse70-eap-openshift"
time="2019-04-22T20:50:31Z" level=info msg="The last in progress imagestream has completed (imagestream event, no update needed)"
time="2019-04-22T20:50:31Z" level=info msg="CRDUPDATE updating progress/error condition after results for fis-karaf-openshift"
```

Though that does not exhibit the timing window see in the "slow" upgrade job where if there is enough concurrency the last event and the post processing do not match.

If it helps next time you are in the office I can show you some log compares to better illustrate.

I'm going to do a few more runs to see if we get lucky and reproduced the problematic timing during this operator's migration test here.  

But IMO this change is safe and should address what I saw in the "delayed" logs.

adambkaplan:
/lgtm

think these are safe enough

gabemontero:
/hold

sorry realized this AM I've got another piece coming

gabemontero:
and forgot to pre-announce

btw ... the image-eco failures are very weird ... I suspect not related to these change, but are appearing somewhat persistent at the moment

gabemontero:
aws vpc quota pain with the last image-eco

gabemontero:
ok additional change added ... more simplication and move off of post change reason field analysis

@adambkaplan ptal and repost lgtm or add comments as you see fit

gabemontero:
aws vpc quota pain on e2e-image-eco and operator 

/retest

gabemontero:
`Unauthorized` on latest image_eco

/test e2e-aws-image-ecosystem

gabemontero:
aws vpc quota pain on e2e-aws

/test e2e-aws

gabemontero:
`fail [github.com/openshift/origin/test/extended/util/cli.go:730]: Apr 23 20:35:05.917: Unauthorized`

/test e2e-aws-image-ecosystem

gabemontero:
we have passing tests @adambkaplan for the changes to assist optimized landing on Progressing==False resulting from the recent upgrade test results 

quick reminder there was one small addition since the last time you looked

gabemontero:
@adambkaplan the former form with the reason field checks were faulty when events arrived in a concurrent fashion, and we failed to actually update the status object, setting Progressing to false 

The logs from the failed run showed this because I saw the log msg from `processImportStatus` saying we are all done, but we did not see `CRDUPDATE updating progress/error condition...`

This change should prevent that concurrency issue, as the passing of the boolean resulting as a result of setting that boolean when we log the "all done" message should make things clear.

gabemontero:
@adambkaplan  https://github.com/openshift/cluster-samples-operator/pull/137#issuecomment-485563984 details the mismatch that occurred with processing events and trying to do string compares of the reason field to dictate when the CRD update that would set progressing == false would occur.  Essentially we were worried about the presence of an imagestream in the reason field when that stream was removed previously.

Since the decision to update really comes from the processing in `processImportStatus` simply have that method return a boolean indicating whether a change requiring an update has occurred.

adambkaplan:
@gabemontero I think my confusion stems from the fact there are _two_ custom resources that get updated:
1. The operator config object - `configs.samples.operator.openshift.io/cluster`
2. The `ClusterOperator` instance - `oc get clusteroperator/openshift-samples`

I think in a separate PR we should clean up the code/logs a bit so it's clear when we are updating the operator config object vs. the `ClusterOperator`

PR #140, opened 2019-04-26
reduce gets on samples operator config object when processing template/imagestream events

gabemontero:
/hold

### DIFF
--- a/pkg/operator/controller.go
+++ b/pkg/operator/controller.go
@@ -90,11 +90,6 @@ func NewController() (*Controller, error) {
 		listers:        listers,
 	}
 
-	c.handlerStub, err = stub.NewSamplesOperatorHandler(kubeconfig)
-	if err != nil {
-		return nil, err
-	}
-
 	// Initial event to bootstrap CR if it doesn't exist.
 	c.crWorkqueue.AddRateLimited(sampopapi.ConfigName)
 
@@ -146,6 +141,13 @@ func NewController() (*Controller, error) {
 	c.crInformer = c.sampopInformerFactory.Samples().V1().Configs().Informer()
 	c.crInformer.AddEventHandler(c.crInformerEventHandler())
 	c.listers.Config = c.sampopInformerFactory.Samples().V1().Configs().Lister()
+
+	crStore := c.crInformer.GetStore()
+
+	c.handlerStub, err = stub.NewSamplesOperatorHandler(kubeconfig, crStore)
+	if err != nil {
+		return nil, err
+	}
 
 	return c, nil
 }
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -20,6 +20,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	restclient "k8s.io/client-go/rest"
+	controllercache "k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/retry"
 
 	imagev1 "github.com/openshift/api/image/v1"
@@ -45,7 +46,7 @@ const (
 	skippedtempskey      = "keyForSkippedTemplatesField"
 )
 
-func NewSamplesOperatorHandler(kubeconfig *restclient.Config) (*Handler, error) {
+func NewSamplesOperatorHandler(kubeconfig *restclient.Config, cachestore controllercache.Store) (*Handler, error) {
 	h := &Handler{}
 
 	h.initter = &defaultInClusterInitter{}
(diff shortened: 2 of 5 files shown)


gabemontero:
based on debug analysis ... my first "non-trivial" idea is deferring the cfg retrieval until after the cache processing and determination of the version.

gabemontero:
/assign @adambkaplan 

I'm going to leave the debug commit in a bit longer to help verify GET reduction, but will remove if/when we are happy to merge these 

I can squash the two BUG 1703233 commits as desired if we go with both the imagestream and template change

PTAL

gabemontero:
@bparees FYI for historical perspective in case you were unaware of Clayton's ask to reduce the amount of GETs on the samples operator config CRD object

bparees:
> @bparees  FYI for historical perspective in case you were unaware of Clayton's  ask to reduce the amount of GETs on the samples operator config CRD  object

why do we have to get it at all?  shouldn't it be in the listwatch cache since we're already presumably watching that object?

gabemontero:
interesting @bparees  .... how we get the config obj never came up during dev / code review .. been that way the whole time

Based on what you said just now, I did look at the informer and after a few hops found `Samples().V1().Configs().Informer().Lister().Get(string)` 

Looks like what you are referring to ... if you think I am on the path you are referring to, I'll create a new PR that tries that and see what is up.

If it works we can reconsider these changes accordingly.

thanks

bparees:
> Based on what you said just now, I did look at the informer and after a few hops found Samples().V1().Configs().Informer().Lister().Get(string)
Looks like what you are referring to ... if you think I am on the path you are referring to

yeah that's what i was referring to.  the only concern i'd have is the cache can be stale, ie if you're doing an update on that object from your operator, and then you immediately get the value out of the cache, it will not necessarily reflect the update you just did.  So depending on what requirements you have on the state you're trying to track, it could be a problem there.  But otherwise, in general we should get stuff out of the listwatcher when we can to avoid api calls.

gabemontero:
On Mon, Apr 29, 2019 at 4:51 PM Ben Parees <notifications@github.com> wrote:

> Based on what you said just now, I did look at the informer and after a
> few hops found Samples().V1().Configs().Informer().Lister().Get(string)
> Looks like what you are referring to ... if you think I am on the path you
> are referring to
>
> yeah that's what i was referring to. the only concern i'd have is the
> cache can be stale, ie if you're doing an update on that object from your
> operator, and then you immediately get the value out of the cache, it will
> not necessarily reflect the update you just did. So depending on what
> requirements you have on the state you're trying to track, it could be a
> problem there. But otherwise, in general we should get stuff out of the
> listwatcher when we can to avoid api calls.
>
OK thanks for the confirm.  And yeah, there are cases were we just read,
and there are cases where we read to get the latest state possible before
updating.

Short term, I think it makes sense to still review these changes in the
interim independent of leveraging the Lister.Get directly.  But I'll start
looking into it and report back as I make progress.

pweil-:
`aws_internet_gateway.igw: error attaching EC2 Internet Gateway (igw-0972061aa7f59914b): timeout while waiting for state to become 'success' (timeout: 2m0s)`

/retest

gabemontero:
ok @adambkaplan @bparees I've pushed a commit that replaces the client based GET on the operator's config CRD with the shared informer cache store.

So far in my testing, I have not hit any issues resulting from a stale cache.  I have a placeholder in this commit to allow for choosing between the two form of GETs, but am not using it yet.

Based on test results here and our general consensus, I'll tweak this commit to consolidate or leverage that choice accordingly.

PTAL / thanks

gabemontero:
aws vpc quota pain in e2e-aws e2e-aws-operator

gabemontero:
I'm running PR https://github.com/openshift/cluster-samples-operator/pull/141 in parallel for comparison

Might be seeing some oddities in the last couple of runs here, perhaps coinciding with the informer/shared cache change .... might need to revert that one

gabemontero:
looking at the full pr history as well for comparison

gabemontero:
actually based on the history, I think it is https://github.com/openshift/cluster-samples-operator/pull/140/commits/de61daf163328733ea72fd1e9e436d24bdd0e072

number of lines of code change, that is the most disruptive

going to do one more test run, but reverting that and keeping the shared informer, etc. change along with the simple template one may be the path

gabemontero:
e2e-aws non-related errors

gabemontero:
yeah I'm reverting https://github.com/openshift/cluster-samples-operator/commit/de61daf163328733ea72fd1e9e436d24bdd0e072

gabemontero:
aws vpc quota pain with e2e-aws-upgrade

/test e2e-aws-upgrade

gabemontero:
ok the `e2e-aws-operator` sample operator pod logs look as they should now during its upgrade e2e test

PR #146, opened 2019-05-09
Bug 1707834: use informer cache instead of api server for GETs on tem…

gabemontero:
…plates and imagestreams

/hold

bz fix targeted for 4.2


gabemontero:
Actually I'm seeing some oddities in the samples operator that could point to crdstore staleness / lack of population and then when we try to create, we get already exists:

```
time="2019-05-09T14:54:51Z" level=info msg="imagestream jboss-amq-62 recreated since delete event"
time="2019-05-09T14:54:51Z" level=info msg="error creating samples: imagestreams.image.openshift.io \"jboss-amq-62\" already exists"
time="2019-05-09T14:54:51Z" level=info msg="CRDUPDATE setting in progress to unknown"
time="2019-05-09T14:54:53Z" level=error msg="unable to sync: retry secret event because in the middle of an sample upsert cycle, requeuing"
time="2019-05-09T14:54:54Z" level=error msg="unable to sync: imagestreams.image.openshift.io \"jboss-amq-62\" already exists, requeuing"
```

where that is preceded after initial creates by a failure on setting progressing to true:

```
time="2019-05-09T14:54:35Z" level=info msg="CRDUPDATE progressing true update"
time="2019-05-09T14:54:35Z" level=error msg="unable to sync: retry secret event because in the middle of an sample upsert cycle, requeuing"
time="2019-05-09T14:54:37Z" level=error msg="unable to sync: retry secret event because in the middle of an sample upsert cycle, requeuing"
time="2019-05-09T14:54:38Z" level=error msg="unable to sync: Operation cannot be fulfilled on configs.samples.operator.openshift.io \"cluster\": the object has been modified; please apply your changes to the latest version and try again, requeuing"
time="2019-05-09T14:54:38Z" level=info msg="management state set to managed"
time="2019-05-09T14:54:38Z" level=info msg="Spec is valid because this operator has not processed a config yet"
time="2019-05-09T14:54:38Z" level=info msg="processing subdir 3scale from dir /opt/openshift/operator/ocp-x86_64"
```

Might need to start considering doing the creates through the store so the cache gets seeded?

Might need to call store.Resync() on initial startup?

pkg/stub/handler.go
```diff
@@ -58,18 +61,18 @@ func NewSamplesOperatorHandler(kubeconfig *restclient.Config, cachestore control
...
+	h.crdstore = configObjCacheStore
 
 	h.Fileimagegetter = &DefaultImageStreamFromFileGetter{}
 	h.Filetemplategetter = &DefaultTemplateFromFileGetter{}
 	h.Filefinder = &DefaultResourceFileLister{}
 
-	h.imageclientwrapper = &defaultImageStreamClientWrapper{h: h}
-	h.templateclientwrapper = &defaultTemplateClientWrapper{h: h}
+	h.imageclientwrapper = &defaultImageStreamClientWrapper{h: h, store: streamCacheStore}
+	h.templateclientwrapper = &defaultTemplateClientWrapper{h: h, store: streamCacheStore}
```
gabemontero: this is also a problem ... gonna address this first, see where we land afterward

gabemontero:
ok all green e2e's 

let's try again

pkg/stub/handler.go
```diff
@@ -887,12 +932,23 @@ func (h *Handler) createSamples(cfg *v1.Config, updateIfPresent, registryChanged
...
+			if kerrors.IsAlreadyExists(err) {
+				continue
+			}
 			return err
 		}
 	}
 	return nil
 }
 
+func (h *Handler) refetchCfgMinimizeConflicts(cfg *v1.Config) *v1.Config {
```
adambkaplan: I'd add a godoc comment here, since errors here are swallowed.

pkg/stub/handler.go
```diff
@@ -848,6 +890,9 @@ func (h *Handler) createSamples(cfg *v1.Config, updateIfPresent, registryChanged
 			if updateIfPresent {
 				cache.RemoveUpsert(imagestream.Name)
 			}
+			if kerrors.IsAlreadyExists(err) {
+				continue
+			}
```
adambkaplan: Add comment (swallowed error)

pkg/stub/handler.go
```diff
@@ -887,12 +932,23 @@ func (h *Handler) createSamples(cfg *v1.Config, updateIfPresent, registryChanged
 
 		err = h.upsertTemplate(template, t, cfg)
 		if err != nil {
+			if kerrors.IsAlreadyExists(err) {
+				continue
+			}
```
adambkaplan: Add comment (swallowed error)

pkg/stub/handler.go

pkg/stub/handler.go
gabemontero: actually now that I've sorted out which informer api to use, and am no longer seeing odd timing windows on deletes/recreates with stale caches, we should not need this anymore ... reverting this addition instead.

pkg/stub/handler.go
gabemontero: actually now that I've sorted out which informer api to use, and am no longer seeing odd timing windows on deletes/recreates with stale caches, we should not need this anymore ... reverting this addition instead.

gabemontero:
comments addressed @adambkaplan ... thanks / ptal

gabemontero:
still planning on adding a commit to add the e2e test (hopefully later today)

gabemontero:
Yeah I'm punting on adding a prometheus query to this repo's e2e.  The kubeconfig's bearertoken/username/password fields are all empty (which is different in ginkgo tests in openshift/origin where we have access to those).  I can't figure out how to get a non-403 response when copying what we did in openshift/origin.

@adambkaplan ptal at the updates based on your comments, and if you are still good, please apply the lgtm

thanks

PR #151, opened 2019-06-07
Removed versioning

gabemontero:
/assign @adambkaplan 

@openshift/sig-developer-experience PTAL

### DIFF
--- a/test/e2e/cluster_samples_operator_test.go
+++ b/test/e2e/cluster_samples_operator_test.go
@@ -747,6 +747,8 @@ func TestSpecManagementStateField(t *testing.T) {
 	verifyImageStreamsGone(t)
 	verifyTemplatesGone(t)
 
+	verifyClusterOperatorConditionsComplete(t)
+
 	err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
 		cfg = verifyOperatorUp(t)
 		cfg.Spec.ManagementState = operatorsv1api.Managed
@@ -810,6 +812,8 @@ func TestSpecManagementStateField(t *testing.T) {
 
 	verifyDeletedImageStreamNotRecreated(t)
 	verifyDeletedTemplatesNotRecreated(t)
+	verifyClusterOperatorConditionsComplete(t)
+
 	// get timestamp to check against in progress condition
 	now = kapis.Now()
 	err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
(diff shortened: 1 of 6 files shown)


test/e2e/cluster_samples_operator_test.go
```diff
@@ -1010,13 +1018,15 @@ func TestUpgrade(t *testing.T) {
...
-		cfg := verifyOperatorUp(t)
-		if cfg.ConditionTrue(samplesapi.MigrationInProgress) {
-			return true, nil
-		}
-		return false, nil
-	})
+	if cfg.Status.ManagementState == operatorsv1api.Managed {
+		err = wait.PollImmediate(1*time.Second, 1*time.Minute, func() (bool, error) {
+			cfg := verifyOperatorUp(t)
+			if cfg.ConditionTrue(samplesapi.MigrationInProgress) {
```
adambkaplan: Potential race condition here if the migration takes less than 1 second?

gabemontero:
@adambkaplan the Unmanaged -> Managed getting back to "level" you are asking about is already covered in one of the existing tests:  https://github.com/openshift/cluster-samples-operator/blob/master/test/e2e/cluster_samples_operator_test.go#L785-L869

That flow is going to be the same regardless of the version also changing once we go from Unmanaged to Managed.

adambkaplan:
@gabemontero if we go from `Unmanaged` to `Managed`, do we always see the `ImageChangesInProgress` condition get updated?

https://github.com/openshift/cluster-samples-operator/blob/master/test/e2e/cluster_samples_operator_test.go#L850-L852

test/e2e/cluster_samples_operator_test.go
gabemontero: Good question.  After reviewing the code paths, I think we can confidently say the potential for a race condition is nil.  Here is my thinking ... between the update of the env above and this check, the following needs to occur, in the following order:
- the existing pod for the modified deployment needs to be terminated
- the new pod/container needs to be created / started
- version mismatch detected, MigrationInProgress is set to true
- imagestream/templates content read from disk, samples updated, image changes in progress set to true
- imagesream imports have to complete
- image changes in progress set to false 
- migration in progress set to false

I've never seen the old pod terminate / new pod created / new pod started take less than a second, much less that whole list.

Agree?

gabemontero:
yes @adambkaplan we always update the samples when we "change into" the managed state, and that results in ImageChangesInProgress getting updated.

If we were at Removed, the samples are of course deleted.
If we were at Unmanaged, we can't be sure what the admin did to the samples.  Rather than inspect each one for changes and only update the altered ones, we just redo them.

PR #155, opened 2019-06-25
resurface import errors into cluster operator condidtions (phase one,…

gabemontero:
… prior to introducing new condition)

Some tweaks that resurface import errors into cluster operator conditions based on the meeting between @bparees @smarterclayton and myself

These changes are a prelim to the acceptance of the new condition requested by @smarterclayton via https://github.com/openshift/api/pull/355

### DIFF
--- a/pkg/apis/samples/v1/types.go
+++ b/pkg/apis/samples/v1/types.go
@@ -328,10 +328,12 @@ func (s *Config) ClearNameInReason(reason, name string) string {
 }
 
 const (
-	noInstallDetailed = "Samples installation in error at %s: %s"
-	installed         = "Samples installation successful at %s"
-	moving            = "Samples processing to %s"
-	removing          = "Deleting samples at %s"
+	noInstallDetailed  = "Samples installation in error at %s: %s"
+	installed          = "Samples installation successful at %s"
+	moving             = "Samples processing to %s"
+	removing           = "Deleting samples at %s"
+	doneImportsFailed  = "Samples installed at %s, with image import failures, see samples operator config object"
+	failedImageImports = "FailedImageImports"
 )
 
 // ClusterOperatorStatusAvailableCondition return values are as follows:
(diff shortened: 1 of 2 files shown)


gabemontero:
/assign @bparees 

pkg/apis/samples/v1/types.go
```diff
@@ -328,10 +328,12 @@ func (s *Config) ClearNameInReason(reason, name string) string {
...
-	noInstallDetailed = "Samples installation in error at %s: %s"
-	installed         = "Samples installation successful at %s"
-	moving            = "Samples processing to %s"
-	removing          = "Deleting samples at %s"
+	noInstallDetailed  = "Samples installation in error at %s: %s"
+	installed          = "Samples installation successful at %s"
+	moving             = "Samples processing to %s"
+	removing           = "Deleting samples at %s"
+	doneImportsFailed  = "Samples installed at %s, with image import failures, see samples operator config object"
+	failedImageImports = "FailedImageImports"
```
bparees: "Progressing=false, reason=FailedImageImports" still strikes me as a slightly weird thing to report to a user, but if @smarterclayton says that's the preferred to to bubble up the information, i guess this accomplishes that.

gabemontero: yeah @bparees my recollection from out discussion was that the reason field on progressing is what was highlighted in the console

Though I have not been able to find that exactly, so maybe I'm off here

gabemontero: bump @bparees @smarterclayton 

please confirm if this form of reporting was in fact one of the to-do's for failed imagestream import notification that came out of our disconnected discussions a couple of weeks ago.

bparees: i think the immediate to do that came out of the meeting was to report this via a new condition on the clusteroperator resource (ie the AttentionRequired condition), but I see no reason not to also have this condition on the samples operator resource, and then use it to drive the update of the clusteroperator resource once we agree on the AttentionRequired condition.

tldr this seems fine.

PR #160, opened 2019-07-30
Bug 1717124: add clusterrole to allow cluster-reader access to the samples config obj

gabemontero:
@openshift/openshift-team-developer-experience fyi ... I think image-registry-operator and openshift-controller-manager-operator could use this as well for the cfg objs ...thoughts?

/assign @adambkaplan 
/assign @enj 

@enj - if you have the cycles, PTAL as well, but conversely I can unassign you if you like

### DIFF
--- a/manifests/03-rbac.yaml
+++ b/manifests/03-rbac.yaml
@@ -50,6 +50,25 @@ rules:
 
 ---
 
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-cluster-reader: "true"
+  name: system:openshift:cluster-samples-operator:cluster-reader
+rules:
+- apiGroups:
+  - samples.operator.openshift.io
+  resources:
+  - configs
+  - configs/status
+  verbs:
+  - get
+  - list
+  - watch
+
+---
+
 kind: RoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:


manifests/03-rbac.yaml
```diff
@@ -50,6 +50,26 @@ rules:
...
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-cluster-reader: "true"
+  name: system:openshift:cluster-samples-operator:cluster-reader
+rules:
+- apiGroups:
+  - samples.operator.openshift.io
+  resources:
+  - configs
+  - configs/status
+- apiGroups:
```
enj: Typo?

gabemontero: yes it is ... thx

gabemontero:
yeah @enj I did not see any e2e's in https://github.com/openshift/cluster-config-operator

you happen to have a pointer to any existing client-go based e2e's  that creates a new user, presumably using https://github.com/openshift/client-go/tree/master/user/clientset/versioned/typed/user/v1, assigns the role to them, presumably using https://github.com/openshift/client-go/tree/master/authorization/clientset/versioned/typed/authorization/v1,  and then uses that user to perform other api access ?

manifests/03-rbac.yaml
gabemontero: update pushed

gabemontero:
looking at all the code involved for doing user create / impersonation testing in our https://github.com/openshift/origin/blob/master/test/extended suite, I'm not taking porting all that to client-go/e2e unless something is already available

enj:
> looking at all the code involved for doing user create / impersonation testing in our https://github.com/openshift/origin/blob/master/test/extended suite, I'm not taking porting all that to client-go/e2e unless something is already available

I was more thinking of e2e in origin to asserts that cluster-reader has access to the expected resources.  Probably a problem for another day.

---

YAML looks fine now.

gabemontero:
On Tue, Jul 30, 2019 at 6:19 PM Mo Khan <notifications@github.com> wrote:

> looking at all the code involved for doing user create / impersonation
> testing in our
> https://github.com/openshift/origin/blob/master/test/extended suite, I'm
> not taking porting all that to client-go/e2e unless something is already
> available
>
> I was more thinking of e2e in origin to asserts that cluster-reader has
> access to the expected resources. Probably a problem for another day.
>
yeah that would be more doable.

Post merge

------------------------------
>
> YAML looks fine now.
>
cool thanks

PR #162, opened 2019-08-02
Bug 1734627: preserve deleted EOL ISTs on upgrade

gabemontero:
@openshift/openshift-team-developer-experience @bparees FYI

/assign @adambkaplan

### DIFF
--- a/pkg/stub/imagestreams.go
+++ b/pkg/stub/imagestreams.go
@@ -208,6 +208,27 @@ func (h *Handler) upsertImageStream(imagestreamInOperatorImage, imagestreamInClu
 		return nil
 	}
 
+	// with the advent of removing EOL images in 4.2, we cannot just update the imagestream
+	// with client-go/Update as that will replace vs. patch; we need to preserve EOL specs
+	// if the imagestream was created during 4.1 as to allow pullthrough on builds for example
+	// so we manually merge the tags in the existing image stream that do not exist in the one
+	// in the payload
+	tagsToAdd := []imagev1.TagReference{}
+	for _, existingTag := range imagestreamInCluster.Spec.Tags {
+		found := false
+		for _, newTag := range imagestreamInOperatorImage.Spec.Tags {
+			if newTag.Name == existingTag.Name {
+				found = true
+				break
+			}
+		}
+		if !found {
+			tagsToAdd = append(tagsToAdd, existingTag)
+		}
+	}
+	for _, tag := range tagsToAdd {
+		imagestreamInOperatorImage.Spec.Tags = append(imagestreamInOperatorImage.Spec.Tags, tag)
+	}
 	imagestreamInOperatorImage.ResourceVersion = imagestreamInCluster.ResourceVersion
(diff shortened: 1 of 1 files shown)


gabemontero:
I could not contrive a way to test this in this repo's unit/e2e tests.  Even in our mimicking of an upgrade, there is only 1 set of payload to work with.

pkg/stub/imagestreams.go
```diff
@@ -208,8 +208,18 @@ func (h *Handler) upsertImageStream(imagestreamInOperatorImage, imagestreamInClu
...
+	// if the imagestream was created during 4.1 as to allow pullthrough on builds for example
+	// so we manually merge the tags from the payload onto the existing image
+	for _, newTag := range imagestreamInOperatorImage.Spec.Tags {
+		for i, oldTag := range imagestreamInCluster.Spec.Tags {
+			if newTag.Name == oldTag.Name {
+				imagestreamInCluster.Spec.Tags[i] = newTag
+			}
+		}
+	}
+	_, err := h.imageclientwrapper.Update(imagestreamInCluster)
```
bparees: it looks like this means you're only updating the spec.tags in the imagestream.  There could be top level imagestream fields/metadata that also changed, right?

gabemontero: good point ... will update

adambkaplan:
> I could not contrive a way to test this in this repo's unit/e2e tests. Even in our mimicking of an upgrade, there is only 1 set of payload to work with.

@bparees I thought we had a 4.1 -> 4.2 upgrade job, is that only run when creating a release payload?

bparees:
> @bparees I thought we had a 4.1 -> 4.2 upgrade job, is that only run when creating a release payload?

we do. but since it doesn't create any buildconfigs while running 4.1, it wouldn't expose this gap.

but i guess you could add a test that confirms that the imagestreamtags you expect to be there, are there (including the EOL ones).  but that test would have to be conditionalized on whether it's running in a "upgraded from 4.1" cluster or a "installed from 4.2" cluster.

pkg/stub/imagestreams.go
gabemontero: there is also a problem with this logic in while EOL ones are getting preserved, new ISTs are not getting added

gabemontero: so as part of the fix I've for adding new tags I've switch back to doing the Update with the imagestream from the operator payload, but where that imagestream has been updated to include missing tags from the older version.

bparees: that sounds like a sensible approach...  essentially making the imagestream update purely additive from a tag perspective.

gabemontero:
update pushed @bparees @adambkaplan PTAL

As an example of some validations:
- the java IS did not have the `11` tag in 4.1, but does after a 4.1 to 4.2 upgrade
- the ruby IS still has the EOL 2.0 and 2.2 tags after the upgrade

bparees:
looks plausible to me

gabemontero:
thanks @bparees .... @adambkaplan agrees to merge

PR #163, opened 2019-08-05
Bug 1735711: lock down timing windows when deletes of the config object come in du…

gabemontero:
…ring samples upsert

@assign @adambkaplan

### DIFF
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -150,13 +150,15 @@ func (h *Handler) prepSamplesWatchEvent(kind, name string, annotations map[strin
 		return nil, "", false, err
 	}
 
+	if cfg.DeletionTimestamp != nil {
+		// we do no return the cfg in this case because we do not want to bother with any progress tracking
+		logrus.Printf("Received watch event %s but not upserting since deletion of the Config is in progress", kind+"/"+name)
+		// note, the imagestream watch cache gets cleared once the deletion/finalizer processing commences
+		return nil, "", false, nil
+	}
+
 	if cfg.ConditionFalse(v1.ImageChangesInProgress) {
 		// we do no return the cfg in these cases because we do not want to bother with any progress tracking
-
-		if cfg.DeletionTimestamp != nil {
-			logrus.Printf("Received watch event %s but not upserting since deletion of the Config is in progress and image changes are not in progress", kind+"/"+name)
-			return nil, "", false, nil
-		}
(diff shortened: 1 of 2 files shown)


pkg/stub/handler.go
```diff
@@ -854,14 +868,30 @@ func (h *Handler) setSampleManagedLabelToFalse(kind, name string) error {
...
+	cfg = h.refetchCfgMinimizeConflicts(cfg)
+	if cfg.DeletionTimestamp != nil {
+		return true
+	}
+	return false
+}
+
+// rc:  error - any api errors during upserts
+// rc:  bool - if we abort because we detected a delete
+func (h *Handler) createSamples(cfg *v1.Config, updateIfPresent, registryChanged bool, unskippedStreams, unskippedTemplates map[string]bool) (error, bool) {
```
adambkaplan: nit - golang convention is return error last
```suggestion
func (h *Handler) createSamples(cfg *v1.Config, updateIfPresent, registryChanged bool, unskippedStreams, unskippedTemplates map[string]bool) (bool, error) {
```

gabemontero: param order switched @adambkaplan - thanks !!

pkg/stub/handler.go
```diff
@@ -724,7 +728,17 @@ func (h *Handler) Handle(event v1.Event) error {
 				return err
 			}
 
-			err = h.createSamples(cfg, true, registryChanged, unskippedStreams, unskippedTemplates)
+			abortForDelete, err := h.createSamples(cfg, true, registryChanged, unskippedStreams, unskippedTemplates)
+			if abortForDelete {
```
adambkaplan: We do this even if err != nil? Typical golang convention (and expectation) is that `err != nil` is handled first by method callers.

adambkaplan: /lgtm cancel

gabemontero: I would call this an exception to the rule, as the delete is blowing everything away, including whatever error tracking might have happened with any error processing

gabemontero: I can add a comment to that effect, assuming we can agree on this.

gabemontero: added clarifying comment @adambkaplan .... PTAL

gabemontero:
flakes flakes flakes  posted one in particular on #4-dev-triage

/retest

PR #167, opened 2019-08-08
Bug 1739214:  set degraded reason/msg on cluster operator obj when degraded==true

gabemontero:
@openshift/openshift-team-developer-experience @bparees FYI / PTAL

see line 97 for the params defined for `setOperatorStatus`

### DIFF
--- a/pkg/operatorstatus/operatorstatus.go
+++ b/pkg/operatorstatus/operatorstatus.go
@@ -60,12 +60,12 @@ type ClusterOperatorWrapper interface {
 
 func (o *ClusterOperatorHandler) UpdateOperatorStatus(cfg *v1.Config) error {
 	var err error
-	failing, msgForProgressing, msgForFailing := cfg.ClusterOperatorStatusFailingCondition()
+	failing, failingReason, failingDetail := cfg.ClusterOperatorStatusFailingCondition()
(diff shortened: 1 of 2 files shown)


stevekuznetsov:
LOL

bparees:
@gabemontero update the test to also check that the reason is being populated?

gabemontero:
To really catch this bug @bparees we'd need unit tests for operatorstatus.go and use of a fake clusteroperator client ... we don't have that at this time.  A bigger to-do.

That said, sure, I'll update the unit test we do have to verify the messages coming out the the analysis method for the config.

bparees:
> To really catch this bug @bparees we'd need unit tests for operatorstatus.go and use of a fake clusteroperator client ... we don't have that at this time. A bigger to-do.

I thought you already had a test that was ensuring the degraded condition was being set (by mucking w/ the time).  that's the test i was referring to.  Maybe i'm misremembering what that test did?

gabemontero:
I've updated the unit test we do have @bparees .... but the error is not in the `ClusterOperatorStatusFailingCondition` path .... i.e. https://github.com/openshift/cluster-samples-operator/blob/master/pkg/apis/samples/v1/types.go#L385-L427

It is in https://github.com/openshift/cluster-samples-operator/blob/master/pkg/operatorstatus/operatorstatus.go#L64-L71

PR #178, opened 2019-08-29
DEVEXP-377: add metrics and alerts

gabemontero:
@openshift/openshift-team-developer-experience @bparees PTAL

There is some validation on the failed imagestream import alert query I need to do, but otherwise, this is far enough along to solicit feedback

Aside from the 4.x metrics/alerts docs, it borrows from both
a) the work @ricardomaraschini did for cluster-image-registry-operator
b) the work I did in collaboration with Ben and Clayton for builds back in 3.10


gabemontero:
interesting ... unit test passed for me locally, but I'm getting `listen tcp :5000: bind: address already in use` in CI when I run a loop on the metrics test in server_test.go

will look into / push an update sometime today

gabemontero:
ok got unit test happy now

gabemontero:
some aws quota and Unauthorized pain on e2e2-aws and image eco runs

/retest

gabemontero:
hmm .... several prometheus e2e's in e2e-aws failed with the last run (the first to get by install flakes)

coincidence or did my injection of new metrics break something :-) ?

manifests/011-prometheus-rules.yaml
```diff
@@ -0,0 +1,36 @@
...
+    - alert: SampleImageStreamImportsFailing
+      expr: max_over_time(sum(openshift_samples_failed_imagestream_import_total)[2h]) > 0
+      annotations:
+        message: |
+          Sample ImageStream imports are consistenly failing over 2 hours.
+    - alert: SamplesDegraded
+      expr: samples_operator_status_degraded == 1
+      annotations:
+        message: |
+          Samples could not be deployed and are in Degraded status.
```
bparees: is this something other operators are doing? it seems slightly odd to have an alert for something that we already have explicit dashboards for.   but i can sort of see the rationale.

manifests/09-serivcemonitor.yaml
```diff
@@ -0,0 +1,24 @@
+---
```
bparees: typo in this filename.

manifests/06-operator.yaml
```diff
@@ -7,11 +7,11 @@ spec:
   replicas: 1
   selector:
     matchLabels:
-      name: cluster-samples-operator
+      app: samples-operator
   template:
     metadata:
       labels:
-        name: cluster-samples-operator
+        app: samples-operator
```
bparees: i don't *think* this change has migration/upgrade implications (the deployment should just get stomped) but i'm not certain.....this might leave the old pods around when you change the label selector?  have you tested this?  Same thing for downgrade implications.

manifests/06-operator.yaml
```diff
@@ -29,6 +29,10 @@ spec:
         operator: "Exists"
         effect: "NoExecute"
         tolerationSeconds: 120 # Evict pods within 2 mins.
+      volumes:
+      - name: samples-operator-tls
+        secret:
+          secretName: samples-operator-tls
```
bparees: do you have logic to handle restarting your metrics server/reloading the tls key when it is rotated?  i didn't see it on my first pass through this PR.

manifests/010-metricservice.yaml
```diff
@@ -0,0 +1,20 @@
...
+  labels:
+    app: samples-operator
+  name: metrics
+  namespace: openshift-cluster-samples-operator
+spec:
+  ports:
+  - name: https
+    port: 443
+    protocol: TCP
+    targetPort: 8443
```
adambkaplan: Aren't you using a different port for the metrics server?

manifests/011-prometheus-rules.yaml
```diff
@@ -0,0 +1,36 @@
...
+    - alert: SamplesInvalidConfig
+      expr: samples_operator_status_invalidconfig == 1
+      annotations:
+        message: |
+          Samples operator has been given an invalid configuration.
+    - alert: SamplesMissingSecret
+      expr: openshift_samples_status_invalidsecret{reason="missing-secret"} == 1
+      annotations:
+        message: |
+          Samples operator cannot find the samples pull secret in the openshift namespace.
```
adambkaplan: Is "in the openshift namespace" implied? (do we drop it?)

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,183 @@
...
+
+	configv1 "github.com/openshift/cluster-samples-operator/pkg/apis/samples/v1"
+	"github.com/openshift/cluster-samples-operator/pkg/client"
+	sampoplisters "github.com/openshift/cluster-samples-operator/pkg/generated/listers/samples/v1"
+
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	separator       = "_"
```
adambkaplan: nit - "separator" is longer to type than "_"

manifests/011-prometheus-rules.yaml
gabemontero: yeah @ricardomaraschini 's analogous PR for image registry operator had this and I left it in when I copied some of his work .... @ricardomaraschini - did you see this in other operators, or came up with it on your own

similarly, I could be easily swayed either way ... keep/remove

manifests/09-serivcemonitor.yaml
gabemontero: 10-4 will address

manifests/06-operator.yaml
gabemontero: to be honest this was me playing around with getting the prometheus instance in the openshift-monitoring namespace to pick up the metrics here (which I have still not gotten to work yet, though have another round of things to try before officially soliciting help) ... I was mimicking the OCM-O here ... if I can get it to work with `name: cluster-samples-operator` I'll switch back

manifests/010-metricservice.yaml
gabemontero: yeah this stemmed from copying from OCM-O in trying to get it to work, but yeah, updating this to match the port the golang code is doing might be at least part of my problem re: getting openshift-monitor pods to fetch my stats

manifests/011-prometheus-rules.yaml
gabemontero: given the fact that samples operator copies the install pull secret to the openshift namespace, I thought the clarification was a smidge justified .... I'd rather err on not assuming implications

pkg/metrics/metrics.go
gabemontero: LOL .... but I don't have to hit the shift key whey typing `separator` ;-)

bparees: it also helps to ensure we are consistent about which separator character we use, especially if you're trying to construct strings for comparison and such.

manifests/011-prometheus-rules.yaml
bparees: +1

bparees: +1 to keeping it, that is.  i definitely lean heavy on the "be explicit" side.

manifests/06-operator.yaml
gabemontero: I do not .... most likely the image-registry operator already had this and @ricardomaraschini 's PR did not have to add it .... I'll look to see what the existing img reg op is doing and general and adjust

manifests/011-prometheus-rules.yaml
ricardomaraschini: This alert was part of the ticket I was working on, not sure if others Operators implement it. I like the idea of having an alerts when something is not OK. A dashboard requires someone to be looking at it to see the problem.

manifests/06-operator.yaml
ricardomaraschini: As a side note: I remember I also had problems making Prometheus to read metrics, make sure you can see the endpoint on Prometheus interface(there is a link called "Status" on the top of the interface). Other than that, be patient, metrics take time to show up there(5 minutes at least).

manifests/06-operator.yaml
ricardomaraschini: I don't think Image Registry Operator nicely handles this. Good catch, i will take a look.

manifests/011-prometheus-rules.yaml
gabemontero: ok let's leave it in then, unless opinions change or new ones come in ... thanks  gentlemen

manifests/06-operator.yaml
gabemontero: I've switched it back and along with reordering some of the manifest files have made some progress.

And with @ricardomaraschini 's tip, I am at least seeing samples operator show up in the status/targets page, but the endpoint list is empty and shows 0/0 up.

I'll post more details in a new, non code review comment.

manifests/06-operator.yaml
gabemontero: will leave this open to handle rotation

also will note here that I was seeing some failed mounts of the secret when I had "non-ideal" manifest ordering .... have addressed that at least with some file renames

manifests/010-metricservice.yaml
gabemontero: switched this back .... things are not quite working but some progress has been made ... will post the latest status in a non-review comment.

manifests/06-operator.yaml
ricardomaraschini: I also had this problem(0/0). When I asked for help on monitoring channel they told me that it was something with the selector(it is like the ServiceMonitor not being able to select the Service from where to read the metrics). 

The only difference between our implementations on this regard is that you are using `matchLabels` while I am not. I don't think this could be causing the issue.

https://github.com/ricardomaraschini/cluster-image-registry-operator/blob/metrics/manifests/0000_90_cluster-image-registry-operator_02_operator-servicemonitor.yaml#L16

gabemontero: I've removed the `matchLabels` piece and @ricardomaraschini and I are actively debugging

gabemontero:
OK, with an assist from @ricardomaraschini I've got the integration with prometheus working and can execute my metrics from the console (after sorting out a few things myself there was a rogue port/targetPort combo in the ServiceMonitor that he caught)

I have the following to-do's
- basic validation of my metrics (the missing tbr cred needs some work, maybe the image import)
- finalize the alert queries based on that 
- the cert rotation recycle @bparees asked for

manifests/06-operator.yaml
gabemontero: OK I have an initial pass at restarting the metrics server, though I'm somewhat worried about goroutine sprawl .... my initial investigations did not uncover an existing pattern to copy.

gabemontero:
I've removed the WIP notation, as I've verified integration with prometheus, and the alert queries, manually, as well as each of the metrics individually.

I've also included a http server restart on tls secret update.

@openshift/openshift-team-developer-experience @bparees @adambkaplan PTAL

manifests/06-operator.yaml
gabemontero: .... maybe create a separate signal handler / stop channel for the http server vs. what is used for the controller workqueues, etc.....

manifests/010-prometheus-rules.yaml
```diff
@@ -0,0 +1,40 @@
...
+spec:
+  groups:
+  - name: SamplesOperator
+    rules:
+    - alert: SampleImageStreamImportsFailing
+      # Testing showed that
+      # a) openshift_samples_failed_imagestream_import_total[2h] returned for each imagestream the count at each sample time
+      # b) max_over_time(openshift_samples_failed_imagestream_import_total[2h]) returned a max for each imagestream across all the sample times
+      # c) max(max_over_time(openshift_samples_failed_imagestream_import_total[2h])) returned a single element and max value across all the imagestreams
+      expr: max(max_over_time(openshift_samples_failed_imagestream_import_total[2h])) > 0
```
bparees: why the max value over 2 hours instead of just the current value?  do we need to fire an alert if 1 hour ago there were some failing imagestreams, but now there are not?

manifests/010-prometheus-rules.yaml
```diff
@@ -0,0 +1,40 @@
...
+    - alert: SamplesMissingSecret
+      expr: openshift_samples_status_invalidsecret{reason="missing-secret"} == 1
+      annotations:
+        message: |
+          Samples operator cannot find the samples pull secret in the openshift namespace.
+    - alert: SamplesMissingTBRCredential
+      expr: openshift_samples_status_invalidsecret{reason="missing-tbr-credential"} == 1
+      annotations:
+        message: |
+          Samples operator cannot find credentials for 'registry.redhat.io' in the samples pull secret in the openshift namespace.
```
bparees: for a disconnected install there might not be a registry.redhat.io cred in the pullsecret....  i'd be wary of throwing this alert.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,180 @@
...
+		addCountGauge(ch, invalidSecretDesc, missingTBRCredential, float64(1))
+		return
+	}
+	_, hasTBRCreds := dockerConfigJSON.Auths["registry.redhat.io"]
+	if hasTBRCreds {
+		addCountGauge(ch, invalidSecretDesc, missingTBRCredential, float64(0))
+		return
+	}
+	logrus.Printf(".dockerconfigjson is missing an entry for registry.redhat.io")
+	addCountGauge(ch, invalidSecretDesc, missingTBRCredential, float64(1))
```
bparees: per above, i don't think we can confidently report this as an error/alert.

pkg/metrics/server.go
```diff
@@ -0,0 +1,75 @@
...
+	router.Handle("/metrics", promhttp.Handler())
+	srv := &http.Server{
+		Addr:    bindAddr,
+		Handler: router,
+	}
+
+	return srv
+}
+
+// StopServer stops the server; for tls secret rotation
```
bparees: the pattern we've used elsewhere is just to kill the entire operator process/pod and let it restart, when it needs to pick up a new cert.  

this isn't wrong, per se, but it might be more complicated/work than the alternative.  @ricardomaraschini recently implemented the restart on filechange logic for OCM i believe.

pkg/operator/controller.go
```diff
@@ -84,7 +87,8 @@ func NewController() (*Controller, error) {
 		cvowrapper:     operatorstatus.NewClusterOperatorHandler(operatorClient),
 		crWorkqueue:    workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "samplesconfig-changes"),
 		osSecWorkqueue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "openshift-secret-changes"),
-		opSecWorkqueue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "openshift-config-namespace-secret-changes"),
+		ocSecWorkqueue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "openshift-config-namespace-secret-changes"),
+		opSecWorkqueue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "openshift-cluster-samples-operator-secret-changes"),
```
bparees: if i understand this you're watching the TLS secret in etcd so you can restart the metrics server when it changes.

but you don't need to do that..the secret is mounted to the pod, when it changes the file on disk will change.  So you can just watch the on-disk file for changes and restart when it changes.  Again there's a pattern for this that @ricardomaraschini recently implemented.

manifests/010-prometheus-rules.yaml
gabemontero: my intent was if it persisted over two hours

maybe `max_over_time(sum(openshift_samples_failed_imagestream_import_total[2h]))` ?

I'll experiment some more once I have a cluster up and report, but of course chime in during that interval as your bandwidth permites.

bparees: i see..yeah i'm not a promql guru so i'm not sure how to get there, but it sounds like maybe you want the min value?  that way if it ever went to zero during those 2 hours, you do *not* throw an alert....only if the min value is 1 (it was failing the entire time) you throw an alert?

`if min_faling_over_2_hour_period > 0 throw alert`

manifests/010-prometheus-rules.yaml
gabemontero: good point

if disconnected, they should be setting the registry override in the samples config 

I'll augment the setting with testing that conditional first, and then update the message to add a clarification around that.

thoughts @bparees ?

pkg/metrics/metrics.go
gabemontero: see my above response on checking registry override first

manifests/010-prometheus-rules.yaml
bparees: yeah i think that can work.

pkg/metrics/server.go
gabemontero: yeah I was concerned about level of disruption, but not enough that I want to blaze new trails here

I'll look for @ricardomaraschini 's recent change (unfortunately I have no recollection of that PR) and go from there.

pkg/operator/controller.go
gabemontero: yep per above that sounds better ... will look for it

pkg/metrics/server.go
bparees: yeah in theory rotation should not be that frequent (daily at most, more likely monthly or something) and the operator needs to be able to handle getting killed/restart anyway since that could happen for other reasons.

gabemontero: fyi the change was https://github.com/openshift/cluster-image-registry-operator/pull/342/commits/ffbb403ef87ede359e941891a28378f80dfe105e

pkg/operator/controller.go
gabemontero: fyi the change was https://github.com/openshift/cluster-image-registry-operator/pull/342/commits/ffbb403ef87ede359e941891a28378f80dfe105e

adambkaplan:
/hold

1. Waiting for 4.3 to be open
2. Needs review from telemetry folks

/cc @sichvoge

pkg/operator/controller.go
gabemontero: part of next push

pkg/metrics/server.go
gabemontero: part of next push

gabemontero:
hmmm ... my vendoring in of spf13/cobra did not go as clean as I expected  ...

ignore any upcoming compile errors in the PR as I push/pull to a cleaner GOPATH env and then give the all clear

gabemontero:
its actually more complicated than adding spf13/cobra ... something under the prometheus/k8s dependency relationship has changed 

going tabula raza ... will report back in a bit

gabemontero:
ok sorted out compile and vendor and pulling in prometheus and spf13/cobra .. in the end, got to simplify glide.yaml by a bunch

still need to sort out unit test failures so wait on examinations until I do that

gabemontero:
OK unit tests fixed (the metric help messages needed updating).

I'm still working on the alert when import failures have persistently failed for two hours.  But otherwise everything else has been addressed.

@bparees @adambkaplan @ricardomaraschini PTAL when time permits.

manifests/010-prometheus-rules.yaml
gabemontero: Yeah I think replacing `max` with `min` in my existing query does the trick:

`min(min_over_time(openshift_samples_failed_imagestream_import_total[2h]))`

gabemontero: check that ... I'm not so sure .... some could be failing and some could succeed ... if so, the min would be 0 for the ones that are succeeding

bparees: the metric name is `openshift_samples_failed_imagestream_import_total`, is it not the total number of imagestream imports currently failing at a given point in time, across all the imagestreams?

gabemontero: Yeah I'm giving up on this one .... we will still mark as degraded if the imports stay failed for two hours, and we will alert on degraded.... essentially the same thing

An admin could then go into prometheus and query on `openshift_samples_failed_imagestream_import_total` to quickly see which imports failed.

bparees: makes sense to me.

gabemontero: yeah perhaps I need to fix the name ... it is really a true/false for whether imports fail on a per imagestream basis.

simply remove the `_total` suffix @bparees ?

gabemontero: but I thought it better to convey which imagestreams are failing import

gabemontero:
Aside from perhaps renaming `openshift_samples_failed_imagestream_import_total` updates to all the current comments addressed.

manifests/010-prometheus-rules.yaml
bparees: I take it that means there is a label on this metric which takes as its value the name of the failing imagestream?

bparees: i would agree that removing the _total seems appropriate, if so.

gabemontero: correct there is a `name` label

will remove _total

gabemontero: Per https://prometheus.io/docs/practices/naming/#metric-names and suffix rules I am going to change `_total` to `_info`

gabemontero:
OK now all current set of comments addressed

PTAL as time permits

manifests/06-metricsservice.yaml
```diff
@@ -0,0 +1,18 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    service.alpha.openshift.io/serving-cert-secret-name: samples-operator-tls
```
bparees: is there a non-alpha version of this api we can use?

pkg/stub/secrets.go
```diff
@@ -15,6 +15,7 @@ import (
 const (
 	coreosPullSecretNamespace = "openshift-config"
 	coreosPullSecretName      = "pull-secret"
+	metricsSecret             = "samples-operator-tls"
```
bparees: is this being used somewhere?

gabemontero: yeah in the end only the deployment needed references to that after I switched to the library-go watch approach.

I'll remove.

manifests/06-metricsservice.yaml
gabemontero: According to https://github.com/openshift/service-ca-operator there is now `service.beta.openshift.io/serving-cert-secret-name`

So yeah it has at least gone to beta now.

I will update.

gabemontero:
pushed separate commit for @bparees 's 2 comments from yesterday.

bparees:
looks good to me.

Dockerfile
```diff
@@ -5,6 +5,7 @@ RUN make build
 
 FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
 COPY --from=builder /go/src/github.com/openshift/cluster-samples-operator/cluster-samples-operator /usr/bin/
+RUN ln -sf /usr/bin/cluster-samples-operator /usr/bin/cluster-samples-operator-watch
```
adambkaplan: Use hard link - I learned this the hard way with @ricardomaraschini.

Dockerfile.rhel7
```diff
@@ -5,6 +5,7 @@ RUN make build
 
 FROM registry.svc.ci.openshift.org/ocp/4.0:base
 COPY --from=builder /go/src/github.com/openshift/cluster-samples-operator/cluster-samples-operator /usr/bin/
+RUN ln -sf /usr/bin/cluster-samples-operator /usr/bin/cluster-samples-operator-watch
```
adambkaplan: Ditto - use hard link.

Dockerfile
gabemontero: updating momentarily

Dockerfile.rhel7
gabemontero: updating momentarily

gabemontero:
removed `s` from `ln` options @adambkaplan

gabemontero:
install flakes on e2e-aws and e2e-aws-upgrade 

/retest

gabemontero:
bump @sichvoge

adding some new metrics / alerts for the samples operator in OpenShift

gabemontero:
per @bparees 's request for the image registry PR  https://github.com/openshift/cluster-image-registry-operator/pull/368 -  @s-urbaniak could we ask you or someone on your team (or maybe @brancz's team) to give this a review to see if we're doing metrics/alerting in a sane way?

manifests/010-prometheus-rules.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    name: samples-operator-alerts
+  name: samples-operator-alerts
+  namespace: openshift-cluster-samples-operator
+spec:
+  groups:
+  - name: SamplesOperator
```
brancz: None of these alerts have a severity. The three severities that we use in OpenShift are:

* info: this alert never fires on its own, it only provides additional context when other alerts of a grouping are firing
* warning: this alert describes a problem that is non urgent, typically this means it's fine for someone to look at during worktime, and it usually creates a ticket in a ticket system
* critical: this needs immediate attention and would wake someone up at night, usually this means that the cluster is in a broken state

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,188 @@
...
+	info            = "info"
+	statusSubsystem = "status"
+
+	// openshift_samples_failed_imagestream_import_info
+	// since we want to index by imagestream name, where each imagestream is either failing or not failing, this is why
+	// we went with the _info suffix vs. the _total suffix
+	failedImportsQuery = sampleNamespace + separator + "failed" + separator + "imagestream" + separator + "import" + separator + info
+
+	degraded = "degraded" + separator + info
+	// openshift_samples_status_degraded_info
```
brancz: if you already write it out here, why not just always write out the string? fwiw that's the most common practice as it's easy to search for metrics that way

manifests/010-prometheus-rules.yaml
gabemontero: I've added the severity label to each of them; part of next push; thanks !!

pkg/metrics/metrics.go
gabemontero: yeah there is some history behind all this wrt unfamiliarity with naming conventions, with some programming style opinions mixed into it ... even all the way back to when we first introduced metrics into openshift builds in 3.10

I've never felt strongly either way, and at least this morning, the idea of some simplification appeals to me :-)

will update in next push - thanks

gabemontero:
pushed updates in response to @brancz comments

manifests/010-prometheus-rules.yaml
```diff
@@ -0,0 +1,39 @@
...
```
brancz: None of these alerts have a `for` condition, they shouldn't fire on a single sample having the respective value. I'd suggest at least `15m` or higher.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,183 @@
...
+	importsFailedDesc = prometheus.NewDesc(failedImportsQuery,
+		"Indicates by name whether a samples imagestream import has failed or not (1 == failure, 0 == succes).",
+		[]string{"name"},
+		nil)
+	invalidSecretDesc = prometheus.NewDesc(invalidSecretQuery,
+		"Indicates if the pull secret is functional in the openshift namespace.  If not functional, the reason supplied either points to it missing in the openshift namesapce, or if it is missing credentials for registry.redhat.io when they should exist (1 == not functional, 0 == functional).",
+		[]string{"reason"},
+		nil)
+	degradedStat = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: sampleNamespace,
```
brancz: namespace and subsystem are likely to be removed in the future from the prometheus client, I recommend to just use `Name`

manifests/010-prometheus-rules.yaml
gabemontero: just pushed `for` 's for each

pkg/metrics/metrics.go
gabemontero: removed use of namespace/subsystem

gabemontero:
pushed/squashed @brancz 's second round of comments

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,174 @@
...
+	failedImportsQuery = "openshift_samples_failed_imagestream_import_info"
+
+	degradedQuery = "openshift_samples_degraded_info"
+
+	invalidConfigQuery = "openshift_samples_invalidconfig_info"
+
+	invalidSecretQuery = "openshift_samples_invalidsecret_info"
+
+	missingSecret        = "missing-secret"
+	missingTBRCredential = "missing-tbr-credential"
```
bparees: are we being inconsistent in use of underscore vs dash  here?

gabemontero: I thought I saw the use of dash somewhere, but I can't find that now, and I just revisited https://prometheus.io/docs/practices/naming/ and their labels example has underscores as well

I'll change it

gabemontero: dash to underscore conversion pushed @bparees

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,174 @@
...
+	configInvalidStat = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: invalidConfigQuery,
+		Help: "Indicates the configuration for the samples operator has an invalid configuration setting.",
+	})
+
+	sc         = samplesCollector{}
+	registered = false
+
+	// Streams is the list of streams from the install payload that is maintained by the Handler
+	Streams = []string{}
```
adambkaplan: Why a package-global variable?

gabemontero: it is populated in files.go and referenced in handler.go

gabemontero:
prometheus-rules.yaml fixes pushed @adambkaplan thanks

pkg/metrics/metrics.go
gabemontero: I've changed it to `streams` and added some public helper methods

PR #179, opened 2019-08-29
Bug 1746342: clean up available/progressing/degraded, including new msg/reason, wh…

gabemontero:
…en unmanaged/removed

@openshift/openshift-team-developer-experience PTAL based on cycles available

/assign @bparees 
(given the "broader discussion" on proper conventions for clusteroperator conventions when unmanaged/removed)

@bparees assuming everyone is OK with these changes given current assumptions, do we want to merge this before any "broader discussions"

### DIFF
--- a/pkg/apis/samples/v1/types.go
+++ b/pkg/apis/samples/v1/types.go
@@ -329,18 +329,21 @@ func (s *Config) ClearNameInReason(reason, name string) string {
 }
 
 const (
-	noInstallDetailed  = "Samples installation in error at %s: %s"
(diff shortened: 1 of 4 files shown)


bparees:
since this is basically just changing the reasons/messages we return, not the actual values of the conditions, i think it should be non-controversial and can merge now, even before we go have the broader discussion.

gabemontero:
return values ordering comment from @bparees addressed
also
- put in a constant for the new value for the reason field
- s/failing/degraded/*

PR #180, opened 2019-08-29
add list of imagestreams failing import to progressing/degraded detailed message

gabemontero:
in service of extended test flake https://bugzilla.redhat.com/show_bug.cgi?id=1745418

adding the specific imagestream import failures in the detailed message for clusteroperator detailed/progressing conditions is more useful then pointing admins to the config object

also, we can check in our extended tests to see which imagestreams failed when determining whether the test run is safe to continue  ... I'll be submitting a openshift/origin PR that does that and resolves https://bugzilla.redhat.com/show_bug.cgi?id=1745418 after this merges 

/assign @adambkaplan 
@openshift/openshift-team-developer-experience  FYI


gabemontero:
ok down to one commit .... @adambkaplan PTAL / bump

PR #182, opened 2019-09-05
Bug 1749152: close gaps for updating clusteroperator during delete of config object

gabemontero:
/assign @bparees 
/assign @adambkaplan 
@openshift/openshift-team-developer-experience 

among other things, if we were in the middle of imagestream imports when the delete occurred, I could see *SOME* delays though not 5 minutes

but there may have been some missed opportunities to update the condition because the code did not think it was modified

also adding some more proactive updating and some removal of retry on samples events

### DIFF
--- a/pkg/operatorstatus/operatorstatus.go
+++ b/pkg/operatorstatus/operatorstatus.go
@@ -9,6 +9,7 @@ import (
 	metaapi "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"k8s.io/apimachinery/pkg/api/errors"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/util/retry"
 
 	configv1 "github.com/openshift/api/config/v1"
@@ -19,6 +20,7 @@ import (
 
 const (
 	ClusterOperatorName = "openshift-samples"
+	doingDelete         = "DeletionInProgress"
 )
 
 // ClusterOperatorHandler allows for wrappering access to configv1.ClusterOperator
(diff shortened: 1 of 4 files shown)


gabemontero:
I might fiddle with this one a bit more

/hold

pkg/operatorstatus/operatorstatus.go
```diff
@@ -58,16 +59,32 @@ type ClusterOperatorWrapper interface {
 	Create(state *configv1.ClusterOperator) (err error)
 }
 
-func (o *ClusterOperatorHandler) UpdateOperatorStatus(cfg *v1.Config) error {
-	var err error
+func (o *ClusterOperatorHandler) UpdateOperatorStatus(cfg *v1.Config, deletionInProgress bool) error {
+	if deletionInProgress {
+		err := o.setOperatorStatus(configv1.OperatorAvailable, configv1.ConditionTrue, "", cfg.Status.Version, doingDelete)
+		if err != nil {
+			logrus.Warningf("error occurred while attempting to set degraded condition: %s", err.Error())
```
adambkaplan: nit:
```suggestion
			logrus.Warningf("error occurred while attempting to set available condition: %s", err.Error())
```

pkg/operatorstatus/operatorstatus.go
```diff
@@ -89,9 +108,11 @@ func (o *ClusterOperatorHandler) UpdateOperatorStatus(cfg *v1.Config) error {
 		"",
 		reasonForProgressing)
 	if err != nil {
-		return nil
+		// note error but try the other conditions
+		returnError = err
+		logrus.Warningf("error occurred while attempting to set progressing condition: %s", err.Error())
 	}
-	return nil
+	return returnError
```
adambkaplan: Might be nice to roll all of the errors into a single error.

gabemontero: part of next push

pkg/operatorstatus/operatorstatus.go
gabemontero: part of next push

gabemontero:
/hold cancel

ok, I've pushed updates for @adambkaplan asks, plus some additional items I found specific to delete processing, as well as some subsequent improvements to image import tracking I found with today's testing that will improve the operator declaring steady state.

PTAL

gabemontero:
On Mon, Sep 9, 2019 at 3:18 PM Adam Kaplan <notifications@github.com> wrote:

> *@adambkaplan* requested changes on this pull request.
>
> @gabemontero <https://github.com/gabemontero> do we have any means of
> writing unit or e2e tests for these changes?
>

It is very timing specific @adambkaplan .... we in fact already augmented
the e2e's previously to do a 3 deletes in a row to try and mimic
this scenario.

PR #75, opened 2019-01-09
Renames

gabemontero:
See https://bugzilla.redhat.com/show_bug.cgi?id=1664469 for the final details on the renaming decisions @openshift/sig-developer-experience made wrt names for the related samples operator pieces

This PR currently has picked the commit from https://github.com/openshift/cluster-samples-operator/pull/74 to get `make generate` functional.

When that PR merges, I'll rebase so that this PR only has the rename commit

But the rename changes themselves are running in my local test env, including the e2e's passing, and so are reviewable from that sense.

PR #183, opened 2019-09-17
config: Bump from v1alpha1 to v1

wking:
Also update `update-generated.sh`.  Catching up with 51c1b577c1 (#75).

I don't really know what either of the files I'm touching are for, so feel free to close if this makes no sense ;).


gabemontero:
The config/config.yaml is not really used anymore except for maybe some dev scenarios and *could* be deleted

tmp/codegen I believe is obsolete ... I'll run some experiments to confirm and if it makes sense nuke those with a separate PR

thanks for poking at this

PR #187, opened 2019-10-03
DEVEXP-418: fetch arch from GOARCH, allow for z/ppc

gabemontero:
/assign @adambkaplan 
@bparees ptal if bandwidth permits
@openshift/openshift-team-developer-experience fyi/ptal as you are up for

### DIFF
--- a/pkg/operatorstatus/operatorstatus.go
+++ b/pkg/operatorstatus/operatorstatus.go
@@ -22,6 +22,7 @@ import (
 const (
 	ClusterOperatorName = "openshift-samples"
 	doingDelete         = "DeletionInProgress"
+	nonX86              = "NonX86Platform"
 )
 
 // ClusterOperatorHandler allows for wrappering access to configv1.ClusterOperator
@@ -61,23 +62,37 @@ type ClusterOperatorWrapper interface {
 	Create(state *configv1.ClusterOperator) (err error)
 }
 
+// this method ensures that Available==true and Degraded==false, regardless of other conditions; it currently
+// allows Progressing to be explicitly set to true or false based on the scenario the caller is addressing
+func (o *ClusterOperatorHandler) setOperatorStatusWithoutInterrogatingConfig(progressing configv1.ConditionStatus, cfg *v1.Config, reason string) {
+	err := o.setOperatorStatus(configv1.OperatorAvailable, configv1.ConditionTrue, "", cfg.Status.Version, reason)
+	if err != nil {
+		logrus.Warningf("error occurred while attempting to set available condition: %s", err.Error())
+	}
+	err = o.setOperatorStatus(configv1.OperatorProgressing, progressing, "", cfg.Status.Version, reason)
+	if err != nil {
+		logrus.Warningf("error occurred while attempting to set progressing condition: %s", err.Error())
+	}
+	err = o.setOperatorStatus(configv1.OperatorDegraded, configv1.ConditionFalse, "", cfg.Status.Version, reason)
+	if err != nil {
+		logrus.Warningf("error occurred while attempting to set degraded condition: %s", err.Error())
+	}
+
+}
+
 func (o *ClusterOperatorHandler) UpdateOperatorStatus(cfg *v1.Config, deletionInProgress bool) error {
 	if deletionInProgress {
-		err := o.setOperatorStatus(configv1.OperatorAvailable, configv1.ConditionTrue, "", cfg.Status.Version, doingDelete)
-		if err != nil {
-			logrus.Warningf("error occurred while attempting to set available condition: %s", err.Error())
-		}
-		err = o.setOperatorStatus(configv1.OperatorProgressing, configv1.ConditionTrue, "", cfg.Status.Version, doingDelete)
-		if err != nil {
-			logrus.Warningf("error occurred while attempting to set progressing condition: %s", err.Error())
-		}
-		err = o.setOperatorStatus(configv1.OperatorDegraded, configv1.ConditionFalse, "", cfg.Status.Version, doingDelete)
-		if err != nil {
-			logrus.Warningf("error occurred while attempting to set degraded condition: %s", err.Error())
-		}
+		o.setOperatorStatusWithoutInterrogatingConfig(configv1.ConditionTrue, cfg, doingDelete)
 		// will ignore errors in delete path, but we at least log them above
 		return nil
 	}
+	if len(cfg.Spec.Architectures) > 0 && cfg.Spec.Architectures[0] != v1.AMDArchitecture && cfg.Spec.Architectures[0] != v1.X86Architecture {
+		o.setOperatorStatusWithoutInterrogatingConfig(configv1.ConditionFalse, cfg, nonX86)
+		// will ignore errors in non-x86 path, but we at least log them above
+		return nil
+
+	}
+
 	errs := []error{}
 	degraded, degradedReason, degradedDetail := cfg.ClusterOperatorStatusDegradedCondition()
 	err := o.setOperatorStatus(configv1.OperatorDegraded,
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"runtime"
 	"strings"
 	"sync"
 	"time"
@@ -45,6 +46,8 @@ import (
 
 const (
 	x86OCPContentRootDir = "/opt/openshift/operator/ocp-x86_64"
+	ppcOCPContentRootDir = "/opt/openshift/operator/ocp-ppc64le"
+	zOCPContentRootDir   = "/opt/openshift/operator/ocp-s390x"
 	installtypekey       = "keyForInstallTypeField"
 	regkey               = "keyForSamplesRegistryField"
 	skippedstreamskey    = "keyForSkippedImageStreamsField"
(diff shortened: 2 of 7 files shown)


pkg/stub/handler_test.go
```diff
@@ -94,8 +94,9 @@ func TestWithArchDist(t *testing.T) {
 func TestWithArch(t *testing.T) {
 	h, cfg, event := setup()
 	processCred(&h, cfg, t)
+	// without a mimic call this simiulates are current PPC/390 stories of no samples content
```
bparees:
```suggestion
	// without a mimic call this simulates our current PPC/390 stories of no samples content
```

test/e2e/cluster_samples_operator_test.go
```diff
@@ -160,8 +168,22 @@ func verifyConditionsCompleteSamplesAdded(t *testing.T) error {
...
+		switch cfg.Status.Architectures[0] {
+		case samplesapi.AMDArchitecture:
+			fallthrough
+		case samplesapi.X86Architecture:
+			if !cfg.ConditionTrue(samplesapi.SamplesExist) {
+				return false, nil
+			}
+		default:
+			if cfg.ConditionTrue(samplesapi.SamplesExist) {
+				return false, fmt.Errorf("samples should not exist on non-x86 arch %s", string(cfg.Status.Architectures[0]))
```
bparees: this is interesting...  effectively i guess it's true (there will be no samples to exist on non-x86) but how is this condition set?  or in this case, how is it "not" being set?  because there were no samples files to process, we never trip the logic that sets this condition?

And is not setting this condition going to have any downstream effects where we expect to see samples exist set to be true at some point (ie are we going to keep retrying something, or potentially set ourselves to degraded, or some other undesirable side effect of samplesexist remaining "false" forever?)

gabemontero: It starts off as false, and then after the api obj creates run without api error, it gets set to true
the api obj creates with our current approach will not be initiated, and subsequently no check to set `SamplesExists` to true will occur.

The fact that we are management state `Removed` should short circuit any initial tries to create, or any retries to create.  And we already force Available==true, progressing==false, degraded==false when we are in `Removed` state.

bparees: hm, i guess i forgot we are bootstrapping to "removed" on these architectures.

gabemontero: I think we've gone back and forth on it @bparees, and in looking at the enhancements doc, we did settle on not doing it.

But when I started messing with the implementation, it became apparent to me that bootstrapping as Removed saved some churn and unnecessary processing within the operator.

That said, you did previously note if the cluster admin does switch to Managed, we have to make sure we do not try to install bad samples.

In general, the lack of files found in the samples operator image's payload assures we don't create any samples.

But I forgot to short circuit the setting of `SamplesExists` to true in the Managed/non-x86 case.
Doing that now.

I should also update the enhancements doc to clear things up, assuming we agree on my slight shift here.

bparees: i have a general preference for *not* bootstrapping to removed as it makes it easier for us to deliver content to those clusters on upgrades in the future.

if we bootstrap them to removed, we can't change that during an upgrade (we can't tell if they chose it or we did, so we can't change it).

if we bootstrap them to managed w/ no content, we can always add new content to the operator in the future and on upgrade, the operator will install it.

so all in all i'd like to pursue a path where the operator is bootstrapped as "managed" in all cases.

gabemontero: ok switched back to "managed" in all cases, with the appropriate short circuit to prevent install attempts but ensure available==true, progressing==false,degraded==false

pkg/operatorstatus/operatorstatus.go
```diff
@@ -61,23 +62,35 @@ type ClusterOperatorWrapper interface {
 	Create(state *configv1.ClusterOperator) (err error)
 }
 
+func (o *ClusterOperatorHandler) constantConditions(progressing configv1.ConditionStatus, cfg *v1.Config) {
```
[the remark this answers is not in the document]
gabemontero: but it should be, where for nonx86 there should be a different reason .... late night rush job missed things 

I'll update along those lines and then see what you think

pkg/operatorstatus/operatorstatus.go
```diff
@@ -61,23 +62,35 @@ type ClusterOperatorWrapper interface {
...
-		err = o.setOperatorStatus(configv1.OperatorDegraded, configv1.ConditionFalse, "", cfg.Status.Version, doingDelete)
-		if err != nil {
-			logrus.Warningf("error occurred while attempting to set degraded condition: %s", err.Error())
-		}
+		o.constantConditions(configv1.ConditionTrue, cfg)
 		// will ignore errors in delete path, but we at least log them above
 		return nil
 	}
+	if len(cfg.Spec.Architectures) > 0 && cfg.Spec.Architectures[0] != v1.AMDArchitecture && cfg.Spec.Architectures[0] != v1.X86Architecture {
+		o.constantConditions(configv1.ConditionFalse, cfg)
```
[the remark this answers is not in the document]
gabemontero: see above

pkg/stub/handler.go
```diff
@@ -597,6 +619,12 @@ func (h *Handler) Handle(event v1.Event) error {
 			return h.crdwrapper.UpdateStatus(cfg, dbg)
 		}
 
+		//TODO adjust this check as we start getting samples for z or ppc
+		if len(cfg.Spec.Architectures) > 0 && cfg.Spec.Architectures[0] != v1.AMDArchitecture && cfg.Spec.Architectures[0] != v1.X86Architecture {
+			logrus.Printf("samples are not installed on non-x86 architectures currently")
+			return nil
+		}
```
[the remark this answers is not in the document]
gabemontero: Both.

That said, I started working on changes further down below to handle this case.

When that is ready, I'll push that, but leave this here, and we can compare.

pkg/stub/handler.go
```diff
@@ -277,6 +280,25 @@ func IsRetryableAPIError(err error) bool {
...
+	case strings.Contains(runtime.GOARCH, v1.PPCArchitecture):
+		cfg.Spec.Architectures = append(cfg.Spec.Architectures, v1.PPCArchitecture)
+	case strings.Contains(runtime.GOARCH, v1.S390Architecture):
+		cfg.Spec.Architectures = append(cfg.Spec.Architectures, v1.S390Architecture)
+	case strings.Contains(runtime.GOARCH, v1.AMDArchitecture):
+		fallthrough
+	case strings.Contains(runtime.GOARCH, v1.X86Architecture):
+		cfg.Spec.Architectures = append(cfg.Spec.Architectures, v1.X86Architecture)
+	default:
+		logrus.Warningf("unrecognized GOARCH %s so setting management state to removed", runtime.GOARCH)
```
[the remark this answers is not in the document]
gabemontero: that message is incorrect ... a leftover from when the changes here were bootstrapping to removed

I'll fix the message, and then we can reassess its location

gabemontero:
updates pushed in separate commit @bparees

pkg/operatorstatus/operatorstatus.go
```diff
@@ -61,23 +62,35 @@ type ClusterOperatorWrapper interface {
 	Create(state *configv1.ClusterOperator) (err error)
 }
 
+func (o *ClusterOperatorHandler) setOperatorStatusWithoutInterrogatingConfig(progressing configv1.ConditionStatus, cfg *v1.Config, reason string) {
```
bparees: warrants a comment to the effect of "this always sets available=true/degraded=false"

gabemontero:
infra flakes on e2e-aws-operator and e2e-aws-upgrade

pkg/stub/handler.go
```diff
@@ -802,6 +828,25 @@ func (h *Handler) Handle(event v1.Event) error {
 					progressing.Reason = progressing.Reason + isName + " "
 				}
 			}
+			if len(progressing.Reason) > 0 {
+				progressing.LastUpdateTime = now
+				progressing.LastTransitionTime = now
+				logrus.Debugf("Handle changing processing from false to true")
+				progressing.Status = corev1.ConditionTrue
+			} else {
+				logrus.Debugln("there are no imagestreams availalble for import")
```
bparees:
```suggestion
				logrus.Debugln("there are no imagestreams available for import")
```

pkg/stub/handler.go
```diff
@@ -802,6 +828,25 @@ func (h *Handler) Handle(event v1.Event) error {
...
+				// see if there are unskipped templates, to set SamplesExists to true
+				if !cfg.ConditionTrue(v1.SamplesExist) {
+					if len(h.templateFile) > len(h.skippedTemplates) {
```
bparees: this looks questionable?  isn't it possible that "h.skippedTemplates" contains names that aren't actually template names, so h.skippedTemplates could be larger than h.templateFile, even though there are still some template names that are not in the skip list?

gabemontero:
weird flake looking errors on latest image-eco, but the debug there showed the sample imagestreams imported OK

gabemontero:
And they are importing for me in my locally patched cluster as well

gabemontero:
next round of updates pushed in separate commit @bparees thanks

pkg/operatorstatus/operatorstatus.go
gabemontero: comment added

pkg/stub/handler.go
gabemontero: yep we don't validate the names put in the skipped field

switched to cross referencing the list of files against the skipped list

pkg/stub/handler.go
```diff
@@ -597,6 +619,12 @@ func (h *Handler) Handle(event v1.Event) error {
 			return h.crdwrapper.UpdateStatus(cfg, dbg)
 		}
 
+		//TODO adjust this check as we start getting samples for z or ppc
+		if len(cfg.Spec.Architectures) > 0 && cfg.Spec.Architectures[0] != v1.AMDArchitecture && cfg.Spec.Architectures[0] != v1.X86Architecture {
+			logrus.Printf("samples are not installed on non-x86 architectures currently")
```
bparees:
```suggestion
			logrus.Printf("samples are not installed on non-x86 architectures")
```

bparees:
one final nit and lgtm.

gabemontero:
ok @bparees nit processed and pushed ... thanks @bparees !!

commits squashed

will defer on merging for a bit in case @adambkaplan or @openshift/openshift-team-developer-experience will get a chance to look

gabemontero:
per discussion with @adambkaplan marking this ready for merge based on @bparees 's and my agreement per https://github.com/openshift/cluster-samples-operator/pull/187#issuecomment-538539458

gabemontero:
per recent emails @bparees @adambkaplan @pweil- @siamaksade 

/cherrypick release-4.2

PR #188, opened 2019-10-17
Bug 1762979: retry failed imagestream imports more than once

gabemontero:
document in detailed messages last time recovery import attempted
additional automated testing

@openshift/openshift-team-developer-experience @bparees


gabemontero:
a series of apparent flakes, though the ones in the image_eco, though unrelated to this PR, are interesting:

`Pull image still failed due to error: unable to retrieve auth token: invalid username/password`

perhaps the CI cluster's TBR pull secret is stale

did not see those in the prior run

retrying to see how persistent

/retest

gabemontero:
yep flakes went away ... either the creds got refreshed or something else what up

@openshift/openshift-team-developer-experience @bparees PTAL if you have cycles

pkg/stub/interfaces.go
```diff
@@ -33,6 +34,7 @@ type ImageStreamClientWrapper interface {
 	Update(is *imagev1.ImageStream) (*imagev1.ImageStream, error)
 	Delete(name string, opts *metav1.DeleteOptions) error
 	Watch() (watch.Interface, error)
+	ImageStreamImports(namesapce string) imagev1client.ImageStreamImportInterface
```
bparees:
```suggestion
	ImageStreamImports(namespace string) imagev1client.ImageStreamImportInterface
```

pkg/stub/handler_test.go
```diff
@@ -451,6 +453,89 @@ func TestImageStreamEvent(t *testing.T) {
...
+	h.processImageStreamWatchEvent(is, false)
+
+	if !cfg.ConditionTrue(v1.ImportImageErrorsExist) {
+		t.Fatalf("Import Error Condition not true: %#v", cfg)
+	}
+
+	importError := cfg.Condition(v1.ImportImageErrorsExist)
+	importErrorLastUpdateTime := cfg.Condition(v1.ImportImageErrorsExist).LastUpdateTime
+	h.processImageStreamWatchEvent(is, false)
+	if !importError.LastUpdateTime.Equal(&importErrorLastUpdateTime) {
```
bparees: i'm a little unclear on the ptr vs copy behavior of a few of these calls...

is cfg.Condition() giving you back a copy, or a reference?  The code above seems to assume it's a reference, but the code below (line 515/516) seems to treat it as a copy.

bparees: also seems slightly odd to pass a pointer/reference to the equal function when you are trying to compare the values, but i'm assuming that's just how the type's equal signature is written for performance reasons..  true?

pkg/stub/interfaces.go
gabemontero: part of next push

pkg/stub/handler_test.go
gabemontero: On the `LastUpdateTime.Equal` calls, yes, that type's equal signature (from k8s/apimachinery) takes a pointer ref for the param.

on cfg.Condition(), yes, it returns a ptr/reference.  Most likely when we first introduced cfg.ConditionUpdate we were returning a copy of the struct.

I'll look into going one way or the other as a separate commit in this PR.  Do you have a preference @bparees as to which direction I go?

gabemontero: Sorry I was confused earlier.  The condition array in the Config object is *NOT* an array of pointers to conditions.  It is an array of Conditions.  

The `cfg.Condition` method returns a pointer as an optimization.  But you still need `cfg.ConditionUpdate` to change the original array properly.

You OK with things as they stand @bparees ?  Or would you prefer say `cfg.Condition()` returning a copy vs. a pointer?

pkg/stub/handler_test.go
```diff
@@ -451,6 +453,89 @@ func TestImageStreamEvent(t *testing.T) {
...
+	}
+
+	h.processImageStreamWatchEvent(is, false)
+
+	if !cfg.ConditionTrue(v1.ImportImageErrorsExist) {
+		t.Fatalf("Import Error Condition not true: %#v", cfg)
+	}
+
+	importError := cfg.Condition(v1.ImportImageErrorsExist)
+	importErrorLastUpdateTime := cfg.Condition(v1.ImportImageErrorsExist).LastUpdateTime
```
bparees:
```suggestion
	initialImportErrorLastUpdateTime := cfg.Condition(v1.ImportImageErrorsExist).LastUpdateTime
```

would make this clearer i think.

pkg/stub/handler_test.go
bparees: i think this logic is still wrong:

```
	if !importError.LastUpdateTime.Equal(&importErrorLastUpdateTime) {
```

importError is a pointer to a copy of the condition.

When the actual condition gets updated by calling `processImageStreamWatchEvent`, The copy of the condition pointed to by importError is not going to be changed.  So this check will always succeed because importError is never going to be changed by anything.


The reason i say it's a copy is you got importError from this logic:
```
		for _, rc := range s.Status.Conditions {
			if rc.Type == c {
				return &rc
			}
		}
```

So "rc" will get a copy of the each condition as the for loop iterates, and then you return a pointer to rc.

This is why, as you say, you still have to call cfg.ConditionUpdate to update the original array.

pkg/stub/handler_test.go
```diff
@@ -451,6 +453,89 @@ func TestImageStreamEvent(t *testing.T) {
...
+	if !importError.LastUpdateTime.Equal(&importErrorLastUpdateTime) {
+		t.Fatalf("Import Error Condition updated too soon: old update time %s new update time %s", importErrorLastUpdateTime.String(), importError.LastUpdateTime.String())
+	}
+
+	importError.LastUpdateTime.Time = metav1.Now().Add(-15 * time.Minute)
+	cfg.ConditionUpdate(importError)
+
+	h.processImageStreamWatchEvent(is, false)
+
+	if cfg.Condition(v1.ImportImageErrorsExist).LastUpdateTime.Equal(&importErrorLastUpdateTime) {
```
bparees: conversely, this condition could never be false.  You set:

```
importError.LastUpdateTime.Time = metav1.Now().Add(-15 * time.Minute)
cfg.ConditionUpdate(importError)
```

So you've explicitly made the condition's lastupdate time different from the `importErrorLastUpdateTime` value you stored, so they'll never be equal, even if `processImageStreamWatchEvent(is, false)` doesn't update it.

pkg/stub/handler_test.go
gabemontero: part of next push

pkg/stub/handler_test.go
gabemontero: ok yep ... I've added some refetch / saves ... ptal

pkg/stub/handler_test.go
```diff
@@ -451,6 +453,95 @@ func TestImageStreamEvent(t *testing.T) {
...
+		},
+	}
+
+	h.processImageStreamWatchEvent(is, false)
+
+	if !cfg.ConditionTrue(v1.ImportImageErrorsExist) {
+		t.Fatalf("Import Error Condition not true: %#v", cfg)
+	}
+
+	importError := cfg.Condition(v1.ImportImageErrorsExist)
```
bparees:
```suggestion
```

pkg/stub/handler_test.go
```diff
@@ -451,6 +453,95 @@ func TestImageStreamEvent(t *testing.T) {
...
+	initialImportErrorLastUpdateTime := cfg.Condition(v1.ImportImageErrorsExist).LastUpdateTime
+	h.processImageStreamWatchEvent(is, false)
+	// refetch to see if updated
+	importError = cfg.Condition(v1.ImportImageErrorsExist)
```
bparees:
```suggestion
	importError := cfg.Condition(v1.ImportImageErrorsExist)
```

(you never used the original copy above).

gabemontero:
nit updates pushed @bparees thanks

PR #192, opened 2019-10-29
switch from glide to go.mod

gabemontero:


gabemontero:
one passing set of tests ... cool

/test all

gabemontero:
broken out into separate commits @adambkaplan bump ptal

gabemontero:
FYI @adambkaplan I'm moving off of github.com/openshift/kubernetes*  and onto upstream k8s, and k8s 1.16.2 in a subsequent PR ... separating the churn

adambkaplan:
@gabemontero you need a dependency magnet for the CRD code generator (this PR breaks it). Since that dep does not exist in your go code, it doesn't get vendored in.

See https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/dependencymagnet/doc.go

gabemontero:
magnet added @adambkaplan

dependencymagnet/doc.go
```diff
@@ -0,0 +1,10 @@
+// +build tools
+
+// dependencymagnet imports code that is not an explicit go dependency, but is used in things
+// like Makefile targets.
+package dependencymagnet
+
+import (
+	_ "github.com/jteeuwen/go-bindata/go-bindata"
+	_ "github.com/openshift/library-go/alpha-build-machinery"
```
adambkaplan: @gabemontero I think you need another dep magnet for `k8s.io/code-generator`.

See reference in https://github.com/openshift/cluster-samples-operator/blob/master/hack/codegen/update-generated.sh#L7

That said I think we should switch to library-go's alpha-build-machinery in a separate PR.

gabemontero:
tests green again @adambkaplan

PR #201, opened 2019-12-02
Bug 1778613: Java's tag name should contain only version

jhadvig:
All tag names in the ImageStream that point to an 'ImageStreamTag' should contain only the version.

/assign @gabemontero


gabemontero:
/close

sorry @jhadvig @spadgett this is not the preferred way to change this

Instead, what is preferred is to
- change https://github.com/jboss-container-images/openjdk/blob/openjdk18/templates/image-streams.json or which ever branch the jboss team thinks this should go in
- then update https://github.com/openshift/library/blob/master/official.yaml#L3 to pull in the version they pull the first PR in
- I'll then create a samples PR which pulls in the latest openshift/library content

PR #205, opened 2019-12-05
Bug 1779933: set version to start for s390/ppc until we have actual samples

gabemontero:
/assign @bparees 
/assign @adambkaplan

### DIFF
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -588,6 +588,10 @@ func (h *Handler) Handle(event util.Event) error {
 		// Every time we see a change to the Config object, update the ClusterOperator status
 		// based on the current conditions of the Config.
 		cfg = h.refetchCfgMinimizeConflicts(cfg)
+		//TODO remove this setting of version once we start getting samples for z or ppc
+		if util.IsNonX86Arch(cfg) {
+			cfg.Status.Version = h.version
+		}
 		err := h.cvowrapper.UpdateOperatorStatus(cfg, false)
 		if err != nil {
 			logrus.Errorf("error updating cluster operator status: %v", err)
@@ -625,7 +629,7 @@ func (h *Handler) Handle(event util.Event) error {
 		}
 
 		//TODO adjust this check as we start getting samples for z or ppc
-		if len(cfg.Spec.Architectures) > 0 && cfg.Spec.Architectures[0] != v1.AMDArchitecture && cfg.Spec.Architectures[0] != v1.X86Architecture {
+		if util.IsNonX86Arch(cfg) {
 			logrus.Printf("samples are not installed on non-x86 architectures")
 			return nil
 		}
--- a/pkg/stub/handler_test.go
+++ b/pkg/stub/handler_test.go
@@ -3,6 +3,7 @@ package stub
 import (
 	"fmt"
 	"os"
+	"runtime"
 	"strings"
 	"time"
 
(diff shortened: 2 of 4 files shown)


pkg/stub/handler.go
```diff
@@ -588,6 +588,10 @@ func (h *Handler) Handle(event util.Event) error {
 		// Every time we see a change to the Config object, update the ClusterOperator status
 		// based on the current conditions of the Config.
 		cfg = h.refetchCfgMinimizeConflicts(cfg)
+		//TODO remove this setting of version once we start getting samples for z or ppc
+		if len(cfg.Spec.Architectures) > 0 && cfg.Spec.Architectures[0] != v1.AMDArchitecture && cfg.Spec.Architectures[0] != v1.X86Architecture {
```
bparees: i'm a little confused as to why this is specifically checking array index 0?

1) do we support multiple architectures being specified?
2) if we do, what happens if x86 is listed first and Z is listed second?

gabemontero: > i'm a little confused as to why this is specifically checking array index 0?
> 
>     1. do we support multiple architectures being specified?

at the moment, now ... we just have a place holder given this is an array to support it in the future

> 
>     2. if we do, what happens if x86 is listed first and Z is listed second?

we hard code the entry at start up, and the first time through should set all the clusteroperator related bits

gabemontero: adding validation type logic that handle a cluster admin replacing our boostrapping with incorrect choices is doable, but more complicated 

where we compare what they set in the array with what the golang arch env vars report, etc.

pkg/stub/handler.go
```diff
@@ -588,6 +588,10 @@ func (h *Handler) Handle(event util.Event) error {
...
+			cfg.Status.Version = h.version
```
adambkaplan: Shouldn't we _always_ set this, and vary whether or not it is reported based on the `Available` status?

pkg/stub/handler.go
adambkaplan: nit - move this to a shared method `isNonX86Arch(cfg)`

pkg/stub/handler.go
gabemontero: no we shouldn't ... all that was sorted out in 4.0/4.1 initial dev

only report the version once you have achieved it

pkg/stub/handler.go
gabemontero: ok I can do that

gabemontero: shared method added

gabemontero:
> Since we need to backport this change and don't have e2e CI for Z/power, is there an easy way for us to add unit tests to verify we're setting version correctly?

Yes I believe so ... will start on that

gabemontero:
unit test added @adambkaplan

pkg/stub/handler_test.go
```diff
@@ -83,9 +84,17 @@ func TestWithDist(t *testing.T) {
...
+		t.Errorf("arch not set on bootstrap")
+	}
+	testValue := runtime.GOARCH
+	if testValue == v1.AMDArchitecture {
+		testValue = v1.X86Architecture
+	}
+	if cfg.Spec.Architectures[0] != testValue {
+		t.Errorf("arch set to %s instead of %s", cfg.Spec.Architectures[0], runtime.GOARCH)
 	}
+
```
adambkaplan: It doesn't look like `validate` verifies that we've set version correctly. Needs to be updated.

gabemontero: the setting of the version happens after the `setup()` call ... it does not need to be in `validate`

gabemontero: we don't need it everwhere

gabemontero: nevermind ... I'll do it everywhere

gabemontero:
> @gabemontero need to check that version is set if we're reporting `Available=true; Progressing=false, Degraded=false`

we don't test that in the unit tests ... we already do in the e2e's

https://github.com/openshift/cluster-samples-operator/blob/master/test/e2e/cluster_samples_operator_test.go#L219-L242

too big of a change to do in the unit tests

pkg/stub/handler_test.go
gabemontero: unit test updated

pkg/stub/handler_test.go
```diff
@@ -98,6 +107,7 @@ func TestWithArchDist(t *testing.T) {
 		conditions,
 		statuses, t)
 
+
```
wking: nit: why do we keep adding empty lines?

PR #211, opened 2020-01-03
Bug 1784315: properly handle removal of samplesRegistry; clean up importError mess…

gabemontero:
…age/reason processing

### DIFF
--- a/pkg/stub/imagestreams.go
+++ b/pkg/stub/imagestreams.go
@@ -260,10 +260,6 @@ func (h *Handler) upsertImageStream(imagestreamInOperatorImage, imagestreamInClu
 }
 
 func (h *Handler) updateDockerPullSpec(oldies []string, imagestream *imagev1.ImageStream, opcfg *v1.Config) {
-	if len(opcfg.Spec.SamplesRegistry) == 0 {
-		return
-	}
-
 	logrus.Debugf("updateDockerPullSpec stream %s has repo %s", imagestream.Name, imagestream.Spec.DockerImageRepository)
 	// don't mess with deprecated field unless it is actually set with something
 	if len(imagestream.Spec.DockerImageRepository) > 0 &&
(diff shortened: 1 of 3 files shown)


gabemontero:
some php/mysql hoopla in image eco... flake or new bug? ... let's see

/retest

gabemontero:
I now see the php issue ... initially lost in the shuffle:

`Jan  6 15:59:23.738: INFO: At 2020-01-06 15:57:22 +0000 UTC - event for cakephp-mysql-example: {buildconfig-controller } BuildConfigInstantiateFailed: error instantiating Build from BuildConfig e2e-test-s2i-php-6js2r/cakephp-mysql-example (0): Error resolving ImageStreamTag php:7.3 in namespace openshift: unable to find latest tagged image`

More sclorg deprecations and template / imagestream mismatch hoopla ... investigating.

@openshift/openshift-team-developer-experience FYI ^^

gabemontero:
Ok the test uses `https://raw.githubusercontent.com/openshift/cakephp-ex/master/openshift/templates/cakephp-mysql.json` but that now redirects to  https://github.com/sclorg/cakephp-ex/blob/master/openshift/templates/cakephp-mysql.json and that was updated 4 days ago to remove EOL php

We had been holding off of that cause the templates had not been updated.

Will try to pull that in here with separate commit.  If there are still conflicts, will have to see where they are and assess.

@openshift/openshift-team-developer-experience another FYI re getting a clean image-eco run.

gabemontero:
OK I've done a generic samples bump that includes bump PHP ... let's see if all the sclorg ducks are in a row

gabemontero:
ok image-eco is now passing

infra flakes otherwise

/retest

gabemontero:
/assign @adambkaplan 

tests are green

gabemontero:
fyi @adambkaplan this has the necessary PHP EOL fix as well ... PTAL

pkg/stub/imagestreams.go
```diff
@@ -260,10 +260,6 @@ func (h *Handler) upsertImageStream(imagestreamInOperatorImage, imagestreamInClu
 }
 
 func (h *Handler) updateDockerPullSpec(oldies []string, imagestream *imagev1.ImageStream, opcfg *v1.Config) {
-	if len(opcfg.Spec.SamplesRegistry) == 0 {
-		return
-	}
-
```
adambkaplan: @gabemontero so if `SamplesRegistry` is cleared out, will we set the default registry to `registry.redhat.io`?

pkg/stub/imagestreams.go
```diff
@@ -503,38 +520,34 @@ func (h *Handler) processImportStatus(is *imagev1.ImageStream, cfg *v1.Config) (
 		logrus.Debugf("no error/progress checks cause stream %s is skipped", is.Name)
```
adambkaplan: Any reason to keep the large, commented code block?

pkg/stub/imagestreams.go
```diff
@@ -503,38 +520,34 @@ func (h *Handler) processImportStatus(is *imagev1.ImageStream, cfg *v1.Config) (
...
+		replaceOccurs := util.NameInReason(cfg, processing.Reason, is.Name)
+		if replaceOccurs {
+			processing.Reason = util.ClearNameInReason(cfg, processing.Reason, is.Name)
+			logrus.Debugf("processing reason now %s", processing.Reason)
+			if len(strings.TrimSpace(processing.Reason)) == 0 && processing.Status != corev1.ConditionFalse {
+				logrus.Println("The last in progress imagestream has completed (import status check)")
+				processing.Status = corev1.ConditionFalse
+				processing.Reason = ""
 			}
+			processing.LastTransitionTime = now
```
adambkaplan: I think this belongs in the `if` block above or needs a separate conditional - not clear that we're guaranteed to transition status from `True` to `False` here.

pkg/stub/imagestreams.go
gabemontero: set it back (if it had been changed from `registry.redhat.io`) yes @adambkaplan

pkg/stub/imagestreams.go
gabemontero: yeah I'd like for the fact that we have commented it out soak a bit longer @adambkaplan  ... if we have to add some form of it back, having the prior form of it here will expedite things

pkg/stub/imagestreams.go
gabemontero: yep within the if block thanks

gabemontero: update pushed @adambkaplan

gabemontero:
left the php commit in to get passing tests in case this merges before https://github.com/openshift/cluster-samples-operator/pull/213 @adambkaplan

gabemontero:
with the php eol pr merged, redid this PR to only have the new commit ... image-eco should still pass

pkg/stub/imagestreams.go
```diff
@@ -503,38 +520,34 @@ func (h *Handler) processImportStatus(is *imagev1.ImageStream, cfg *v1.Config) (
...
+	logrus.Printf("There are no more image imports in flight for imagestream %s", is.Name)
+	if processing.Status == corev1.ConditionTrue {
+		now := kapis.Now()
+		// remove this imagestream name, including the space separator, from processing
+		logrus.Debugf("current reason %s ", processing.Reason)
+		replaceOccurs := util.NameInReason(cfg, processing.Reason, is.Name)
+		if replaceOccurs {
+			processing.Reason = util.ClearNameInReason(cfg, processing.Reason, is.Name)
+			logrus.Debugf("processing reason now %s", processing.Reason)
+			if len(strings.TrimSpace(processing.Reason)) == 0 && processing.Status != corev1.ConditionFalse {
```
adambkaplan: `processing.Status` is guaranteed to be `ConditionTrue` here. You can drop the condition check (would be interesting to see if the compiler optimizes this away).

gabemontero: updated thanks @adambkaplan

PR #214, opened 2020-01-07
DEVEXP-507: bootstrap as removed for IPv6

gabemontero:
/assign @bparees 
/assign @adambkaplan 

@openshift/openshift-team-developer-experience FYI

### DIFF
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -1,6 +1,7 @@
 package stub
 
 import (
+	"crypto/tls"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -30,17 +31,17 @@ import (
 	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	imagev1client "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
 	imagev1lister "github.com/openshift/client-go/image/listers/image/v1"
+	configv1lister "github.com/openshift/client-go/samples/listers/samples/v1"
 	templatev1client "github.com/openshift/client-go/template/clientset/versioned/typed/template/v1"
 	templatev1lister "github.com/openshift/client-go/template/listers/template/v1"
-	configv1lister "github.com/openshift/client-go/samples/listers/samples/v1"
 
 	operatorsv1api "github.com/openshift/api/operator/v1"
 	v1 "github.com/openshift/api/samples/v1"
 	"github.com/openshift/cluster-samples-operator/pkg/cache"
 	sampopclient "github.com/openshift/cluster-samples-operator/pkg/client"
-	"github.com/openshift/cluster-samples-operator/pkg/util"
 	"github.com/openshift/cluster-samples-operator/pkg/metrics"
 	operatorstatus "github.com/openshift/cluster-samples-operator/pkg/operatorstatus"
+	"github.com/openshift/cluster-samples-operator/pkg/util"
 
 	sampleclientv1 "github.com/openshift/client-go/samples/clientset/versioned/typed/samples/v1"
 )
@@ -145,6 +146,7 @@ type Handler struct {
 	upsertInProgress bool
 	secretRetryCount int8
 	version          string
+	tbrCheckFailed   bool
 }
 
 // prepSamplesWatchEvent decides whether an upsert of the sample should be done, as well as data for either doing the upsert or checking the status of a prior upsert;
@@ -304,6 +306,39 @@ func (h *Handler) updateCfgArch(cfg *v1.Config) *v1.Config {
 	return cfg
 }
 
+func (h *Handler) tbrInaccessible() bool {
+	if h.configclient == nil {
+		// unit test environment
+		return false
+	}
+	err := wait.PollImmediate(5*time.Second, 3*time.Minute, func() (bool, error) {
+		tlsConf := &tls.Config{}
+		conn, err := tls.Dial("tcp", "registry.redhat.io:443", tlsConf)
+		if err != nil {
+			logrus.Infof("test connection to registry.redhat.io failed with %s", err.Error())
+			return false, nil
+		}
+		defer conn.Close()
+		err = conn.Handshake()
+		if err != nil {
+			logrus.Infof("test connection to registry.redhat.io experienced SSL handshake error %s", err.Error())
+			// these can be intermittent as well so we'll retry
+			return false, nil
+		}
+		logrus.Infof("test connection to registry.redhat.io successful")
+		return true, nil
+
+	})
+
+	if err == nil {
+		h.tbrCheckFailed = true
+		return false
+	}
+
+	logrus.Infof("unable to establish HTTPS connection to registry.redhat.io after 3 minutes, bootstrap to Removed")
+	return true
+}
+
 func (h *Handler) CreateDefaultResourceIfNeeded(cfg *v1.Config) (*v1.Config, error) {
 	// assume the caller has call lock on the mutex .. out pattern is to have that as
 	// high up the stack as possible ... loc because need to
@@ -354,7 +389,17 @@ func (h *Handler) CreateDefaultResourceIfNeeded(cfg *v1.Config) (*v1.Config, err
 		cfg.Kind = "Config"
 		cfg.APIVersion = v1.GroupName + "/" + v1.Version
 		cfg = h.updateCfgArch(cfg)
-		cfg.Spec.ManagementState = operatorsv1api.Managed
+		switch {
+		// TODO as we gain content for non x86 platforms we can remove the nonx86 check
+		case util.IsNonX86Arch(cfg):
+			cfg.Spec.ManagementState = operatorsv1api.Removed
+			cfg.Status.Version = h.version
+		case h.tbrInaccessible():
+			cfg.Spec.ManagementState = operatorsv1api.Removed
+			cfg.Status.Version = h.version
+		default:
+			cfg.Spec.ManagementState = operatorsv1api.Managed
+		}
 		h.AddFinalizer(cfg)
 		// we should get a watch event for the default pull secret, but just in case
 		// we miss the watch event, as well as reducing churn with not starting the
@@ -509,11 +554,11 @@ func (h *Handler) Handle(event util.Event) error {
 		// and event delete flag true
 		if event.Deleted {
 			logrus.Info("A previous delete attempt has been successfully completed")
-			h.cvowrapper.UpdateOperatorStatus(cfg, true)
+			h.cvowrapper.UpdateOperatorStatus(cfg, true, h.tbrCheckFailed)
 			return nil
 		}
 		if cfg.DeletionTimestamp != nil {
-			h.cvowrapper.UpdateOperatorStatus(cfg, true)
+			h.cvowrapper.UpdateOperatorStatus(cfg, true, h.tbrCheckFailed)
 			// before we kick off the delete cycle though, we make sure a prior creation
 			// cycle is not still in progress, because we don't want the create adding back
 			// in things we just deleted ... if an upsert is still in progress, return an error;
@@ -588,11 +633,7 @@ func (h *Handler) Handle(event util.Event) error {
 		// Every time we see a change to the Config object, update the ClusterOperator status
 		// based on the current conditions of the Config.
 		cfg = h.refetchCfgMinimizeConflicts(cfg)
-		//TODO remove this setting of version once we start getting samples for z or ppc
-		if util.IsNonX86Arch(cfg) {
-			cfg.Status.Version = h.version
-		}
-		err := h.cvowrapper.UpdateOperatorStatus(cfg, false)
+		err := h.cvowrapper.UpdateOperatorStatus(cfg, false, h.tbrCheckFailed)
 		if err != nil {
 			logrus.Errorf("error updating cluster operator status: %v", err)
 			return err
@@ -626,12 +667,6 @@ func (h *Handler) Handle(event util.Event) error {
 			dbg := "spec corrected"
 			logrus.Printf("CRDUPDATE %s", dbg)
 			return h.crdwrapper.UpdateStatus(cfg, dbg)
-		}
-
-		//TODO adjust this check as we start getting samples for z or ppc
-		if util.IsNonX86Arch(cfg) {
-			logrus.Printf("samples are not installed on non-x86 architectures")
-			return nil
 		}
 
 		h.buildSkipFilters(cfg)
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -57,7 +57,7 @@ func ConditionUnknown(s *samplev1.Config, c samplev1.ConfigConditionType) bool {
 	return false
 }
 
-func AnyConditionUnknown(s *samplev1.Config, ) bool {
+func AnyConditionUnknown(s *samplev1.Config) bool {
 	for _, rc := range s.Status.Conditions {
 		if rc.Status == corev1.ConditionUnknown {
 			return true
@@ -150,7 +150,6 @@ const (
 	// numConfigConditionType is a helper constant that captures the number possible conditions
 	// defined above in this const block
 	numconfigConditionType = 7
-
 )
 
 // ClusterOperatorStatusAvailableCondition return values are as follows:
@@ -295,7 +294,7 @@ func ClusterOperatorStatusProgressingCondition(s *samplev1.Config, degradedState
 
 // ClusterNeedsCreds checks the conditions that drive whether the operator complains about
 // needing credentials to import RHEL content
-func ClusterNeedsCreds(s *samplev1.Config, ) bool {
+func ClusterNeedsCreds(s *samplev1.Config) bool {
 	if strings.TrimSpace(s.Spec.SamplesRegistry) != "" &&
 		strings.TrimSpace(s.Spec.SamplesRegistry) != "registry.redhat.io" {
 		return false
--- a/test/e2e/cluster_samples_operator_test.go
+++ b/test/e2e/cluster_samples_operator_test.go
@@ -3,6 +3,7 @@ package e2e
 import (
 	"fmt"
 	"io/ioutil"
+	"net"
 	"os"
 	"path/filepath"
 	"strings"
@@ -36,7 +37,7 @@ import (
 
 var (
 	kubeConfig     *rest.Config
-	operatorClient *configv1client.ConfigV1Client
+	configClient   *configv1client.ConfigV1Client
 	kubeClient     *kubeset.Clientset
 	imageClient    *imageset.Clientset
 	templateClient *templateset.Clientset
@@ -56,8 +57,8 @@ func setupClients(t *testing.T) {
 			t.Fatalf("%#v", err)
 		}
 	}
-	if operatorClient == nil {
-		operatorClient, err = configv1client.NewForConfig(kubeConfig)
+	if configClient == nil {
+		configClient, err = configv1client.NewForConfig(kubeConfig)
 		if err != nil {
 			t.Fatalf("%#v", err)
 		}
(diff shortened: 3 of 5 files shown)


test/e2e/cluster_samples_operator_test.go
```diff
@@ -592,6 +595,9 @@ func verifyDeletedImageStreamNotRecreated(t *testing.T) {
 	if !verifyX86(t) {
 		return
 	}
+	if util.IPv6Only() {
+		return
+	}
```
adambkaplan: This could be interesting...I suspect that the test runner is on our CI server. No idea how the ipv6 clusters communicate with the rest of the world if they are truly single-stack (rather than dualstack).

Another reason why this info should be exposed to the cluster vs. our hacks.

pkg/stub/handler.go
```diff
@@ -304,6 +305,43 @@ func (h *Handler) updateCfgArch(cfg *v1.Config) *v1.Config {
 	return cfg
 }
 
+//TODO leaving this in for now to promote discussion of this which started in planning and enhancement proprosal;
+// can delete if need be as PR review process completes 
+func (h *Handler) Disconnected() bool {
```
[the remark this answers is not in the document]
gabemontero: I agree @adambkaplan which is why the call to this is commented out.

Between anyone on @openshift/openshift-team-developer-experience and @bparees I was hoping to brainstorm a bit on the specifics of how we would do this.  If not here in the PR, we could use this as a starting point for discussion in scrum or maybe the team arch meeting.  But if folks have immediate thoughts, please do chime in.

Otherwise I can remove it from this PR and add it to a separate one after we are done here with IPv6 to 

Thoughts?

test/e2e/cluster_samples_operator_test.go
gabemontero: Good point.

To date the working doc has `you can look at the network configuration, either in install-config.yaml for things that run at bootstrap time, or in the “cluster” network.config.openshift.io object later on. eg, look at the entries in spec.ServiceNetwork and if they are all IPv6, then you are on a single-stack IPv6 cluster. `

I'll take a look at network object and follow up in forum-ipv6 and then circle back to the PR, at least for the test runner.

I think we are OK with the in-operator stuff.

gabemontero: ok I've pushed the update that leverages the cluster network config to determine ipv4 vs. ipv6 for the framework

to clarify on why we should use the k8s utility within the operator, it checks the pod locally and avoids an RPC call to the api server.

PTAL @adambkaplan

pkg/stub/handler.go
```diff
@@ -304,6 +305,43 @@ func (h *Handler) updateCfgArch(cfg *v1.Config) *v1.Config {
...
+	}
+	if pods == nil {
+		// path possible with unit tests
+		return false
+	}
+	img := ""
+	for _, pod := range pods.Items {
+		for _, c := range pod.Spec.Containers {
+			img = strings.TrimSpace(c.Image)
+			if strings.HasPrefix(img, "registry.redhat.io") ||
```
bparees: now that i think about it this is probably not a valid check.  the pod image field will still reference registry.redhat.io, it's just that there will be an image content source policy that causes the pull to actually go to the mirror registry.

pkg/stub/handler.go
```diff
@@ -304,6 +305,43 @@ func (h *Handler) updateCfgArch(cfg *v1.Config) *v1.Config {
...
+				strings.HasPrefix(img, "registry.svc.ci.openshift.org") {
+				logrus.Info("Samples operator install payload comes from known, network accessible registry")
+				return false
+			}
+		}
+	}
+	out, err := exec.Command("ping", "registry.redhat.io", "-c 5", "-i 3").Output()
```
bparees: this is "more valid" but i suspect we'll run into network envs where icmp is blocked.  (or even registry.redhat.io disabling icmp for some reason at some point).

bparees: better to attempt an actually https connection

adambkaplan: Fodder for further conversation - I can see a customer asking "why is the Samples operator trying to make a request to registry.redhat.io?"

pkg/stub/handler.go
gabemontero: Yeah I suppose we could list all ImageContentSourcePolicies and inspect them for our two registries here.

Thoughts on that?

gabemontero: Or if there are any ImageContentSourcePolicies assume disconnected perhaps ...

bparees: i don't think you can draw any certain conclusions from the ICSP configuration

pkg/stub/handler.go
gabemontero: true ... assuming a https connection with skip tls verify is OK ?   ...simplifies things somewhat ...

bparees: strictly speaking you need to take any proxy config into account, otherwise you're going to skip samples install in a cluster that requires a proxy.  (and samples do work/import even if the cluster requires a proxy).

skipping tls for this check is probably fine which does simplify the proxy-config-handling logic.

bparees: (if there is not a proxy, you should not need to skiptls though, since the redhat registry should be signed with a CA that's in your system trusts already)

gabemontero: Between this and the ICSP thread above, and given the https://github.com/openshift/enhancements/pull/172#discussion_r363490443 comment regarding a simple "we are disconnected" bit in the cluster, thoughts on the feasibility of simply pushing for this instead @bparees ?

bparees: simply pushing for the explicit config indication?  i'm all for pushing that discussion to a conclusion and then coming back to this if we don't get what we want there.

pkg/util/util.go
```diff
@@ -333,6 +336,20 @@ func IsNonX86Arch(cfg *samplev1.Config) bool {
...
 }
 
+func IPv6Only() bool {
+	ip, err := utilnet.ChooseHostInterface()
+	if err != nil {
+		logrus.Warningf("Problem determining IP version via k8s helpers: %s", err.Error())
+		return false
+	}
+	if ip.To4() != nil {
+		logrus.Info("Samples operator running in an environment with IPv4")
```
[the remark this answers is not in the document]
gabemontero: part of next push

test/e2e/cluster_samples_operator_test.go
```diff
@@ -153,6 +154,46 @@ func verifyX86(t *testing.T) bool {
 	return true
 }
 
+func verifyIPv6(t *testing.T) bool {
+	err := wait.PollImmediate(1*time.Second, 1*time.Minute, func() (bool, error) {
+		networkConfig, err := configClient.Networks().Get("cluster", metav1.GetOptions{})
+		if !stub.IsRetryableAPIError(err) && err != nil {
+			t.Logf("verifyIPv6 got retryable error %s", err.Error())
```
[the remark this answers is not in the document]
gabemontero: In my experience `err.Error()` is easier to read than `%v` and there is not a huge difference in the information relayed.

Sorry I'm inclined to leave as is.

Fixing message to say got unretryable in next push

test/e2e/cluster_samples_operator_test.go
```diff
@@ -153,6 +154,46 @@ func verifyX86(t *testing.T) bool {
...
+			return false, err
+		}
+		if err != nil {
+			t.Logf("verifyIPv6 got unretryable error %s", err.Error())
```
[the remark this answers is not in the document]
gabemontero: fixed in next push

test/e2e/cluster_samples_operator_test.go
```diff
@@ -153,6 +154,46 @@ func verifyX86(t *testing.T) bool {
...
+				}
+				if ip.To4() != nil {
+					t.Logf("verifyIPv6 found ipv4 %s", ip.String())
+					return true, nil
+				}
+				t.Logf("verifyIPv6 IP %s not ipv4", ip.String())
+			}
+		}
+		t.Logf("verifyIpv6 done looping through cluster networks, found no ipv4")
+		return false, nil
```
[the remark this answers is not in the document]
gabemontero: I was concerned about how quickly `networkConfig.Status.ClusterNetwork` might get populated with its entries.  Hence I did not abort the poll as you note.

How about the compromise that if the len of  `networkConfig.Status.ClusterNetwork` is greater than one when I get here I abort immediately as you note.  But if it zero, I keep polling?

gabemontero:
initial response to your 2nd wave of comments pushed @adambkaplan (though I've countered on a coupled of things)

pkg/stub/handler.go
gabemontero: yes that ... do you have suggestions on how we should propagate the discussion past https://github.com/openshift/enhancements/pull/172#discussion_r363490443 ... who/how we should engage @bparees ?

bparees: seems like we should start with @crawford though i feel like this was pushed back on previously.  Still seems to me that it's information that other operators would find useful.

bparees: (the other issue may have been that literally nothing actually knows..there's no explicit "this is a disconnected install" setting/config that gets passed in during installation, so it's not clear how the config info would get set in the first place.  That may have been the blocker last time we discussed this)

gabemontero: OK, I'm going to circle back then to just determining by network connection.  With that, I'm thinking
- I'll check proxy config ... if one exists, I'm going to punt, and say "not disconnected, or connected".  It seems odd to me if a customer requires a proxy when they are also running disconnected.  Or such a combination is obscure enough that I don't want to deal with getting the proxy related certs.

But if proxy and disconnected together is a thing, please enlighten me.

- With that, I'll try to just make the https connection ... any unexpected items with certs, I'll go skip tls

@bparees thoughts?

- re: customer worried about the samples operator accessing `registry.redhat.io` @adambkaplan ... IMO there will be some much activity to that from the image registry, a small set of connection attempts from the samples operator will get lost in the shuffle

bparees: your supposition seems reasonable to me (that if they've configured a proxy, that proxy is enabling external access and thus imagestream import should succeed as long as all our import components are properly consuming the proxy).

gabemontero: ok @bparees I've pushed the new form that ignores the image for the operator pod, and then
1) sees if a proxy is configured 
2) tries for one mintue to get a successful https / tls connection to registry.redhat.io

PTAL

pkg/stub/handler.go
```diff
@@ -304,6 +305,48 @@ func (h *Handler) updateCfgArch(cfg *v1.Config) *v1.Config {
...
+			// these can be intermittent as well so we'll retry
+			return false, nil
+		}
+		logrus.Infof("test connection to registry.redhat.io successful")
+		return true, nil
+
+	})
+
+	if err == nil {
+		return false
```
bparees: should be return true?

pkg/stub/handler.go
```diff
@@ -304,6 +305,48 @@ func (h *Handler) updateCfgArch(cfg *v1.Config) *v1.Config {
...
+	}
+
+	logrus.Infof("unable to establish HTTPS connection to registry.redhat.io after 1 minute, bootstrap to Removed")
+	return false
```
bparees: or this should be true (since the method name is "Disconnected", returning false implies we are *not* disconnected)

gabemontero: ah forgot about the name of the method (was originally just a placeholder)

will address momentarily

gabemontero: method name addressed @bparees PTAL

bparees: you still need to change this to return `true` right?

gabemontero: duh ... yes ... just a moment

gabemontero: ok no returning true thanks @bparees

bparees:
tbr accessibility logic looks ok to me now

gabemontero:
> tbr accessibility logic looks ok to me now

thanks @bparees 

I've updated the associated enhancements proposal https://github.com/openshift/enhancements/pull/172 accordingly as well

@adambkaplan do you want to take a pass at the "bootstrap removed if no TBR flow" in addition to the ipv6 as well?

pkg/stub/handler.go
```diff
@@ -354,7 +395,15 @@ func (h *Handler) CreateDefaultResourceIfNeeded(cfg *v1.Config) (*v1.Config, err
 		cfg.Kind = "Config"
 		cfg.APIVersion = v1.GroupName + "/" + v1.Version
 		cfg = h.updateCfgArch(cfg)
-		cfg.Spec.ManagementState = operatorsv1api.Managed
+		switch {
+		// order of calls matters here ... bootstrap Removed when IPv6 regardless of disconnected checks
+		case util.IPv6Only():
+			cfg.Spec.ManagementState = operatorsv1api.Removed
+		case h.tbrInaccessible():
+			cfg.Spec.ManagementState = operatorsv1api.Removed
```
adambkaplan: So here's an idea - what if we just test for TBR access? The only reason we are bootstrapping `Removed` is that _today_ `registry.redhat.io` does not support IPv6 connections today. This improves the install experience by getting us around a whole host of failure modes for the samples operator - disconnected, crappy proxy, poor network, etc. Admins can always switch the samples operator to be `Managed` day 2.

Bonus - when `registry.redhat.io` does support IPv6, we don't need to make any code changes.

pkg/stub/handler.go
```diff
@@ -589,7 +638,8 @@ func (h *Handler) Handle(event util.Event) error {
 		// based on the current conditions of the Config.
 		cfg = h.refetchCfgMinimizeConflicts(cfg)
 		//TODO remove this setting of version once we start getting samples for z or ppc
-		if util.IsNonX86Arch(cfg) {
+		// and registry.redhat.io starts support IPv6
+		if util.IsNonX86Arch(cfg) || util.IPv6Only() {
```
adambkaplan: If we're removed, do we reach this code path? If so then we don't need to check for IPv6 here.

pkg/stub/handler.go
```diff
@@ -629,7 +679,8 @@ func (h *Handler) Handle(event util.Event) error {
 		}
 
 		//TODO adjust this check as we start getting samples for z or ppc
-		if util.IsNonX86Arch(cfg) {
+		// and registry.redhat.io starts support IPv6
+		if util.IsNonX86Arch(cfg) || util.IPv6Only() {
```
adambkaplan: Similar - if we're removed, do we reach this code path?

pkg/stub/handler.go
gabemontero: I like the idea.  I'll sort through the implications in the other comments below.

pkg/stub/handler.go
gabemontero: bootstrapping as removed has some implications wrt forcing the necessary clusteroperator bits to appease the CVO during initial install, like the version here.

But maybe  we could do that during the initial create and remove this.  I'll look into it.

pkg/stub/handler.go
gabemontero: the code here was more of a paranoid make sure we don't try to install

I think we can remove it at this point.

pkg/stub/handler.go
gabemontero: One fly in the ointment did just occur to me ... do we support proxy in IPv6 ?

Assuming so, I think we need to tweak the tbr check to include ipv6.

gabemontero:
I've pushed the next set up updates based on our exchange today @adambkaplan

gabemontero:
green tests

pkg/stub/handler.go
```diff
@@ -304,6 +305,52 @@ func (h *Handler) updateCfgArch(cfg *v1.Config) *v1.Config {
...
+		logrus.Infof("test connection to registry.redhat.io successful")
+		return true, nil
+
+	})
+
+	if err == nil {
+		return false
+	}
+
+	logrus.Infof("unable to establish HTTPS connection to registry.redhat.io after 1 minute, bootstrap to Removed")
```
dmage: 3 minutes?

pkg/stub/handler.go
adambkaplan: @gabemontero I would assume that we do support proxy in IPv6

pkg/stub/handler.go
```diff
@@ -304,6 +305,52 @@ func (h *Handler) updateCfgArch(cfg *v1.Config) *v1.Config {
...
+		// if not set, unit test environment, so don't bother
+		proxy, err := h.configclient.Proxies().Get("cluster", metav1.GetOptions{})
+		if err == nil &&
+			(len(proxy.Status.HTTPProxy) > 0 || len(proxy.Status.HTTPSProxy) > 0) {
+			// if proxies are involved we assume we are not in a disconnected scenario
+			return true, nil
+		}
+
+		tlsConf := &tls.Config{}
+		conn, err := tls.Dial("tcp", "registry.redhat.io:443", tlsConf)
```
adambkaplan: should we leave `registry.redhat.io` hard-coded?

pkg/stub/handler.go
```diff
@@ -304,6 +305,52 @@ func (h *Handler) updateCfgArch(cfg *v1.Config) *v1.Config {
...
+		return true
+	}
+	err := wait.PollImmediate(5*time.Second, 3*time.Minute, func() (bool, error) {
+		// if not set, unit test environment, so don't bother
+		proxy, err := h.configclient.Proxies().Get("cluster", metav1.GetOptions{})
+		if err == nil &&
+			(len(proxy.Status.HTTPProxy) > 0 || len(proxy.Status.HTTPSProxy) > 0) {
+			// if proxies are involved we assume we are not in a disconnected scenario
+			return true, nil
+		}
```
adambkaplan: So the assumption here is that if the cluster has a global proxy, it should allow connections to registry.redhat.io.

pkg/stub/handler.go
```diff
@@ -304,6 +305,52 @@ func (h *Handler) updateCfgArch(cfg *v1.Config) *v1.Config {
...
+	if h.configclient == nil {
+		// unit test environment
+		return false
+	}
+	// we still check for this in case we have proxies in a ipv6 env;
+	// if proxies will not be supported in ipv6 single stack, we can
+	// remove this
+	if util.IPv6Only() {
+		return true
+	}
```
adambkaplan: Do we need to do this here? Consider the following deployment topology:

1. IPv6 cluster that has no direct connection to the internet.
2. Proxy set up in a dual-stack DMZ (allows IPv4 and IPv6 connections).
3. IPv6 cluster configured to be proxied via the DMZ.

In this case the samples is in an IPv6 cluster, but may nonetheless be able to connect to registry.redhat.io. This may be an edge case, and the work-around is simple if the customer wants samples content.

pkg/stub/handler.go
gabemontero: yep thanks

pkg/stub/handler.go
gabemontero: correct ... it was a conclusion @bparees and I reached in https://github.com/openshift/cluster-samples-operator/pull/214#discussion_r364829755

pkg/stub/handler.go
gabemontero: aside from this, per the earlier discussion, proxy in ipv6 being supported also confirms removal per the code comment above

hostnames still leverage DNS (in addition to static config and SLAAC) in ipv6, so the hostname below should sort out regardless of ipv6 / ipv4

Net ... I'm removing :-)

pkg/stub/handler.go
gabemontero: that said ... if a proxy is in play the TBR connection test below should pass as well ... I'll remove

gabemontero:
latest comments addressed, along with a corresponding adjustment to the clusteroperator update 

commits squashed

@adambkaplan ptal

gabemontero:
and green tests

gabemontero:
> /lgtm
> 
> Given the direction of this PR, we should update the descriptions in the JIRA story to make it clear we're removing samples if `registry.redhat.io` is not accessible for any reason.

Jira updated @adambkaplan thanks

gabemontero:
> @gabemontero: #214 failed to apply on top of branch "release-4.3":
> 
> ```
> error: Failed to merge in the changes.
> Using index info to reconstruct a base tree...
> M	pkg/stub/config.go
> M	pkg/stub/handler.go
> M	test/e2e/cluster_samples_operator_test.go
> Falling back to patching base and 3-way merge...
> Auto-merging test/e2e/cluster_samples_operator_test.go
> Auto-merging pkg/stub/handler.go
> CONFLICT (content): Merge conflict in pkg/stub/handler.go
> Auto-merging pkg/stub/config.go
> Patch failed at 0001 bootstrap as removed for if tbr inaccessible after 3 minutes
> ```

alas manual merge needed for backport to 4.3.z for official ipv6 support

gabemontero:
see 4.3 PR noted above ^^

PR #221, opened 2020-01-17
DEXP-507: Ipv6 disc boot removed 2

gabemontero:
based on initial test runs with disconnected, proxy by @xiuwang 

@openshift/openshift-team-developer-experience @bparees fyi

/assign @adambkaplan

### DIFF
--- a/manifests/03-rbac-proxies-role-binding.yaml
+++ b/manifests/03-rbac-proxies-role-binding.yaml
@@ -1,0 +1,11 @@
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: cluster-samples-operator-proxy-reader
+subjects:
+  - kind: ServiceAccount
+    name: cluster-samples-operator
+    namespace: openshift-cluster-samples-operator
+roleRef:
+  kind: ClusterRole
+  name: cluster-samples-operator-proxy-reader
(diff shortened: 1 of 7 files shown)


manifests/03-rbac.yaml
```diff
@@ -69,6 +69,22 @@ rules:
...
+  - apiGroups:
+      - config.openshift.io
+    resources:
+      - proxies
+    verbs:
+      - get
+      - list
+      - watch
+
+---
```
adambkaplan: CVO folks want us to move away from having lists in manifests. Please move this to a separate file. If you keep it prefixed `03-rbac-...` it will run in the ordering I believe we want.

See https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/operators.md#names-of-manifest-files

manifests/03-rbac.yaml
```diff
@@ -95,3 +111,17 @@ subjects:
 roleRef:
   kind: ClusterRole
   name: cluster-samples-operator
+
+---
+
+kind: ClusterRoleBinding
```
adambkaplan: ditto

manifests/03-rbac.yaml
gabemontero: done

manifests/03-rbac.yaml
gabemontero: done

adambkaplan:
/lgtm cancel

@gabemontero please squash commits

gabemontero:
squashed

gabemontero:
the new rbac files got lost in the shuffle @adambkaplan ... just pushed them, retagging

PR #222, opened 2020-01-17
Bug 1792102: Install imagestreams and templates on Power and Z [4.4]

yselkowitz:

PR #225, opened 2020-01-20
enable sample imagestreams and templates on Z

gabemontero:


gabemontero:
need https://github.com/openshift/origin/pull/24434 for e2e-aws-jenkins

gabemontero:
ok @pweil- @bparees @adambkaplan @ausil @yselkowitz here is my original version of the unlock s390/ppc commit (which I had pointed @yselkowitz to), along with the s390x/ppc64le content from openshift/library which we just merged/processed

We can either use this PR (associating it with @yselkowitz's BZs to facilitate backports) or he can duplicate this in his.

Per our meeting, placing a

/hold

on this until we get the comeback on the test coverage.

In the interim, I'll monitor the e2e's in case any non-CI flakes/issues occur.

tmp/build/library-sync.sh
```diff
@@ -13,13 +13,18 @@ pushd library-master
...
 mv * ..
 popd # official
 rmdir official
 popd # ocp-x86_64
+pushd ocp-ppc64le
+pushd official
+mv * ..
+popd # official
+rmdir official
+popd #ocp-ppc64le
```
yselkowitz: What about `ocp-s390x`?

tmp/build/library-sync.sh
```diff
@@ -29,6 +34,6 @@ git add operator
 rm t.tar
 rm -rf library-master
 
-echo "REMEMBER TO PATCH THE DATAGRID TEMPLATES TO apps/v1 FOR THEIR StatefulSets INSTANCES,"
-echo " and fix openjdk-8-rhel8:1.1 until jboss fixes,"
-echo " and nodejs* templates off of v8"
\ No newline at end of file
+echo "REMEMBER:"
+echo " - fix openjdk-8-rhel8:1.1 until jboss fixes,"
```
yselkowitz: Not relevant to this PR, but FYI this has been fixed in their `develop` branch, not sure why the library is using a specific tag for openjdk x86 images.

test/e2e/cluster_samples_operator_test.go
```diff
@@ -136,7 +136,7 @@ func verifyOperatorUp(t *testing.T) *samplesapi.Config {
 	return cfg
 }
 
-func verifyX86(t *testing.T) bool {
+func verifySupportedArch(t *testing.T) bool {
```
yselkowitz: If the e2e changes were a separate commit, it could be handled regardless of whose PR is merged.

tmp/build/library-sync.sh
gabemontero: they are in the process of switching to the `release`  tag/branch

tmp/build/library-sync.sh
gabemontero: interestingly I still see a s390x subfolder but will address / re-run thx

test/e2e/cluster_samples_operator_test.go
gabemontero: done

tmp/build/library-sync.sh
gabemontero: updated thanks

yselkowitz: right, without this you get e.g. `ocp-s390x/official/perl` instead of just `ocp-$ARCH/perl` like the other arches.

pkg/stub/handler.go
```diff
@@ -406,10 +406,6 @@ func (h *Handler) CreateDefaultResourceIfNeeded(cfg *v1.Config) (*v1.Config, err
 		cfg.APIVersion = v1.GroupName + "/" + v1.Version
 		cfg = h.updateCfgArch(cfg)
 		switch {
-		// TODO as we gain content for non x86 platforms we can remove the nonx86 check
-		case util.IsNonX86Arch(cfg):
```
yselkowitz: Any reason not to remove the now-unused `IsNonX86Arch` function from `pkg/util/util.go` like in my PR?

gabemontero: simple just in case / help refresh my memory / etc. ... understanding I may be the exception, I tend to not remove unused / commented out code immediately, but instead will go back periodically and do so.

gabemontero:
ok finally got a clean jenkins install, but of course are now running into the need for https://github.com/openshift/origin/pull/24448

adambkaplan:
/retitle Bug 1792102: Enable imagestreams and templates on Power and Z

adambkaplan:
@gabemontero makes sense to squash the 1st two commits? IMO leaving the imported in a separate commit is a good convention. Otherwise looks good.

gabemontero:
> @gabemontero makes sense to squash the 1st two commits? IMO leaving the imported in a separate commit is a good convention. Otherwise looks good.

thanks @adambkaplan 

separating the commits out stemmed from when there was some indecision regarding whether this PR or https://github.com/openshift/cluster-samples-operator/pull/222 would merge first (where this PR pulled content from openshift/library once the content was available from there, and the other PR was a manual update.

I'm inclined to leave things as is just in case.

But otherwise for all this, per our conf call on this, my understanding is the next step is confirmation that the image-eco/jenkins e2e's testing is in place, we are all comfortable, and then merges start happening here.

Of course let me know if you have heard something more recent.

adambkaplan:
bump @crawford @yselkowitz regarding getting CI set up to run the OpenShift Jenkins and image-eco suites on Power and Z.

Note that per discussion on the Group D arch call today, we are no longer considering feature backports outside of the IPv6 initiatives.

gabemontero:
Update:

We are now tracking https://issues.redhat.com/browse/MULTIARCH-149 from @jaypoulz wrt multi arch inclusion of the image-eco e2e's in their periodics (though that is just a piece of what Jeremy is tackling there).

That said, based on the outlook of when MULTIARCH-149 lands, we may pick https://github.com/openshift/cluster-samples-operator/pull/225/commits/5ae0bb2c3e361d6e26fa4e2d7b1cea158a4567c3 from this PR into a new one and merge it in order to resolve https://bugzilla.redhat.com/show_bug.cgi?id=1818476

gabemontero:
Update:

@jaypoulz tells me the s390 testing is imminent

ppc64le is still a few weeks out, so I'm going to restructure this PR to break out the s390 from ppc64le

when that is ready, we'll see about either creating a custom payload with the s390 commits, run the tests, see if there is any broken content to take out, open openshift/scl bugs as needed, etc.

or will merge the s390 commits vs. the custom payload, and same subsequent steps.

@jaypoulz and @andymcc will get back to me on the custom payload

@adambkaplan @pweil- FYI

gabemontero:
OK I've pushed the analogous commits for just enabling s390 content

I've got a branch https://github.com/gabemontero/cluster-samples-operator/tree/more-s390-2 that has these 3 commits plus the 3 analogous commits for power

per my earlier comment, based on how the upcoming testing and staging goes, we'll probably merge this s390 only PR prior to the power changes (which according to @jaypoulz are weeks out).

also based on the timings we'll see about breaking up the associated bugzillas if the power content cannot be merged by the time 4.5 wraps up

gabemontero:
Update: the multi-arch team is building a custom payload with this branch/PR  and will try out the image-eco/jenkins e2e

@jaypoulz told me they'll report here with the results when available

gabemontero:
Per discussions in slack, the ppc64le content tucked aware in a separate branch has been brought back in, along with recent fixes outside of this immediate effort.

As of yesterday, the s390 run was almost clean for image-eco (jenkins has been on the back burner for now).
The 3 errors were in 2 categories
a) one more instance of ruby24 testing to be no-op'ed on non-x86
b) a known mysql:8.0 issue

@jaypoulz and @ReninJose tell me they haven't had a clean ppc run yet, and are fighting some unrelated CI fires.

gabemontero:
today's update

we have passing image eco on s390x !!!

a jenkins e2e run is cooking now (less optimistic there, possibly lots of hidden centos usage to sort out)

power is having disk issues @jaypoulz is trying to sort through with his team

gabemontero:
update:  really close

PPC has 
- a ruby/rails item during builds the multi-arch guys are chasing
- mysql runs fine serially, but hiccups when run in the whole suite ... most likely a resource issue in the test system, and I'm not viewing it as a blocker, though I think @jaypoulz said they may poke around with it some more

for jenkins on s390, I got a round of both manual and e2e testing (from https://github.com/gabemontero/origin/commits/image-eco-s390-no-centos) succeeding using a s390 cluster they gave me

@ReninJose will be trying out jenkins PPC in the short term

gabemontero:
OK @ReninJose got a passing jenkins run in PPC !

the multi arch guys sorted out that a missing dep was not yum installed and that is why the ruby builds fail on PPC

I gave @jaypoulz a simply reproducer with explanations in slack ... the gist is that dependency needs to be added to the yum installs being done in the s2i ruby builder images ... for example, https://github.com/sclorg/s2i-ruby-container/blob/master/2.5/Dockerfile.rhel7#L37-L47

They'll connect with SCL

yselkowitz:
> the multi arch guys sorted out that a missing dep was not yum installed and that is why the ruby builds fail on PPC
> 
> I gave @jaypoulz a simply reproducer with explanations in slack ... the gist is that dependency needs to be added to the yum installs being done in the s2i ruby builder images ... for example, https://github.com/sclorg/s2i-ruby-container/blob/master/2.5/Dockerfile.rhel7#L37-L47

If installing libffi-devel into the shipped containers the only way to handle this (which would mean that only those native extensions whose devel dependencies are installed in the containers can be installed in an S2I build), then I suppose so.  Or, given that at least python's image also installs it, should this be centralized into s2i-base with the rest of the devel packages?

gabemontero:
> > the multi arch guys sorted out that a missing dep was not yum installed and that is why the ruby builds fail on PPC
> > I gave @jaypoulz a simply reproducer with explanations in slack ... the gist is that dependency needs to be added to the yum installs being done in the s2i ruby builder images ... for example, https://github.com/sclorg/s2i-ruby-container/blob/master/2.5/Dockerfile.rhel7#L37-L47
> 
> If installing libffi-devel into the shipped containers the only way to handle this (which would mean that only those native extensions whose devel dependencies are installed in the containers can be installed in an S2I build), then I suppose so. Or, given that at least python's image also installs it, should this be centralized into s2i-base with the rest of the devel packages?

sure, assuming installing the library does not have an adverse effect on the other language builder images that do not need it, and it is needed on multiple language builders, putting it in the base is something to consider.

The SCL repo for that is the "s2-base-container" .... https://github.com/sclorg/s2i-base-container/tree/master/base is where the dockerfiles are stored.

yselkowitz:
While we should add libffi-devel to the ruby container directly or indirectly (via s2i-base), this should not block anything because it only relates to an old version of a particular gem, and technically equally applies to all architectures (we really want to use system libffi on x86 too, after all).  In the meantime, a simple workaround is to update to a (even slightly) newer ffi gem whose bundled libffi is new enough to support all our arches, to which end I have filed https://github.com/sclorg/rails-ex/pull/126.

jaypoulz:
Documented Yaakov's analysis in ticket: https://projects.engineering.redhat.com/browse/RHELPLAN-47099

gabemontero:
/hold cancel

/lgtm

OK, @jaypoulz has opened https://projects.engineering.redhat.com/browse/RHELPLAN-47099 for tracking update of the s2i base image for the libffi-devel situation previously discussed here (where @yselkowitz 's *-ex PRs are the shorter term workaround.  With that, the existing mysql issue opened, which we can use for any customer cases that come in for these items, and the successes we've had, I'm calling testing complete here.  Removing the hold, adding the lgtm

Generic e2e updates from https://github.com/openshift/origin/pull/25014 can happen out of band.  The multi arch team can continue to use that branch in the interim.

Once this has merged and @jaypoulz / @ReninJose presumably can validate with official payloads vs. payloads built from this branch, we can move on to any back port discussions, using BZs per the openshfit process to make that happen.

gabemontero:
I'm leaving the commits as is (i.e. not squashed as we typically do) on purpose, as it could facilitate backports if/when that comes.

gabemontero:
in ci-flake cycle ... pulling out of merge queue, try to let CI settle

/hold

gabemontero:
@akram @waveywaves it looks like one of the 3 changes between to 2 of you to 
has broke pipeline_origin_bld.go again in 4.6 / master branch since I fixed it last month.

Perhaps related to you all trying to get the test passing in earlier versions of OCP (4.2 / 4.3 / 4.4) ?

/hold cancel

gabemontero:
everything else in e2e-aws-builds passed in the last run of it
plus, this PR is only updating s390x/ppc64le, hence no updates to the existing x86_64 templates/imagestreams 

hence chose ot use override

gabemontero:
@akram @waveywaves last run with the client plugin issue was https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-samples-operator/225/pull-ci-openshift-cluster-samples-operator-master-e2e-aws-builds/1274363517285175296

akram:
ack @gabemontero

PR #227, opened 2020-02-05
removed unused code generation related dependencies

gabemontero:
No longer needed since api and client have moved to openshift/api and openshift/client-to

@openshift/openshift-team-developer-experience FYI

/assign @adambkaplan 

I'll open a bz for this if we decide to do this now as part of 4.4 @adambkaplan , otherwise I'll just keep this on the back burner until master opens up for 4.5

/hold


dependencymagnet/doc.go
```diff
@@ -5,8 +5,5 @@
 package dependencymagnet
 
 import (
-	_ "github.com/jteeuwen/go-bindata/go-bindata"
-	_ "github.com/openshift/library-go/alpha-build-machinery"
-	_ "k8s.io/code-generator"
 	_ "github.com/openshift/api/samples"
```
adambkaplan: @gabemontero is this needed to fetch the vendored CRD YAML files?

gabemontero: yes @adambkaplan 

with the src code now importing `github.com/openshift/api/samples` this entire file is technically unnecessary

but this magnet file will be useful next time we deal with the chicken/egg scenario where we want to
- go mod vendor a new dependency
- then have your favorite golang editor find that dependency when you update your code to use the new dependency

so I left the one entry we still use

but if you feel strongly about it I can just delete the dependency magnet file

gabemontero: I'm going to go ahead and remove this file altogether @adambkaplan 

I've learned `go mod` is better/easier with golang 1.13 about bringing in new dependencies since originally crafting this PR.  This file is now "fully" unneeded IMO :-)

Will update shortly.

gabemontero:
hmm ... the git outage may have lost my last push .... I'll try again

gabemontero:
there we go .... now it is just a set of file deletes and removing two lines from go.mod

PR #228, opened 2020-02-18
Bug 1804763: add imagestream for oauthproxy image

bparees:


bparees:
@gabemontero do we have any sort of test coverage for these imagestreams?

bparees:
/hold
need @openshift/sig-auth acknowledgement since this is their image.

mfojtik:
/cc @sttts 

I don't see problem shipping this via image stream as people can use it today anyway (and they do). ImageStream only makes it available for disconnected users.

mfojtik:
/lgtm

Will wait for @sttts to unhold.

gabemontero:
> @gabemontero do we have any sort of test coverage for these imagestreams?

Not as part of image-eco @bparees 

I've tangentially seen some of the used in CI in other contexts of course (must-gather for artifact collection ... don't recall if that uses cli or not).

sttts:
/retest

As discussed on Slack with @bparees @jwforres @smarterclayton this only makes the image stream available to components outside of OpenShift core. This **does not imply any support or any future compatibility**. Flags will change.

/hold cancel

PR #235, opened 2020-02-22
[release-4.2] bug 1805250: add imagestream for oauthproxy image

bparees:


bparees:
/assign @gabemontero 

needed to manually pick it because 4.2 doesn't tolerate scheduled imports on SHA tags.

mfojtik:
The :latest tag name is wrong. It should be a version (v1, v0.0.1, v1-alpha1 or something). We should also add description to this image stream..

PR #237, opened 2020-03-02
Revert "add imagestream for oauthproxy image"

sttts:
This reverts #228


mfojtik:
@bparees please approve, we need to find out better solution that will work long term for us.

bparees:
/approve

overriding the BZ requirement

PR #238, opened 2020-03-02
Revert "add imagestream for oauthproxy image"

sttts:
Reverting #228 for master.

This reverts commit 053daf06bad2b21b47eb441a8e921c88e7c956b0.


mfojtik:
@bparees please approve, we need to find out better solution that will work long term for us.

PR #241, opened 2020-03-10
manifests: add openshift/oauth-proxy:v4.4 image

mfojtik:


gabemontero:
the 4.4 tag is OK for master/4.5 @mfojtik ?

mfojtik:
@gabemontero yes, it is OK... in fact we plan to ship v4.4 tag into 4.3 as well.

PR #246, opened 2020-03-13
Bug 1813175: abort metrics in general if removed/unmanaged

gabemontero:


pkg/metrics/metrics.go
```diff
@@ -106,10 +112,9 @@ func (sc *samplesCollector) Collect(ch chan<- prometheus.Metric) {
 	importFailuresReason := importFailures.Reason
 	for _, stream := range streams {
 		_, skipped := skips[stream]
-		dontCare := cfg.Status.ManagementState == operatorv1.Removed || cfg.Status.ManagementState == operatorv1.Unmanaged
 
-		if skipped || dontCare {
-			// unmanaged/removed/skipping-an-imagestream means we consider the metric as 0 since we don't care if it failed import
+		if skipped {
```
coreydaley: This might be more concise code

```
        skips := make(map[string]struct{})
	skips["somestream"] = struct{}{}
	
	if _, ok := skips["somestream"]; ok {
		//do some stuff if the stream is in the skips map
	}
```

gabemontero:
thanks @coreydaley !

PR #252, opened 2020-03-23
manifests: Add the new 'tools' image as an image stream for debug

smarterclayton:
The tools image is the base for 'tests' and will be used by default
for debug, ensuring that admins get a fast and quick deploy (because
the base image is reused). We will add necessary administrative tools
to the image over time to ensure admins are happy.

/hold

until the image is available in OSBS and the nightly streams


smarterclayton:
Built on both sides

/hold cancel

PR #254, opened 2020-04-02
WIP: payload only streams nonx86

gabemontero:
will possibly do this for https://bugzilla.redhat.com/show_bug.cgi?id=1818476

the intention is to copy the install pull secret to the openshift namespace, bootstrap as managed but with only content installed by the CVO per the manifest files

this is the first 2 commits from https://github.com/openshift/cluster-samples-operator/pull/225

after a regression test pass on x86 as is, I will put a temporary commit that removes all x86 samples, to mimic what will occur on ppc/z


gabemontero:
image-eco suffered from the current github incident

gabemontero:
ok pushing the temp no x86 content mimic what will happen in ppc/s390

gabemontero:
install failed 

/test e2e-aws-operator

gabemontero:
ok the samples operator is what is holding up the install in e2e-aws-operator

going to have to work on this manually in a local cluster

PR #258, opened 2020-04-06
handle concurrent metrics_test.go hiccups

gabemontero:


gabemontero:
> /lgtm
> 
> As you've discovered, we run tests in parallel by default

yes even with https://github.com/openshift/cluster-samples-operator/blob/master/Makefile#L18

just finally got aggravated enough by it after a few months

PR #260, opened 2020-04-08
Bug 1822381: revert pr 177; no override jenkins* registry hosts for mirrors

gabemontero:
Details in bz, but with import stream via sha in, we don't need to override the jenkins image stream registry to support disconnected, and that gets in the way in other scenarios

/assign @bparees 

see https://github.com/openshift/cluster-samples-operator/pull/177


vrutkovs:
That also seems to affect 4.4 - @gabemontero could you file a bug and cherrypick it there?

PR #249, opened 2020-03-18
remove pull secret copy, defer to image registry

gabemontero:
With @ricardomaraschini 's https://github.com/openshift/image-registry/pull/226 merged here is my change here the change that remove's samples operator's copying of the install pull secret to the openshift namespace

I still have my prometheus metrics around whether the pull secret has TBR credentials, but it now just looks at the install pull secret directly.

Have not tried it in a local cluster yet so marked WIP for now

@openshift/openshift-team-developer-experience FYI

PR #264, opened 2020-04-20
Bug 1825774: add back missing parts to make openshift-config namespace secret list…

gabemontero:
…er functional

got a little too aggressive in removing operator related bits for the openshift-config secret lister

still need to create a workqueue and informer to enable it to obtain the secret

even though we no longer copy the secret, still inspect it to make sure it has creds for registry.redhat.io

/assign @ricardomaraschini 
/assign @dmage 

if one of you all have cycles please review - thanks

### DIFF
--- a/pkg/operator/controller.go
+++ b/pkg/operator/controller.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 
+	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metaapi "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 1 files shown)


gabemontero:
was able to verify on a local cluster ... able to retrieve the secret now

gabemontero:
> It looks good to me but I admit I don't understand exactly why we need to react to changes on these secrets. Would you mind explain a little bit in here?

Yep I'll add some comments in the code.  I gave a very terse summary in this PR's description, but perhaps too terse.  For posterity, I'll clarify here as well:
1) rather than use a vanilla client to access the openshift-config secrets, we want to use the lister/sharedinformer/workqueue controller apparatus so we get watches and caching (so we don't have to hit the api server whenever we access the data)
2) but you cannot just partially set that up, as I originally attempted in  https://github.com/openshift/cluster-samples-operator/pull/249
3) otherwise the cache never gets populated, and we get "not found" errors when we attempt to get secrets that actually exist

gabemontero:
ok @ricardomaraschini I've added the comment explaining stuff ... let me know if you think that is sufficient, and if it is, I'll squash, etc.

thanks

ricardomaraschini:
> ok @ricardomaraschini I've added the comment explaining stuff ... let me know if you think that is sufficient, and if it is, I'll squash, etc.
> 
> thanks

Yep, thanks for that extra explanation. Squash them and I will LGTM.

gabemontero:
commits squashed @ricardomaraschini  - thanks

gabemontero:
unit test flake gone

PR #240, opened 2020-03-06
Bug 1811143: purge removed imagestreams as a part of upgrade from progessing/impor…

gabemontero:
…terrors

from @vrutkovs 's 4.3.5 upgrade testing today

@openshift/openshift-team-developer-experience FYI

See bugzilla for analysis ... I believe hitting this is low.

@adambkaplan or @bparees ... preferences on who wants to be assigned ?

PR #265, opened 2020-04-27
Bug 1828065: correctly handle removed templates watch events as part of an upgrade

gabemontero:
basically https://github.com/openshift/cluster-samples-operator/pull/240 was incomplete with respect to samples that are removed from release to release

there is a window where removed templates can gum up the works where repeated attempts to process removed templates result in repeated 

```
time="2020-04-27T06:43:12Z" level=info msg="open : no such file or directory error reading file []"
time="2020-04-27T06:43:12Z" level=info msg="CRDUPDATE event temp udpate err"
```

which leads to a storm of addtional updates to the samples operator config object 
that can severely slow down getting the progressing condition to false

/assign @bparees 
/assign @adambkaplan 

@bparees I assigned you in case you want this expedited and included in 4.4.0 ... I'm guessing the answer to that is "no" based on the current state of the 4.4.0 upgrade BZ bugs but I did not want to assume.  Feel free to unassign / comment as you deem fit.  Thanks.

### DIFF
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -120,11 +120,11 @@ type Handler struct {
 	imageclientwrapper    ImageStreamClientWrapper
 	templateclientwrapper TemplateClientWrapper
 
-	crdlister           configv1lister.ConfigLister
-	streamlister        imagev1lister.ImageStreamNamespaceLister
-	tplstore            templatev1lister.TemplateNamespaceLister
-	cfgsecretlister     corev1lister.SecretNamespaceLister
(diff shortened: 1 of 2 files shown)


bparees:
> @bparees I assigned you in case you want this expedited and included in 4.4.0 ... I'm guessing the answer to that is "no" based on the current state of the 4.4.0 upgrade BZ bugs but I did not want to assume.

correct, it's a no for 4.4.0 from me as long as we expect the upgrade will eventually complete in some reasonable timeframe (hours).

gabemontero:
all green tests

PR #284, opened 2020-06-03
pull in sso compatibility fixes for 4.5 and above

gabemontero:
/assign @marun 

heads up, sso-cd-x509-https.json was not previously shipped with OCP, so its deletion in openshift/library did not need to occur here.


gabemontero:
I'll cherry pick back to 4.5 and use your BZ after you lgtm this and it merges @marun

PR #288, opened 2020-06-05
Fix samplesRegistry replacement in imageStreams when using paths

aerdei:
Currently, when defining samplesRegistry in the CR with a value of newhost:1111/path, the resulting imageStream DockerImage values become newhost:1111/path/path/path instead of newhost:1111/path/oldpaths.
This PR makes it possible to have a path in samplesRegistry in the CR. This could be beneficial when using external registries with paths (e.g. Artifactory remote registries).

### DIFF
--- a/pkg/stub/handler_test.go
+++ b/pkg/stub/handler_test.go
@@ -331,6 +331,10 @@ func TestProcessed(t *testing.T) {
 	if is == nil || !strings.HasPrefix(is.Spec.DockerImageRepository, cfg.Spec.SamplesRegistry) {
 		t.Fatalf("stream repo not updated %#v, %#v", is, h)
 	}
+	is, _ = fakeisclient.Get("baz")
+	if is == nil || !strings.HasPrefix(is.Spec.DockerImageRepository, cfg.Spec.SamplesRegistry) {
+		t.Fatalf("stream repo not updated %#v, %#v", is, h)
+	}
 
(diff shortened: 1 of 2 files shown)


gabemontero:
> Currently, when defining samplesRegistry in the CR with a value of newhost:1111/path, the resulting imageStream DockerImage values become newhost:1111/path/path/path instead of newhost:1111/path/oldpaths.
> This PR makes it possible to have a path in samplesRegistry in the CR. This could be beneficial when using external registries with paths (e.g. Artifactory remote registries).

Interesting / thanks @aerdei !

Any chance you add a unit test that validates this change (it did pass the existing validations)?

The current validations are in https://github.com/openshift/cluster-samples-operator/blob/master/pkg/stub/handler_test.go#L294, admittedly a somewhat involved flow.  If you want to take it on there, great.  Otherwise, a new test method which validates your scenario and simply calls `coreUpdateDockerPullSpec` is fine with me as well.

gabemontero:
/approve

I'll also note there is an RFE on the books for expanding the power of this aspect to override more than just the registry host:port, but it has not been prioritized as of yet.

aerdei:
Hi @gabemontero ,
Thanks! I have updated the unit tests. I also took the liberty to add an additional test imageStream file that has a .spec.tags.from.name that more closely resembles current use cases (DockerImage names in openshift/library imageStreams). Let me know what you think!

gabemontero:
/lgtm

thanks again @aerdei

PR #297, opened 2020-06-25
Bug 1792102: Install imagestreams and templates on Power and Z [4.4]

yselkowitz:
This is a backport of #225, with only minor patch context changes in patches 0001 and 0004.


gabemontero:
/override ci/prow/e2e-aws-image-ecosystem

upstream introduced a regression with django postgreql yesterday, which is causing image eco to fail...that was the only failure in the last run ... we are working with upstream on how to address, but in the interim, overriding to unblock this PR

PR #299, opened 2020-07-01
Bug 1852997: Update Jenkins Templates with trusted-ca-bundle changes

waveywaves:


gabemontero:
/lgtm

/hold

@waveywaves unhold the PR once you get past CI flakes and get clean tests 

of course if only e2e-aws-jenkins fails consistently perhaps something is wrong with the updated templates

waveywaves:
@gabemontero I have updated the template to the right one. I will retrigger the CI later, there are too many flakes.

waveywaves:
@gabemontero

waveywaves:
@gabemontero Can you PTAL ?

gabemontero:
/hold

@waveywaves there has been some restructuring of the samples payload since you submitted this PR

minimally, the tmp/build has been removed from the path ... as it stands now, this change won't be picked up

I'll also take a look at the end to end flow here, form the templates you have defined in openshift/jenkins, to the official.yaml updates you all have in openshift/library, and make sure we are injecting this via the proper process.  We may be, but I'll double check and report back.

gabemontero:
Yeah @waveywaves @jkhelil are not the trusted-ca changes not already present in https://github.com/openshift/library/blob/master/operator/ocp-x86_64/official/jenkins/templates/jenkins-ephemeral.json#L41-L50 and https://github.com/openshift/cluster-samples-operator/blob/master/assets/operator/ocp-x86_64/jenkins/templates/jenkins-ephemeral.json#L41-L50

At least if I am reading the git history correctly at https://github.com/openshift/jenkins/commits/master/openshift/templates/jenkins-ephemeral.json and I understand what the "trusted-ca-bundle changes" means.

In less you all are concerned with something else, I think we can just close this PR.  But please confirm.

PR #301, opened 2020-07-01
Bug 1853070: update base image version in dockerfiles

gabemontero:
started to see errors in CI with the old 4.0 stuff


gabemontero:
Also need to fix https://github.com/openshift/release/blob/master/ci-operator/config/openshift/cluster-samples-operator/openshift-cluster-samples-operator-master.yaml

PR #307, opened 2020-07-24
Include both OCP and OKD content during library sync

vrutkovs:
This would ensure `assets` would include both OCP and OKD content. OKD-specific Dockerfile which includes OKD-only content is created  (OCP content would remain intact).

Operator is updated to handle both types of content, so samples-operator built from `Dockerfile.okd` would serve OKD only content. This Dockeffile is not used yet - this would be addressed in follow up PRs.

Related to https://github.com/openshift/okd/issues/34


gabemontero:
Going to launch a cluster off of this PR and poke around a little bit prior to officially reviewing

Certainly I see the desire to have samples in OKD.

And historical reference, back in early 4.0, we had a OKD vs. OCP install notion before it became apparent that OKD was a ways off.

I suspect we will want some more configuration around this, and possibly even an EP.  But I'll get more solid one way or the other with that opinion after finishing my review.

@bparees @sbose78 @adambkaplan @coreydaley FYI

gabemontero:
/test unit 

(known flake)

gabemontero:
OK so using both OCP and OKD images from this PR's branch in test clusters, things look as I would expect.

I'm generally OK with the code changes as is.  And after thinking about it a bit more, I'm fine with the switch being at the image level vs. config in the samples operator.

And tests are green.

Before merging, I'd still like to hear opinions from @bparees @sbose78 @adambkaplan on whether we need an EP (simple as it might be) for adding this.

thanks @vrutkovs

bparees:
so OKD content installs w/ OKD and OCP content installs with OCP? (no OKD content installed w/ OCP, right?)

i don't think we need an EP as long as there's no plan to allow admins to "pick" their content, or switch between the two later, etc.

Any implications for disconnected installs/mirroring doc, registry replacement behavior, or some that other fiddliness?

gabemontero:
> so OKD content installs w/ OKD and OCP content installs with OCP? (no OKD content installed w/ OCP, right?)

correct @bparees 
> 
> i don't think we need an EP as long as there's no plan to allow admins to "pick" their content, or switch between the two later, etc.

I know of no such plans.  @vrutkovs - is that correct? ... i.e. no plans to allow switching between OKD and OCP content at some point, or possibly "upgrade" a cluster from OKD to OCP (though maybe a holistic change like that would not matter)?

> 
> Any implications for disconnected installs/mirroring doc, registry replacement behavior, or some that other fiddliness?

docker.io and quay.io were already in the replacement list ... those are the only two registries I see in the OKD imagestreams
so I think we are good there

but good question @bparees :-)

vrutkovs:
> > i don't think we need an EP as long as there's no plan to allow admins to "pick" their content, or switch between the two later, etc.

>I know of no such plans. @vrutkovs - is that correct? ... i.e. no plans to allow switching between OKD and OCP content at some point, or possibly "upgrade" a cluster from OKD to OCP (though maybe a holistic change like that would not matter)?

The change is contained within one operator, no other changes required - I'd prefer to avoid EP process. Mirroring and disconnected cases don't require any updates - its simply an operator update, resulting in the same imagestreams on OCP.
There is no plan to let admins choose OCP w/ OKD content or vice versa.

We certainly would love to experiment with OKD -> OCP upgrade (the other way around is probably pointless) but we're not there yet. What are the implications of that? Seems the "upgrade" would complete, but some imagestream might be unavailable? That can be specified in release notes

gabemontero:
> > > i don't think we need an EP as long as there's no plan to allow admins to "pick" their content, or switch between the two later, etc.
> 
> > I know of no such plans. @vrutkovs - is that correct? ... i.e. no plans to allow switching between OKD and OCP content at some point, or possibly "upgrade" a cluster from OKD to OCP (though maybe a holistic change like that would not matter)?
> 
> The change is contained within one operator, no other changes required - I'd prefer to avoid EP process. Mirroring and disconnected cases don't require any updates - its simply an operator update, resulting in the same imagestreams on OCP.
> There is no plan to let admins choose OCP w/ OKD content or vice versa.

OK sounds good.  No EP.

> 
> We certainly would love to experiment with OKD -> OCP upgrade (the other way around is probably pointless) but we're not there yet. What are the implications of that? Seems the "upgrade" would complete, but some imagestream might be unavailable? That can be specified in release notes

I believe it would at most boil down to deleting imagestreams from OKD which do not exist in OCP (as they would not fall under the OCP support contract).  For imagestreams with common names, the api server update should replace the OKD version with the OCP version.  Same with templates.

But getting consensus on my take on this would be the discussion to be had, either in github PR fashion or OCP EP fashion.

But as you noted, we are not there yet, so a discussion for another day.

/lgtm

bparees:
> I believe it would at most boil down to deleting imagestreams from OKD which do not exist in OCP (as they would not fall under the OCP support contract). For imagestreams with common names, the api server update should replace the OKD version with the OCP version. Same with templates.

deleting imagestreams (or replacing them with OCP equivalents) would be problematic since it would potentially break existing workloads that referenced them.

but as you say, discussion for another time.

vrutkovs:
/hold

`ci/prow/e2e-aws-image-ecosystem` is permafailing

gabemontero:
so we've had problems with that cakephp mysql sample in the past that lead to https://bugzilla.redhat.com/show_bug.cgi?id=1793116

the stack trace in the recent failures looks *similar* ... though I am not knowledgeable enough of mysql to say for sure after a quick glance

while upstream SCL was sorting things out, we maintained the use of mysql:5.7, keeping it in the imagestream and manually updating the templates after the import from openshift/library

recently things were suppose to have been fix, and I saw a passing image eco, though when I look at the PR test history now, given all the failures, I think those passing tests have been pruned from the list

So either
a) something changed upstream between that passing test and my lgtm
b) or I misread the passing test thing (though I was pretty sure)

Unless somebody beats me to it, on Monday
a) I'll look at https://github.com/sclorg/cakephp-ex and https://github.com/sclorg/mysql-container to see if there were any recent changes that may have regressed things
b) otherwise, a manual update to this PR to pull back in mysql:5.7 into the mysql imagestream and update the cakephp templates to use 5.7 should be attempted

gabemontero:
networking/sdn degraded on image eco install

/test e2e-aws-image-ecosystem

vrutkovs:
/hold cancel

Pushed a commit which re-adds mysql 5.7 and adjusts cakephp examples to use it

PR #308, opened 2020-07-28
switch to mysql:8 and enable auth plugin needed for running in openshift

gabemontero:
Pull in fix for https://bugzilla.redhat.com/show_bug.cgi?id=1793116

upstream version:  https://github.com/sclorg/cakephp-ex/pull/116

PR #309, opened 2020-07-30
OKD: add mysql 5.7 imagestream

vrutkovs:
CakePHP template still requires MySQL 5.7


vrutkovs:
A better way would be moving off 5.7 though, #308

PR #310, opened 2020-07-31
Update OKD contents sync script

vrutkovs:
Addresses https://github.com/openshift/cluster-samples-operator/pull/308#issuecomment-667179164


gabemontero:
yep it looks like that does the trick @vrutkovs .... just a question of whether this PR or https://github.com/openshift/cluster-samples-operator/pull/308 merges first and if whoever comes after has to rebase

PR #317, opened 2020-08-26
Bug 1872080: Sync RHEL/CentOS/UBI 8 imagestreams changes

yselkowitz:

PR #319, opened 2020-08-27
[TESTONLY] update SCL templates to EL/UBI 8 imagestreams

yselkowitz:
This PR, built on top of #317, is intended to test changing the templates to EL8/UBI8 imagestreams tags through image eco.


yselkowitz:
Interesting, according to the test results, both OCP and OKD image eco e2e passed with this.  @gabemontero anything else I should consider before pushing template changes into the SCL repos?

gabemontero:
the e2e-aws run had errors unrelated to samples

okd-e2e-aws-operator is the known issue I'm chasing today that I think should be resolved soon

let's see how okd-e2e-aws-builds does and if that passed, I think we can claim this successful

yselkowitz:
OK.  Probably hard to tell from the diff, but besides just moving the templates to -el8/-ubi8 tags, this also moves dancer from mysql 5.7 to 8.0 (with the same MYSQL_DEFAULT_AUTHENTICATION_PLUGIN fix used in cakephp) and the dotnet example from 3.0 (which is EOL) to 3.1.  Does image eco test those too?

gabemontero:
> OK. Probably hard to tell from the diff, but besides just moving the templates to -el8/-ubi8 tags, this also moves dancer from mysql 5.7 to 8.0 (with the same MYSQL_DEFAULT_AUTHENTICATION_PLUGIN fix used in cakephp) and the dotnet example from 3.0 (which is EOL) to 3.1. Does image eco test those too?

dancer proved way too flaky with its downloads so it has been removed:  https://github.com/openshift/origin/blob/master/test/extended/image_ecosystem/sample_repos.go#L178-L206

a TEST only PR to openshift/origin after this stuff merges is probably your path of least resistance to ultimately try it out 

dotnet as I recall was never included in image-eco

yselkowitz:
Any thoughts on the remaining failures?

gabemontero:
okd-e2e-aws is a CI flake
okd-e2e-aws-upgrade failed with more core components than samples getting hit by possibly env issues
okd-e2e-aws-operator was an aws/terraform install flake

but let's rerun okd-e2e-aws-operator with the prow fix and try to confirm your okd/centos updates at least create the expected imagestreams both scl and non-scl

/test okd-e2e-aws-operator

yselkowitz:
Don't think we need to keep dealing with these flakes, since this isn't being merged anyway.  We'll pick this up again once everything is ready to be merged through the usual channels.

PR #322, opened 2020-09-09
Update Jenkins nodejs-agent image to v12

waveywaves:


gabemontero:
I believe @yselkowitz has already set things into motion to change the nodejs agent to v12 and ubi8 based

If this is not seen in the 4.6 payloads yet, I suspect it is simply in progress downstream in brew/art land ... he can clarify.

But as I explained to @waveywaves in https://coreos.slack.com/archives/C014MHHKUSF/p1599650249045700 .... image providers directly modifying sample operator repo is not the process for updating images in the samples imagestreams.

PR #331, opened 2020-10-07
Bug 1886890: Override jenkins-agent-base imagestream

yselkowitz:
Just like jenkins-agent-maven and jenkins-agent-nodejs, derive the
pullspec for jenkins-agent-base from the release payload.

I'm thinking this should be backported to 4.6 as well so that it doesn't
start picking up 4.7+ versions once those GA.

### DIFF
--- a/manifests/06-operator.yaml
+++ b/manifests/06-operator.yaml
@@ -60,6 +60,8 @@ spec:
             value: "0.0.1-snapshot"
           - name: IMAGE_JENKINS
             value: quay.io/openshift/origin-jenkins:latest
+          - name: IMAGE_AGENT_BASE
+            value: quay.io/openshift/origin-jenkins-agent-base:latest
           - name: IMAGE_AGENT_NODEJS
             value: quay.io/openshift/origin-jenkins-agent-nodejs:latest
           - name: IMAGE_AGENT_MAVEN
--- a/manifests/image-references
+++ b/manifests/image-references
@@ -30,6 +30,10 @@ spec:
     from:
       kind: DockerImage
       name: quay.io/openshift/origin-jenkins:latest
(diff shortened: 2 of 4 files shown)


yselkowitz:
If you choose to accept this, then jenkins-agent-base needs to be changed to `for_payload: true` in ocp-build-data; I can file that PR too.

gabemontero:
/test e2e-aws-jenkins

ocp in addition to the okd run that has already passed

gabemontero:
/assign @akram 

for awareness ... since you have an agent base imagestream defined in https://github.com/openshift/jenkins/blob/master/openshift/imagestreams/jenkins-agent-base-rhel.yaml it does make sense to override the image with the payload like we do for jenkins, maven, nodejs

but let us know if you have any questions or concerns .... if @akram is good, we can approve/lgtm

@yselkowitz if this were to get backported to 4.6.z (it is too late for 4.6.0) then we'd need to get a bugzilla opened and assoicated with the PR so we can employ the cherrypick/bz clone process that is expected for getting stuff in the z-stream in OCP.

Either the SAMPES or JENKINS components should suffice.

gabemontero:
> If you choose to accept this, then jenkins-agent-base needs to be changed to `for_payload: true` in ocp-build-data; I can file that PR too.

cool / thanks @yselkowitz

gabemontero:
bump @akram 

have you had a chance to consider this?

gabemontero:
@akram has given me his +1 in slack

/lgtm

akram:
yes approved , sorry for the delay @gabemontero and @yselkowitz

gabemontero:
unrelated ci/env flake on that last image-eco run

/test e2e-aws-image-ecosystem

PR #339, opened 2020-11-18
Add IBM Cloud managed annotations to CVO manifests

csrwng:
This matches https://github.com/openshift/enhancements/pull/445 and does not change behavior.


gabemontero:
/hold

putting out a fire elsewhere, will unhold when done just in case I have to change manifests files

PR #340, opened 2020-11-18
Bug 1898745: actually set imagestreams in progress to false in imagestream event path

gabemontero:
@coreydaley we missed something in code/review that has shown up in CI this evening :-(

PR #338, opened 2020-11-12
set progressing false on imagestream events as well if no active streams

gabemontero:
Seen a few runs in CI today for various repos fail where the the samples operator did not get out of progressing == true state

When examining the pods logs, some odd message after our attempt to update progressing to false after getting a configmap delete event failed:

```
2020-11-12T17:29:36.043718980Z W1112 17:29:36.043675      16 reflector.go:424] github.com/openshift/client-go/image/informers/externalversions/factory.go:101: watch of *v1.ImageStream ended with: an error on the server ("unable to decode an event from the watch stream: stream error: stream ID 225; INTERNAL_ERROR") has prevented the request from succeeding
2020-11-12T17:29:45.622049521Z time="2020-11-12T17:29:45Z" level=info msg="CRDERROR progressing false update"
2020-11-12T17:29:45.622089759Z time="2020-11-12T17:29:45Z" level=error msg="unable to sync: context deadline exceeded, requeuing"
2020-11-12T17:29:51.556538089Z time="2020-11-12T17:29:51Z" level=error msg="unable to sync: Timeout: request did not complete within requested timeout 34s, requeuing"
```

Even with returning an error, the delete event for the config map is not requeued.

We continue to get imagestream events.

This change (re)adds attempts to set progressing to false on imagestream events as well, when it is clear all the imagestream configmaps are gone.  We may get a few conflicts when several imagestreams come in at the same time, but in the end that is no big  deal.

PR #313, opened 2020-08-13
BUILD-125: move imagestream tracking out of operator config object

gabemontero:
@sbose78 @adambkaplan @coreydaley @otaviof 

For today's hackday,
a first pass at something I've mentioned in a few team meetings in the recent past.

Move tracking of the samples imagestream progress out of the operator config object and into separate objects per imagestream (configmaps in this case).

I was able to remove a fair bit of the complicated caching code around minimizing updates to the config object to avoid conflicts.

Right now, the configmap data map is not getting used, but a subsequent step of moving the import errors out of the operator config object and into these per imagestream objects could leverage that.  Didn't have the bandwidth to pursue that today.

Local bring up testing looks good (*seem* a tad faster but I did not have time to measure today ... absolutely no conflicts across multiple runs ... I usually see a handful after an afternoon of testing), and a local run of the operator e2e was passing but only halfway through when I signed off for the day.  I'll report here on the local run later ... otherwise we'll see how this PR's run does.

PR #341, opened 2020-11-18
Bug 1898745: deprecate ImageChangesInProgress, make sure it stays false

gabemontero:
/hold 

during the analysis of upgrade issues exposed via search.ci that lead to https://github.com/openshift/cluster-samples-operator/pull/340

another simplification occurred to me that goes with our move off of tracking conditions on the samples operator cluster scoped config resource  .... deprecating the `ImageChangesInProgress` condition and use the newer approaches for tracking imagestream status, as well as the approach started in 4.6 of not gating the report of being ready to the CVO, namely
- active == true
- progressing == false
- degraded == false
- version set
on the state of imagestreams after we have created them (i.e. their imports do not gate the process)

Now, so far (after about 15 hours),  since #340 merged, the samples upgrade has appeared clean on search.ci, so I'm not chomping at the bit to merge this.

Minimally, I want to get a longer sample size.  If some more flakes show up in the coming days, and it looks like this change will help, we'll go sooner.

Otherwise, it is a judgement call with the team as to whether we go with this before 4.7 DCUT, or wait until 4.8.

But long term, I really like this simplification.  And it continues along the pivoting we've been doing with samples since we've come to the agreement that tying imports and dependencies on registry.redhat.io does not lend well to CVO status.

@pedjak - this PR, and my notes above, might provide educational value to your team as we move closer to transitioning.  It touches on the recent changes I mentioned today, as well as some of the key concepts, with respect to both how this operator deals with its operands, as well as how it reports to its operator, the CVO.  Could be worth you all taking a look as part of your educational cycles.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,8 +64,7 @@ The samples resource maintains the following conditions in its status:
 - SamplesExists
 -- Indicates the samples have been created in the openshift namespace
 - ImageChangesInProgress
--- True when imagestreams have been created/updated, but we are awaiting acknowledgment all the changes were applied  
--- False when all imagestream changes are acknowledged 
+-- This condition is deprecated as of the 4.7 branch of this repository.  `ImageStream` image imports are no longer tracked in real time via conditions on the samples config resource, nor do in progress `ImageStreams` directly affect updates to the `ClusterOperator` instance `openshift-samples`.  Prolonged errors with `ImageStreams` are reported now by Prometheus alerts. 
 -- The list of pending imagestreams will be represented by a configmap for each imagestream in the samples operator's namespace (where the imagestream name is the configmap name).  When the imagestream has completed imports, the respective configmap for the imagestream is deleted.
 - ImportCredentialsExist
 -- Credentials for pulling from `registry.redhat.io` exist in the `pull-secret` Secret in the `openshift-config` namespace
--- a/pkg/stub/config.go
+++ b/pkg/stub/config.go
@@ -216,9 +216,9 @@ func (h *Handler) buildFileMaps(cfg *v1.Config, forceRebuild bool) error {
 		}
 	}
 
-	cm, err := h.configmapclientwrapper.Get(util.IST2ImageMap)
+	cmFromControllerCache, err := h.configmapclientwrapper.Get(util.IST2ImageMap)
 	if kerrors.IsNotFound(err) {
-		cm = &corev1.ConfigMap{}
+		cm := &corev1.ConfigMap{}
 		cm.Name = util.IST2ImageMap
 		cm.Namespace = v1.OperatorNamespace
 		cm.Annotations = map[string]string{}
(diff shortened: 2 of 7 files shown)


gabemontero:
@pedjak - the other PRs besides #340  that have been part of this 4.7 simplification / maintainability effort are 

- https://github.com/openshift/cluster-samples-operator/pull/338
- https://github.com/openshift/cluster-samples-operator/pull/313

@adambkaplan @coreydaley - check out this PR's description if you have cycles ... I'll also touch upon this in our next scrum.

gabemontero:
/test okd-unit 

(well known metric port conflict flake)

gabemontero:
I saw one more failure profile overnight in one particular periodic, along with a rehearsal job related to that periodic.

key particulars:  APIServerConflictError (on the samples exist update / samples config crd resource)
and a "fatal error: concurrent map iteration and map write" in the pod logs

from the search:  matched 0.09% of failing runs and 0.43% of jobs
over thousands of runs.

this PR, with the setting of available/progressing/version immediately, should mitigate the first item.  The second commit addresses the second.

I'll bring up in scrum today and we'll decide on who to review, etc.

/test e2e-aws-upgrade
(get a few more green runs)
/retest
(get past unrelated flakes)

gabemontero:
some docker.io throttling and unrelated flakes in e2e-aws and okd-e2e-aws

first additional upgrade e2e passsed

/test e2e-aws-upgrade
/test e2e-aws

gabemontero:
several components failed in upgrade job https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-samples-operator/341/pull-ci-openshift-cluster-samples-operator-master-e2e-aws-upgrade/1329438645215563776 but I have confirmed `openshift-samples` was not one of them

wking:
> ... as well as the approach started in 4.6 of not gating the report of being ready to the CVO ...

Do we have CI test-cases that depend on the imports having succeeded?  Because if folks are relying on the imports, and the imports fail or are slow, and the tests fail, but the samples operator is saying "no problems on my end", I expect there will be some confusion ;).

On the other hand, I'd also be fine with samples claiming it was happy regardless of import success, as long as any samples-consuming tests (there may be none) have their own checks for import success, and we can reasonably expect any other samples consumers to have their own, similar guards.

gabemontero:
> > ... as well as the approach started in 4.6 of not gating the report of being ready to the CVO ...
> 
> Do we have CI test-cases that depend on the imports having succeeded? Because if folks are relying on the imports, and the imports fail or are slow, and the tests fail, but the samples operator is saying "no problems on my end", I expect there will be some confusion ;).
> 

Yes we do @wking .... but see below ....

> On the other hand, I'd also be fine with samples claiming it was happy regardless of import success, as long as any samples-consuming tests (there may be none) have their own checks for import success, and we can reasonably expect any other samples consumers to have their own, similar guards.

When we moved off of marking samples degraded a release or two ago when imports failed, we augmented the verification we already had around this (which previously figured out if build test x failed cause the ruby imagestream could not import), including adding specific tests that just verified the imports we needed were there, to clearly isolate that root cause.

Good questions, but I think we are good here.  Thanks :-)

gabemontero:
will give okd-e2e-aws one more try, otherwise will override

/test okd-e2e-aws

gabemontero:
yep okd-e2e-aws are unrelated flakes we can skip/override as needed

but let's get a clean run of e2e-aws-builds in addition to okd-e2e-aws-builds

gabemontero:
in case build not being required bypasses things

PR #343, opened 2020-12-02
Annotate manifests for single-node-developer cluster profile

cfergeau:
This partially implements phase 1 of https://github.com/openshift/enhancements#482
and does not change behavior. Initially, all cluster-samples-operator
manifests are included in the single-node-developer cluster profile.
Follow-on PRs may exclude any of these that are not needed in the
profile.

### DIFF
--- a/manifests/01-namespace.yaml
+++ b/manifests/01-namespace.yaml
@@ -4,8 +4,9 @@ metadata:
   annotations:
     include.release.openshift.io/ibm-cloud-managed: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
-    openshift.io/node-selector: ""
+    include.release.openshift.io/single-node-developer: "true"
     include.release.openshift.io/single-node-production-edge: "true"
+    openshift.io/node-selector: ""
   labels:
     openshift.io/cluster-monitoring: "true"
   name: openshift-cluster-samples-operator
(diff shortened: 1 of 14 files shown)


manifests/01-namespace.yaml
```diff
@@ -4,8 +4,9 @@ metadata:
   annotations:
     include.release.openshift.io/ibm-cloud-managed: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
-    openshift.io/node-selector: ""
+    include.release.openshift.io/single-node-developer: "true"
     include.release.openshift.io/single-node-production-edge: "true"
+    openshift.io/node-selector: ""
```
gabemontero, line 9: can you clarify / remind me the resulting behavior of the node-selector annotation setting above @cfergeau ?

gabemontero: /hold 

pending the answer to the above question

cfergeau: The `node-selector` annotation is already there before this change, I moved it to group all `include.release.openshift.io` annotations together, but I'm only adding a ` include.release.openshift.io/single-node-developer: "true"` annotation

Before:
```
  annotations:
    include.release.openshift.io/ibm-cloud-managed: "true"
    include.release.openshift.io/self-managed-high-availability: "true"
    openshift.io/node-selector: ""
```

After:
```
  annotations:
    include.release.openshift.io/ibm-cloud-managed: "true"
    include.release.openshift.io/self-managed-high-availability: "true"
    openshift.io/node-selector: ""
```

gabemontero: ahh ... sorry, missed that ... my bad

/hold cancel

/lgtm

PR #344, opened 2020-12-03
pull latest openshift/library for updating samples for 4.7

gabemontero:

PR #346, opened 2020-12-03
Imagestreams update

yselkowitz:
This includes the following updates:
* Update golang to 1.14.7 in UBI 8
* Add nginx 1.18 in UBI 7 and 8
* Add nodejs 14 in UBI 7 and 8
* Add perl 5.30 in UBI 8
* Add php 7.4 in UBI 8
* Add ruby 2.7 in UBI 8
* Update RHDM and RHPAM from 7.8 to 7.9
* Update SSO to 7.4.3, including ppc64le and s390x
* Update JWS to 5.4, including ppc64le and s390x


gabemontero:
I opened https://github.com/openshift/cluster-samples-operator/pull/344 earlier today @yselkowitz 

and it was the results from `./library-sync.sh`

yselkowitz:
I got a few more updates into SCL imagestreams today so you might need to update yours then

yselkowitz:
nm the diffstat appears to be the same so I guess we were doing this simultaneously.
/close

gabemontero:
> I got a few more updates into SCL imagestreams today so you might need to update yours then

I just re-ran ./library-sync.sh in my branch and there we no deltas ... think we are good

PR #348, opened 2020-12-08
Bug 1905319: request required minimal mem resource per openshift conventions

gabemontero:
patterned off of fellow control pane component https://github.com/openshift/cluster-openshift-apiserver-operator/blob/master/manifests/0000_30_openshift-apiserver-operator_07_deployment.yaml


adambkaplan (approved):
/lgtm

Looks like this is enough to get the operator up and running.

PR #350, opened 2020-12-22
Bug 1766287: Improve usability of hello-openshift imagestream

yselkowitz:
- Set local lookup policy on CVO manifests
- Update hello-openshift pullspec in CVO manifests

/assign @adambkaplan

### DIFF
--- a/manifests/08-openshift-imagestreams.yaml
+++ b/manifests/08-openshift-imagestreams.yaml
@@ -164,10 +164,12 @@ metadata:
     include.release.openshift.io/single-node-developer: "true"
     include.release.openshift.io/single-node-production-edge: "true"
 spec:
+  lookupPolicy:
+    local: true
   tags:
   - name: latest
     importPolicy:
       scheduled: true
     from:
       kind: DockerImage
-      name: docker.io/openshift/hello-openshift:latest
+      name: quay.io/openshift/origin-hello-openshift:latest
--- a/manifests/image-references
+++ b/manifests/image-references
@@ -61,4 +61,4 @@ spec:
   - name: hello-openshift
     from:
       kind: DockerImage
-      name: docker.io/openshift/hello-openshift:latest
+      name: quay.io/openshift/origin-hello-openshift:latest


manifests/08-openshift-imagestreams.yaml
```diff
@@ -145,6 +159,8 @@ metadata:
     include.release.openshift.io/single-node-developer: "true"
     include.release.openshift.io/single-node-production-edge: "true"
 spec:
+  lookupPolicy:
+    local: true
```
adambkaplan, line 163: openshift/oauth-proxy resolves to a Docker Hub image. There is a risk that an end-user is pulling the (obsolete) oauth-proxy image from Docker Hub today, and with this switch will start pulling the image from the internal registry.

https://hub.docker.com/r/openshift/oauth-proxy

manifests/08-openshift-imagestreams.yaml
```diff
@@ -164,10 +180,12 @@ metadata:
...
```
adambkaplan, line 184: Likewise, openshift/hello-openshift resolves to a Docker Hub image.

adambkaplan (commented):
/hold

@yselkowitz over Slack you asked, "Why don't we set `lookupPolicy: local: true` on everything?", and at the time I didn't have a good reason not to.

Now that the PR is in front of me, I have found a reason why we haven't set this field. Our convention for imagestream tag pull specs can overlap with actual Docker Hub images. In the case of `oath-proxy` and `hello-openshift` we would replace the docker hub image with the one from the payload on upgrade. To me, this is a breaking change.

Unfortunately this means that @cyril-ui-developer is out of luck here. His next best option is to add the `alpha.image.policy.openshift.io/resolve-names: '*'` annotation to the Deployment's pod template spec.

See https://docs.openshift.com/container-platform/4.6/openshift_images/using-imagestreams-with-kube-resources.html#images-managing-images-enabling-imagestreams-kube_using-imagestreams-with-kube-resources

manifests/08-openshift-imagestreams.yaml
yselkowitz: That probably would be a good thing, but let's leave that for further discussion in a separate PR.

manifests/08-openshift-imagestreams.yaml
yselkowitz: That is actually the entire point of the (new in 4.7) hello-openshift imagestream, to easily replace uses of the unmaintained, single-arch, and now rate-limited image on docker.io.

yselkowitz:
Repushed the PR limiting the changes to hello-openshift for now.

adambkaplan (approved):
/hold cancel

/approve

adambkaplan:
@yselkowitz please squash your commits.

yselkowitz:
Squashed as requested.

adambkaplan (approved):
/lgtm

Tested via cluster-bot and the dev console. Image replacement still works when the `alpha.image.policy.openshift.io/resolve-names: '*'` annotation is removed.

smarterclayton:
I asked for this image to be removed from the product in https://github.com/openshift/enhancements/pull/438 - I really don't want to be adding stuff around this.  Ingress should be using their operator image and a subcommand of their operator binary, not depending on a generic example image.

gabemontero:
> I asked for this image to be removed from the product in [openshift/enhancements#438](https://github.com/openshift/enhancements/pull/438) - I really don't want to be adding stuff around this. Ingress should be using their operator image and a subcommand of their operator binary, not depending on a generic example image.

OK @smarterclayton I've subscribed to that EP (I see @yselkowitz already responded to you there on current usage of this imagestream) and ack that I'm now officially aware of your wishes

My first thought is that I need to invesitigate / understand when the existing multi arch / console / serverless uses will be replaced so as not to break them.  Or do you want this imagestream removed now as a way to trigger their replacement?

PR #354, opened 2021-01-28
Update jenkins-rhel.json

yigitpolat:
v4.0 tag is no longer available on the Red Hat Registry. It can be replaced with any tag, but since other tags may also be deleted, I just simply wanted to use the latest tag. However, the pod has never skipped the readiness checks. So that, I tried v4.4.0 and worked perfectly.

### DIFF
--- a/assets/operator/ocp-s390x/jenkins/imagestreams/jenkins-rhel.json
+++ b/assets/operator/ocp-s390x/jenkins/imagestreams/jenkins-rhel.json
@@ -44,7 +44,7 @@
 				},
 				"from": {
 					"kind": "DockerImage",
-					"name": "registry.redhat.io/openshift4/ose-jenkins:v4.0"
+					"name": "registry.redhat.io/openshift4/ose-jenkins:v4.4.0"
 				},
 				"generation": null,
 				"importPolicy": {},


adambkaplan (commented):
@akram shouldn't this go through openshift/library?

/cc @gabemontero

gabemontero:
Yeah @yigitpolat that reference is not actually used if when we install the jenkins imagestreams.

Since Jenkins is part of the OCP payload, we by default overwrite that reference with the payload reference.  See https://github.com/openshift/cluster-samples-operator/blob/master/pkg/stub/jenkins.go

That said, if you would like to see that reference updated to a relevant image (say you decided to install that imagesteram in another namespace), the process for making that change is
- file a PR against https://github.com/openshift/jenkins, changing https://github.com/openshift/jenkins/blob/master/openshift/imagestreams/jenkins-rhel.yaml and https://github.com/openshift/jenkins/blob/master/openshift/imagestreams/jenkins-rhel.json
- once that merges, a backgound job will pull that into https://github.com/openshift/library
- then about once a release will pull https://github.com/openshift/library content into this repo

We generally do not edit the content in this repo with manual PRs like this.

/close

PR #357, opened 2021-02-19
Bug 1930570: Update Jenkins monitored templates names

waveywaves:


gabemontero:
@waveywaves could never wait for me to do my periodic sync from openshift/library via the `./library-script` :-)

Kidding aside ...

...his changes here are also in https://github.com/openshift/library/blob/master/operator/ocp-x86_64/official/jenkins/templates/jenkins-ephemeral-monitored.json so we won't lose these changes when I in fact do that sync.

the okd jobs are flaking on cluster bring up.  and the okd jenkins job passed, which is the one that covers these changes

I'll give the remaining okd jobs another chance or two and if they are still flaking on cluster bring up I'll skip them and tag for merge

/retest

waveywaves:
Hey @gabemontero :)
Sorry for that, we had a bug open for some time wrt incorrect naming of the templates so thought it necessary to open the PR manually. Thank you for reviewing this !

cc @jitendar-singh @akram @jkhelil

PR #364, opened 2021-03-31
Fix Typo

pcarney8:
fix typo: https://github.com/openshift/cluster-samples-operator/blob/master/manifests/010-prometheus-rules.yaml#L45


gabemontero:
/ok-to-test

/lgtm

thanks for the fix

PR #368, opened 2021-04-13
Bug 1948782: revert single-node-production-edge annotation

osherdp:
reverts #342


gabemontero:
most likely we can skip the okd e2e's @osherdp if they have failures ... I believe there are still a couple of known issues there

I'll do that after they all complete

But we'll have to retest  / examine any non-okd e2e failures

PR #374, opened 2021-05-13
Bug 1960334: manifests: fix selector in ServiceMonitor

vrutkovs:
Invalid selector in ServiceMonitor makes CVO attempt to apply this manifest on every sync iteration

### DIFF
--- a/manifests/06-servicemonitor.yaml
+++ b/manifests/06-servicemonitor.yaml
@@ -18,4 +18,5 @@ spec:
       caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
       serverName: metrics.openshift-cluster-samples-operator.svc
   selector:
-    name: cluster-samples-operator
+    matchLabels:
+      name: cluster-samples-operator


gabemontero:
good catch @vrutkovs thanks

For review I started to compare with other operators and am wondering if we need to change https://github.com/openshift/cluster-samples-operator/blob/master/manifests/06-metricsservice.yaml#L21 as well.

If you agree, can you update that file in the same way with this PR ?

The next step to consider I suppose would be to add an explicit label like https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/manifests/09_deployment.yaml#L6-L7 and then switch the selectors in the service/servicemonitor/deployment to https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/manifests/09_deployment.yaml#L14-L16 ... I'm having trouble finding where selector -> matchLabel works on the metadata.name in addition to metadata.labels.  Were you able to find that?  Or confirm it works off of the name?

thanks again

vrutkovs:
>am wondering if we need to change

Oh, good catch, updated the PR.

>I suppose would be to add an explicit label like ... I'm having trouble finding where selector -> matchLabel works on the metadata.name in addition to metadata.labels

`selector` can't work with `metadata.name`, just labels. I think using `name` is fine, altough a bit non-standard (usually folks use `app`)

gabemontero:
> > am wondering if we need to change
> 
> Oh, good catch, updated the PR.
> 
> > I suppose would be to add an explicit label like ... I'm having trouble finding where selector -> matchLabel works on the metadata.name in addition to metadata.labels
> 
> `selector` can't work with `metadata.name`, just labels. I think using `name` is fine, altough a bit non-standard (usually folks use `app`)

Yeah I saw the "usually folks use app" pattern when getting up to speed on this.

I'll defer to you since you don't think we need to change it right now, and I'm presuming once this merges, you'll be able to sort out things are fine just like when you found the issue with the old form (which I do see ref to btw in older versions of k8s doc).

gabemontero:
the e2e-aws failure was weird .... it looks like the samples operator was not launched

maybe has something to do with the change here ? .... if the deployment start up was attempted and failed, I'm not finding it in the must gather

going to /hold

and see how the other non-okd runs come in

xiuwang:
Can't install openshift-samples with the pr built image.
I will add must-gather logs to bugzilla directly later.
 $oc get co openshift-samples 
NAME                VERSION   AVAILABLE   PROGRESSING   DEGRADED   SINCE
openshift-samples

vrutkovs:
Oh, right, changing `manifests/06-metricsservice.yaml` was a mistake - its a Service, not ServiceMonitor

PR #376, opened 2021-05-18
seed relatedObjects in manifest yaml

gabemontero:
Per http://mailman-int.corp.redhat.com/archives/aos-devel/2021-May/msg00139.html from @wking 

Note:  the operator fills out `RelatedObjects` in https://github.com/openshift/cluster-samples-operator/blob/c1b73fe82b9bee65b28f886dcc930a5205680d6c/pkg/operatorstatus/operatorstatus.go#L226-L230


wking (approved):
/lgtm

PR #380, opened 2021-06-24
Bug 1975539: delete hello-openshift in payload imagestream via CVO annotation

gabemontero:
/assign @jottofar 

@jottofar please lgtm if I have interpreted https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/object-deletion.md and this bz correctly

I'm guessing I don't need to update `image-references` as well since this change is intended for deletion, but we'll see how the cluster install goes.  If there are complaints from the CVO around samples operator then that is the most likely explanation.

@dperaza4dustbit FYI ... these "in payload" imagestreams IIRC are also discussed in the skill xfer videos.

### DIFF
--- a/manifests/08-openshift-imagestreams.yaml
+++ b/manifests/08-openshift-imagestreams.yaml
@@ -144,3 +144,27 @@ spec:
       from:
         kind: DockerImage
         name: quay.io/openshift/origin-oauth-proxy-samples:v4.4
+---
+# we have to remove imagestreams here not via deletion, but per a CVO annotation
+# in order to properly conform with its conventions
+kind: ImageStream
+apiVersion: image.openshift.io/v1
+metadata:
+  namespace: openshift
+  name: hello-openshift
+  annotations:
+    release.openshift.io/delete: "true"
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    include.release.openshift.io/single-node-production-edge: "true"
+spec:
+  lookupPolicy:
+    local: true
+  tags:
+    - name: latest
+      importPolicy:
+        scheduled: true
+      from:
+        kind: DockerImage
+        name: quay.io/openshift/origin-hello-openshift:latest


gabemontero:
upgrade failure https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-samples-operator/380/pull-ci-openshift-cluster-samples-operator-master-e2e-aws-upgrade/1408065065768194048 appears unrelated to this change ... failed due to a required secret problem

gabemontero:
aws iam pain with https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-samples-operator/380/pull-ci-openshift-cluster-samples-operator-master-e2e-aws-image-ecosystem/1408065065663336448

gabemontero:
ok nothing yet related to this change with some passing e2e's as well

/retest

gabemontero:
> /lgtm

thanks @jottofar

gabemontero:
other operators besides samples are what are failing the upgrade jobs

gabemontero:
/hold

@jottofar - a question:  I pulled up the must-gather from one of the e2e build runs and I see the `hello-openshift` imagestream in the openshift namespace.

That is not the desired result - I was expecting to not see that, with the imagestream getting removed because of the new annotation.

Can you explain?

Here was the must gather I pulled:  https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-samples-operator/380/pull-ci-openshift-cluster-samples-operator-master-e2e-aws-operator/1408065065743028224/artifacts/e2e-aws-operator/gather-must-gather/artifacts/must-gather.tar

the imagesream:

```
- apiVersion: image.openshift.io/v1
  kind: ImageStream
  metadata:
    annotations:
      include.release.openshift.io/ibm-cloud-managed: "true"
      include.release.openshift.io/self-managed-high-availability: "true"
      openshift.io/image.dockerRepositoryCheck: "2021-06-24T14:37:44Z"
      release.openshift.io/delete: "true"
    creationTimestamp: "2021-06-24T14:37:44Z"
    generation: 2
    name: hello-openshift
    namespace: openshift
    resourceVersion: "15123"
    uid: 9ff32845-0222-4b18-96b9-d055fadbbc00
  spec:
    lookupPolicy:
      local: true
    tags:
    - annotations: null
      from:
        kind: DockerImage
        name: quay.io/openshift/origin-hello-openshift:latest
      generation: 2
      importPolicy:
        scheduled: true
      name: latest
      referencePolicy:
        type: Source
  status:
    dockerImageRepository: image-registry.openshift-image-registry.svc:5000/openshift/hello-openshift
    tags:
    - items:
      - created: "2021-06-24T14:37:44Z"
        dockerImageReference: quay.io/openshift/origin-hello-openshift@sha256:cc8f6ac26d8b9a1a027900d3f343b5e12dea2dc8ac1a5c64c1fd8c62ff2a385c
        generation: 2
        image: sha256:cc8f6ac26d8b9a1a027900d3f343b5e12dea2dc8ac1a5c64c1fd8c62ff2a385c
      tag: latest
```

jottofar:
>   I was expecting to not see that, with the imagestream getting removed because of the new annotation.
>
> Can you explain?    tag: latest
>
The deletes only happen during an upgrade. So you would need to be upgrading to a build that includes this PR. I can't remember if the `-upgrade` CI Jobs upgrade to one's PR build but when I checked their CVO logs I did not see the delete occurring. When it does occur searching the CVO log for `Delete` will turn up relevant logs.

gabemontero:
> > I was expecting to not see that, with the imagestream getting removed because of the new annotation.
> > Can you explain?    tag: latest
> 
> The deletes only happen during an upgrade. So you would need to be upgrading to a build that includes this PR. I can't remember if the `-upgrade` CI Jobs upgrade to one's PR build but when I checked their CVO logs I did not see the delete occurring. When it does occur searching the CVO log for `Delete` will turn up relevant logs.

I see .... well, then I think we'll have to abandon this approach then, since we don't want to re-introduce that imagestream just so we can delete it on upgrade.

I'll have to instead add some special case logic to clean up the hello-openshift imagestream to satisfy the report you generated.

I will add a comment in the imagestream manifest that if anything is to be deleted in the future, we go with this annotation ... though there is still the use case of a fresh install at a given level vs. an upgrade.

Interesting conundrum there.

jottofar:
> well, then I think we'll have to abandon this approach then, since we don't want to re-introduce that imagestream just so we can delete it on upgrade

No, you're good. CVO will see the Delete annotation and does not create/reconcile the manifest contents so it will not be created on install.
If you want to test it you could use cluster-bot and do a `--force` upgrade to your PR build.

gabemontero:
@jottofar and I have been chatting in slack and we have in fact found a CVO bug wrt imagestreams and the delete stuff

he'll be addressing that ... once that is in, I'll retest, make sure the hello world imagestream is in fact *NOT* created and will then unhold

gabemontero:
/test e2e-aws-operator

in case the e2e-aws-upgrade fails in such a way that I cannot check the openshift imagestreams to confirm hello-openshift is not present

thanks @jottofar

gabemontero:
ok hello-openshift was NOT installed in the https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-samples-operator/380/pull-ci-openshift-cluster-samples-operator-master-e2e-aws-operator/1409856204607328256 run

/hold cancel

/retest

PR #384, opened 2021-08-09
Bug 1990140: add connection with timeout in TBR accessibility check to expedite 'disconnected' mode

gabemontero:
/assign @dperaza4dustbit

### DIFF
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"os"
 	"runtime"
 	"strings"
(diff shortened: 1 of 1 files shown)


pkg/stub/handler.go
```diff
@@ -349,16 +350,27 @@ func (h *Handler) tbrInaccessible() bool {
...
+		connWithTimeout, err := net.DialTimeout("tcp", "registry.redhat.io:443", 15*time.Second)
+		if err != nil {
+			logrus.Infof("test connection with timeout failed with %s", err.Error())
+			return false, err
+		}
+		defer connWithTimeout.Close()
+		// still do the tls form of connect (which does not have the handy timeout form of dial) to confirm
+		// ssl handshake is OK
 		tlsConf := &tls.Config{}
 		conn, err := tls.Dial("tcp", "registry.redhat.io:443", tlsConf)
```
dmage, line 365: I think you can use `tls.Client(connWithTimeout, tlsConf)` (maybe with minor tweaks: https://cs.opensource.google/go/go/+/refs/tags/go1.16.7:src/crypto/tls/tls.go;l=154-164;drc=refs%2Ftags%2Fgo1.16.7)

gabemontero: yep I'll add that - thanks @dmage

gabemontero: after an iteration to addres ` either ServerName or InsecureSkipVerify must be specified in the tls.Config` I have it working

pushing update momentarily

gabemontero:
unrelated flake in e2e-aws that is already noted in sippy (fails ~20% of the time)

/test e2e-aws

pkg/stub/handler.go
```diff
@@ -349,16 +350,25 @@ func (h *Handler) tbrInaccessible() bool {
 		}
 	}
 	err = wait.PollImmediate(5*time.Second, 3*time.Minute, func() (bool, error) {
-		tlsConf := &tls.Config{}
-		conn, err := tls.Dial("tcp", "registry.redhat.io:443", tlsConf)
+		// we have seen cases in the field with disconnected cluster where the default connection timeout can be
+		// very long (15 minutes in one case); so we do an initial non-tls connection were we can specify a quicker
+		// timeout to filter out that scenario and default to tbr inaccessible / Removed in an expedient fashion
+		connWithTimeout, err := net.DialTimeout("tcp", "registry.redhat.io:443", 15*time.Second)
```
dperaza4dustbit, line 356: 15 still a best guest number that might change in the future. is this something we can make configurable so we can have the customer change if it does not work for them?

dperaza4dustbit (commented):
lgtm, just one question for me

pkg/stub/handler.go
gabemontero: good question @dperaza4dustbit 

config changes where we have to change the config object and add a new field and expensive / costly

admittedly a "gut feeling", but for what we are dealing with here, it would be better to just have a hard coded value that is sufficient

at most, maybe add an environment variable on the deployment that could be read

minimally, I would be agreeable to running say the e2e-aws-operator test suite repeated times (maybe a dozen) to get a warmer fuzzy 

/test e2e-aws-operator

gabemontero:
several operators below samples were degraded in e2e-aws-upgrade

/retest

pkg/stub/handler.go
dperaza4dustbit: OK understand @gabemontero , maybe take me through the process of making a config change to get a feeling on the price there.

gabemontero: yep let's add that to Thursday's call, along with the cross referencing CI flakes with sippy that I mentioned in your openshift/origin PR

Issue #385, opened 2021-08-18
cluster-samples-operator and stuck upgrade from 4.6 to 4.7 with restricted networks


toastbrotch:
Hi

we had set cluster-samples-operator to unmanaged and we have limited registries to our own (quay which ist not pull-thru-able) with allowedRegistriesForImport in the "images.config cluster" in our network restricted environment.

As we startet upgrade from 4.6.latest to 4.7.22. the whole upgrade got stuck with the creation of the is/hello-openshift. The only way we were able to finish upgrade was adding quay.io, registry.redhat.io and registry.access.redhat.com to the allowedRegistriesForImport and set cluster-samples-operator to Managed. i'm not sure if the steps described at https://github.com/openshift/cluster-samples-operator#troubleshooting helped us anything at all. nevertheless i had to remove the override in the clusterversion as otherwise openshift had always a warning hanging around (sorry forgot which).

to be honest this took us almost 1 day to figure out, which is way to much for just samples, we're not even interested. i hope the whole removal of the cluster-samples-operator would be possible in future... so its off and stays off without breaking my cluster.

regards,ivo

gabemontero:
Yeah across all the openshift operators, you should be using Removed instead of Unmanaged if you don't want that operator running.

Definitely something that has evolved since 4.1 when this stuff first arrived.

I can take a look at updating the README to better explain this.

gabemontero:
We also have a recent fix wrt delays in start up when in disconnected clusters.  See https://github.com/openshift/cluster-samples-operator/pull/384

Also, overriding is for development purposes only.  Having overrides violates your warranting for official OCP support.  I'll make sure the README makes that clear.

toastbrotch:
Why are samples a clusteroperator? Why not just another operator that is completely removable?

gabemontero:
Yeah unfortunately it was the only reliable means of delivering samples back in the 4.0/4.x days a few years ago.  Although not a given/mandatory as you noted, more of our customers at the time (and probably still the case) needed them, so this is how we landed.

There have been internal discussions in engineering as well as external requests from time to time to pull samples out of the payload.

I am told that among other things https://issues.redhat.com/browse/RFE-722 will be re-opened soon and decoupling samples from the OCP payload could be part of that.  It is also possible it will be achieved via separate RFEs as well.

@sbose78 is currently working on some internal proposals for this.

But it is not far enough along @toastbrotch that I can give you a definitive target date.

[...]

PR #394, opened 2021-09-08
Bug 2002368: acccount for image api returning invalid on imagestream create based on allowed/blocked registry settings

gabemontero:
Fixes https://github.com/openshift/cluster-samples-operator/issues/385

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,32 @@ Samples: `oc get is -n openshift`, `oc get templates -n openshift`  … use of -
 
 Deletion of the CRD instance will reset the samples operator to the default configuration, but leave the current revision of the samples operator pod running.
 
-If there is a bug in the samples operator deletion logic, to reset the samples operator configuration by stopping the current pod and starting a new one:
+You can also set the management state of the samples operator to `Removed` if you do not need samples running in your
+cluster:
+
+```yaml
+apiVersion: samples.operator.openshift.io/v1
+kind: Config
+metadata:
+  name: cluster
+spec:
+  architectures:
+  - x86_64
+  managementState: Removed
+```
+
+# "In-payload" imagestreams 
+
+The imagestreams defined at (https://github.com/openshift/cluster-samples-operator/blob/master/manifests/08-openshift-imagestreams.yaml)[https://github.com/openshift/cluster-samples-operator/blob/master/manifests/08-openshift-imagestreams.yaml]
+are not managed by the samples operator.  These are special imagestreams that point to images in the install payload, and their creation and updates are handled by the 
+[Cluster Version Operator](https://github.com/openshift/cluster-version-operator)
+
+# Development
+
+NOTE:  this next step violates your cluster support contract if you are an OCP customer.  But if your are using a cluster
+for development, and you wish to experiment with and update the samples operator code itself, you can reset the image used
+for the samples operator via the following steps:
+
 - Run `oc edit clusterversion version` and add an override entry to the spec for the Deployment of the samples operator so it is unmanaged:
 
 ```yaml
@@ -131,8 +156,8 @@ spec:
 ```
 
 - Scale down the deployment via `oc scale deploy cluster-samples-operator --replicas=0`
-- Edit the samples resource, remove the finalizer
-- Delete the samples resource, config map/secrets in operator namespace, samples in openshift namespace
+- Edit the samples resource, `oc edit config.samples cluster`, remove the finalizer
+- Delete the samples resource, `oc delete config.samples cluster`, config map/secrets in operator namespace, imagestreams and templates in openshift namespace
 - Scale up the deployment via `oc scale deploy cluster-samples-operator --replicas=1`
 
 
--- a/manifests/010-prometheus-rules.yaml
+++ b/manifests/010-prometheus-rules.yaml
@@ -86,10 +86,13 @@ spec:
         severity: info
       annotations:
         message: |
-          Samples operator could not access 'registry.redhat.io' during its initial installation and it bootstrapped as removed.
+          One of two situations has occurred.  Either
+          samples operator could not access 'registry.redhat.io' during its initial installation and it bootstrapped as removed.
           If this is expected, and stems from installing in a restricted network environment, please note that if you
           plan on mirroring images associated with sample imagestreams into a registry available in your restricted
           network environment, and subsequently moving samples operator back to 'Managed' state, a list of the images
           associated with each image stream tag from the samples catalog is
           provided in the 'imagestreamtag-to-image' config map in the 'openshift-cluster-samples-operator' namespace to
           assist the mirroring process.
+          Or, the use of allowed registries or blocked registries with global imagestream configuration will not allow
+          samples operator to create imagestreams using the default image registry 'registry.redhat.io'.
--- a/pkg/stub/config.go
+++ b/pkg/stub/config.go
@@ -75,6 +75,14 @@ func (h *Handler) SpecValidation(cfg *v1.Config) error {
 			}
 		}
 	}
+
+	// if the global imagestream config prevents the creation of imagestreams with the samples registry, we mark
+	// invalid config
+	if h.imageConfigBlocksImageStreamCreation(cfg.Spec.SamplesRegistry) {
+		err := fmt.Errorf("global openshift image configuration prevents the creation of imagestreams using the registry %s", cfg.Spec.SamplesRegistry)
+		return h.processError(cfg, v1.ConfigurationValid, corev1.ConditionFalse, err, "%#v")
+	}
+
 	h.GoodConditionUpdate(cfg, corev1.ConditionTrue, v1.ConfigurationValid)
 	return nil
 }
@@ -282,14 +290,17 @@ func (h *Handler) processError(opcfg *v1.Config, ctype v1.ConfigConditionType, c
 
 		// most reasons for degraded status are handled in ClusterOperatorStatusDegradedCondition based on
 		// state in the opcfg correlated with other factors; but when the error should directly drive the
-		// reason text, we set the reason here; at this time, that means either API Server errors or
-		// errors accessing the local file system
+		// reason text, we set the reason here; at this time, that means either API Server errors,
+		// errors accessing the local file system, or global imagestream config blocking the use of our configured
+		// registry
 		reason := kerrors.ReasonForError(err)
 		switch {
 		case len(reason) > 0 && reason != kapis.StatusReasonUnknown:
 			status.Reason = fmt.Sprintf("APIServer%sError", reason)
 		case cstatus == corev1.ConditionUnknown && ctype == v1.SamplesExist:
 			status.Reason = "FileSystemError"
+		default:
+			status.Reason = "GlobalImageStreamConfigRegistryMismatch"
 		}
 
 		util.ConditionUpdate(opcfg, status)
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -306,6 +306,16 @@ func IsRetryableAPIError(err error) bool {
 	return false
 }
 
+func ShouldNotGoDegraded(err error) bool {
+	// retryable and conflict errors should be self-explanatory for not going degraded;
+	// in particular with conflicts, they are quite possible with imagestreams during startup given
+	// the multi tag, highly concurrently nature of creating / updating;
+	// for invalid, we special case this because that is what the imagestream apiserver endpoint
+	// returns if an registry is missing from the images allowed registry list, or is explicitly
+	// listed in the blocked registry list
+	return IsRetryableAPIError(err) || kerrors.IsConflict(err) || kerrors.IsInvalid(err)
+}
+
 // this method assumes it is only called on initial cfg create, or if the Architecture array len == 0
 func (h *Handler) updateCfgArch(cfg *v1.Config) *v1.Config {
 	switch {
@@ -911,7 +921,7 @@ func (h *Handler) Handle(event util.Event) error {
 			}
 
 			if err != nil {
-				if !IsRetryableAPIError(err) && !kerrors.IsConflict(err) {
+				if !ShouldNotGoDegraded(err) {
 					h.processError(cfg, v1.SamplesExist, corev1.ConditionUnknown, err, "error creating samples: %v")
 					dbg := "setting samples exists to unknown"
 					logrus.Printf("CRDUPDATE %s", dbg)
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -3,6 +3,7 @@ package stub
 import (
 	"context"
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net"
@@ -25,6 +26,7 @@ import (
 	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/util/retry"
 
+	configv1 "github.com/openshift/api/config/v1"
 	imagev1 "github.com/openshift/api/image/v1"
 	operatorsv1api "github.com/openshift/api/operator/v1"
 	v1 "github.com/openshift/api/samples/v1"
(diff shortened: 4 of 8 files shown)


gabemontero:
/hold

I can get a bit more sophisticated / robust with this.

gabemontero:
aws flake

/test e2e-aws-upgrade

dperaza4dustbit (approved):
/lgtm

gabemontero:
yeah I'm going to pivot from our prior conversations @dperaza4dustbit 

I'm going to watch allowed/blocked registries, and if we are blocked, I'm going to bootstrap as removed, similar to what we do with ipv6/proxy/disconnected .... a "separate cluster config" situation makes samples install untenable.

I'll add a new reason code when we are removed, and I believe the existing alert we have there should be sufficient.

With that, I don't think there would be any future work for you team to do wrt new metrics and alerts.

This will require a doc update with @rolfedh (which I just mentioned in scrum with him).  We'll need to work with him here in this PR, perhaps with a Jira on his board, to track.

/assign @rolfedh

pkg/stub/imagestreams.go
```diff
@@ -175,7 +175,7 @@ func (h *Handler) upsertImageStream(imagestreamInOperatorImage, imagestreamInClu
 				// return the error so the caller can decide what to do
 				return err
 			}
-			if IsRetryableAPIError(err) || kerrors.IsConflict(err) {
+			if ShouldNotGoDegraded(err) {
 				logrus.Printf("CRDUPDATE: retryable error %s with imagestream update %s", err.Error(), imagestreamInCluster.Name)
```
stbenjam, line 179: This is a pre-existing problem but this line causes panics, imagestreamInCluster is guaranteed to be nil (line 170 checks)

https://bugzilla.redhat.com/show_bug.cgi?id=2003683

gabemontero: thanks for heads up @stbenjam  ... I've added a commit in this PR for now, but have also pinged @dperaza4dustbit about this 

we'll either address here with that commit , or he can do a separate PR, based on his taking ownership of samples (we are in the middle of the transition this week).

gabemontero:
/hold

the new global image config check needs some work

manifests/010-prometheus-rules.yaml
```diff
@@ -86,10 +86,13 @@ spec:
...
+          One of two situations has occurred.  Either
+          samples operator could not access 'registry.redhat.io' during its initial installation and it bootstrapped as removed.
           If this is expected, and stems from installing in a restricted network environment, please note that if you
           plan on mirroring images associated with sample imagestreams into a registry available in your restricted
           network environment, and subsequently moving samples operator back to 'Managed' state, a list of the images
           associated with each image stream tag from the samples catalog is
           provided in the 'imagestreamtag-to-image' config map in the 'openshift-cluster-samples-operator' namespace to
           assist the mirroring process.
+          Or, the use of allowed registries or blocked registries with global imagestream configuration will not allow
+          samples operator to create imagestreams using the default image registry 'registry.redhat.io'.
```
gabemontero, line 98: potential doc updates for these changes @dperaza4dustbit

pkg/stub/config.go
```diff
@@ -282,14 +290,17 @@ func (h *Handler) processError(opcfg *v1.Config, ctype v1.ConfigConditionType, c
...
+		// errors accessing the local file system, or global imagestream config blocking the use of our configured
+		// registry
 		reason := kerrors.ReasonForError(err)
 		switch {
 		case len(reason) > 0 && reason != kapis.StatusReasonUnknown:
 			status.Reason = fmt.Sprintf("APIServer%sError", reason)
 		case cstatus == corev1.ConditionUnknown && ctype == v1.SamplesExist:
 			status.Reason = "FileSystemError"
+		default:
+			status.Reason = "GlobalImageStreamConfigRegistryMismatch"
```
gabemontero, line 303: A doc update around this new reason for our config invalid condition is probably warranted @dperaza4dustbit

pkg/stub/handler.go
```diff
@@ -348,16 +417,28 @@ func (h *Handler) tbrInaccessible() bool {
...
-	proxy, err := h.configclient.Proxies().Get(context.TODO(), "cluster", metav1.GetOptions{})
-	if err != nil {
-		logrus.Printf("unable to retrieve proxy configuration as part of testing registry.redhat.io connectivity: %s", err.Error())
-	} else {
-		if len(proxy.Status.HTTPSProxy) > 0 || len(proxy.Status.HTTPProxy) > 0 {
-			logrus.Printf("with global proxy configured assuming registry.redhat.io is accessible, bootstrap to Managed")
-			return false
+	if h.imageConfigBlocksImageStreamCreation("registry.redhat.io") ||
+		h.imageConfigBlocksImageStreamCreation("registry.access.redhat.io") ||
+		h.imageConfigBlocksImageStreamCreation("quay.io") {
```
gabemontero, line 422: doc update around this @dperaza4dustbit

gabemontero:
just the okd upgrade ... non related failure

/skip

coreydaley:
@gabemontero I've got this in my queue to look at in the morning.

README.md
```diff
@@ -117,7 +117,32 @@ Samples: `oc get is -n openshift`, `oc get templates -n openshift`  … use of -
...
+spec:
+  architectures:
+  - x86_64
+  managementState: Removed
+```
+
+# "In-payload" imagestreams 
+
+The imagestreams defined at (https://github.com/openshift/cluster-samples-operator/blob/master/manifests/08-openshift-imagestreams.yaml)[https://github.com/openshift/cluster-samples-operator/blob/master/manifests/08-openshift-imagestreams.yaml]
+are not managed by the samples operator.  These are special imagestreams that point to images in the insatll payload, and their creation and updates are handled by the
```
coreydaley, line 137:
```suggestion
are not managed by the samples operator.  These are special imagestreams that point to images in the install payload, and their creation and updates are handled by the 
```

README.md
```diff
@@ -131,8 +156,8 @@ spec:
 ```
 
 - Scale down the deployment via `oc scale deploy cluster-samples-operator --replicas=0`
-- Edit the samples resource, remove the finalizer
-- Delete the samples resource, config map/secrets in operator namespace, samples in openshift namespace
+- Edit the samples resource, `oc edit config.samples cluster`, remove the finalizer
+- Delete the samples resource, `oc deletet config.samples cluster`, config map/secrets in operator namespace, imagestreams and templates in openshift namespace
```
coreydaley, line 160:
```suggestion
- Delete the samples resource, `oc delete config.samples cluster`, config map/secrets in operator namespace, imagestreams and templates in openshift namespace
```

pkg/stub/handler.go
```diff
@@ -306,6 +306,16 @@ func IsRetryableAPIError(err error) bool {
 	return false
 }
 
+func ShouldNotGoDegraded(err error) bool {
+	// retryable and conflict errors should be self-explanatory for not going degraded;
+	// in particular with conflicts, they are quite possible with imagestreams during startup given
+	// the multi tag, highly concurrently nature of creating / updating;
```
coreydaley, line 312:
```suggestion
	// the multi tag, highly concurrent nature of creating / updating;
```

pkg/stub/handler.go
```diff
@@ -306,6 +306,16 @@ func IsRetryableAPIError(err error) bool {
...
+	// for invalid, we special case this because that is what the imagestream apiserver endpoint
+	// returns if an registry is missing from the images allowed registry list, or is explicitly
```
coreydaley, line 314:
```suggestion
	// returns if a registry is missing from the images allowed registry list, or is explicitly
```

pkg/stub/handler.go
```diff
@@ -327,6 +337,75 @@ func (h *Handler) updateCfgArch(cfg *v1.Config) *v1.Config {
...
+			logrus.Printf("unable to retrieve image configuration as part of testing registry.redhat.io connectivity: %s", err.Error())
+			return false, nil
+		} else {
+			// check allowed domain list
+			if len(imgCfg.Spec.AllowedRegistriesForImport) > 0 {
+				for _, rl := range imgCfg.Spec.AllowedRegistriesForImport {
+					logrus.Printf("considering allowed registries domain %s for %s", rl.DomainName, name)
+					if strings.HasSuffix(name, rl.DomainName) {
+						msg := fmt.Sprintf("the allowed registries domain %s allows imagestream creation access for search param %s", rl.DomainName, name)
+						logrus.Printf(msg)
```
coreydaley, line 361:
```suggestion
						logrus.Printf(("the allowed registries domain %s allows imagestream creation access for search param %s", rl.DomainName, name)
```

coreydaley: This could be updated throughout the pull request. There shouldn't be a reason to create the `msg` variable and then pass it into `Printf` or `Errorf` unless it is going to be used again somewhere else.

pkg/stub/handler.go
```diff
@@ -338,16 +417,28 @@ func (h *Handler) tbrInaccessible() bool {
...
+		if err != nil {
+			logrus.Printf("unable to retrieve proxy configuration as part of testing registry.redhat.io connectivity: %s", err.Error())
+			return false, nil
+		} else {
+			if len(proxy.Status.HTTPSProxy) > 0 || len(proxy.Status.HTTPProxy) > 0 {
+				logrus.Printf("with global proxy configured assuming registry.redhat.io is accessible, bootstrap to Managed")
+				return true, nil
+			}
 		}
+		return false, fmt.Errorf("proxy is not configured")
```
coreydaley, line 438:
```suggestion
		return false, errors.New("proxy is not configured")
```

pkg/stub/handler.go
```diff
@@ -327,6 +337,75 @@ func (h *Handler) updateCfgArch(cfg *v1.Config) *v1.Config {
...
+	// unit tests.  In the actual runtime if we cannot create the client the operator errors out on startup.
+	if h.configclient == nil {
+		return false
+	}
+	err := wait.PollImmediate(5*time.Second, 20*time.Second, func() (done bool, err error) {
+		// if the image config allowed registry or blocked registry list will prevent the creation of imagestreams,
+		// we consider this inaccessible
+		imgCfg, err := h.configclient.Images().Get(context.TODO(), "cluster", metav1.GetOptions{})
+		if err != nil {
+			logrus.Printf("unable to retrieve image configuration as part of testing registry.redhat.io connectivity: %s", err.Error())
```
coreydaley, line 352: should `registry.redhat.io` be hard coded here? Or should it use the name of the registry that is passed in as `name`?

pkg/stub/handler.go
```diff
@@ -327,6 +337,75 @@ func (h *Handler) updateCfgArch(cfg *v1.Config) *v1.Config {
...
+			return false, nil
+		} else {
```
coreydaley, line 354: I don't think you need an else block here

pkg/stub/handler.go
```diff
@@ -327,6 +337,75 @@ func (h *Handler) updateCfgArch(cfg *v1.Config) *v1.Config {
...
+						msg := fmt.Sprintf("the allowed registry %s allows imagestream creation for search param %s", r, name)
+						logrus.Printf(msg)
+						return true, nil
+					}
+				}
+				msg := fmt.Sprintf("no allowed registries items will permit the use of %s", name)
+				return false, fmt.Errorf(msg)
+
+			}
+
```
coreydaley, line 383: Do we need a section to check for a partial match of domain name for blocked registries also, like we do above for allowed registries?

pkg/stub/handler.go
```diff
@@ -327,6 +337,75 @@ func (h *Handler) updateCfgArch(cfg *v1.Config) *v1.Config {
...
+	if h.configclient == nil {
+		return false
+	}
+	err := wait.PollImmediate(5*time.Second, 20*time.Second, func() (done bool, err error) {
+		// if the image config allowed registry or blocked registry list will prevent the creation of imagestreams,
+		// we consider this inaccessible
+		imgCfg, err := h.configclient.Images().Get(context.TODO(), "cluster", metav1.GetOptions{})
+		if err != nil {
+			logrus.Printf("unable to retrieve image configuration as part of testing registry.redhat.io connectivity: %s", err.Error())
+			return false, nil
```
coreydaley, line 353: Would it make sense to just have this section in the `wait.PollImmediate` instead of the whole logic for this function?

pkg/stub/handler.go
```diff
@@ -327,6 +337,75 @@ func (h *Handler) updateCfgArch(cfg *v1.Config) *v1.Config {
 	return cfg
 }
 
+func (h *Handler) imageConfigBlocksImageStreamCreation(name string) bool {
```
coreydaley, line 340: Is it documented anywhere what order these rules are processed in? Does being specifically allowed override being specifically denied? Does a full match of an allowed registry override a partial match of a blocked registry, etc?

coreydaley (changes requested):

pkg/stub/handler.go
gabemontero: yeah there were some cases where I was both creating a msg for a log as well as for an error to return

in those cases I'll try to address

in the cases where both are in fact occurring, I'll keep the pattern of defining the string once

pkg/stub/handler.go
gabemontero: where is `errors` suppose to come from ? ... my IDE package searches are not turning anything up

pkg/stub/handler.go
gabemontero: yep

pkg/stub/handler.go
gabemontero: yep

pkg/stub/handler.go
coreydaley: https://pkg.go.dev/errors

coreydaley: Part of the Go standard library

pkg/stub/handler.go
coreydaley: > yep

Yep to which one? :)

pkg/stub/handler.go
gabemontero: nope the domain thing only exists for allowed

then it is explicit registry listing for the stuff under RegistrySource

pkg/stub/handler.go
gabemontero: sorry not following / can't parse

coreydaley: Would it make sense for only the Get() to be inside of the wait.PollImmediate, instead of all the code within the function being in there? Probably doesn't really matter honestly, just a thought.

pkg/stub/handler.go
gabemontero: order does not matter .... a problem with any of them, we punt

even if they did something like putting something in both allowed and blocked ... too bad, we punt until you fix your config so it makes sense

in any event, nuances like that should be handled by the owners of the config api itself ... they should have validation to catch that 

ultimately the doc updates I'll work out with @rolfedh will get into that

gabemontero:
updated I did make are in separate commits for now @coreydaley .... I'll squash after you sign off on those, plus we sort out remaining questions

also, ideally we do not merge until I hear again from @dperaza4dustbit on these more recent changes ... he lgtmed an earlier version

pkg/stub/handler.go
gabemontero: switch to param name

gabemontero:
don't know why I couldn't find "errors" before ... switch will be up momentarily @coreydaley

pkg/stub/handler.go
gabemontero: ah ... you are right that is better .... update up in a couple

gabemontero:
round 2 updates up @coreydaley

gabemontero:
known unit flake with okd version .. all other okd's passed

/skip

/retest

pkg/stub/config.go
dperaza4dustbit: got it

pkg/stub/handler.go
dperaza4dustbit: got it

gabemontero:
great thanks @coreydaley @dperaza4dustbit 

I've squashed the review commits from @coreydaley 's reviews

if I could get one more lgtm from one of you, we'll go from there

about to ping @rolfedh on slack now to get the doc updates going

gabemontero:
unrelated flakes in okd build test failure; sig-builds passed

/skip

gabemontero:
unrelated sig-instrumentation flake image eco

/test e2e-aws-image-ecosystem

gabemontero:
when it is time to cherrypick, the 4.9 and other z streams will need both this PR and https://github.com/openshift/cluster-samples-operator/pull/397

gabemontero:
/assign @jitendar-singh 

ideally you apply the qe approved label to this PR

/assign @rolfedh 

ideally you apply the docs approved label to this PR, where the input I gave you at https://docs.google.com/document/d/1fN_-2_ZPC6DQBIYJkrlqL_0i9GSVbqJ8YyhB8hRns-o/edit is the starting point you need

PR #397, opened 2021-09-22
Bug 2006947: fix proxy portion of tbr inaccessible check

gabemontero:


gabemontero:
/hold 

until e2e-aws-proxy is available

gabemontero:
hard to tell ... probably a mysql flake in image eco give it passed earlier today and I see no recent commits there or in cakephp

non related flakes in okd

/skip

gabemontero:
and we have a e2e-aws-proxy up and running

gabemontero:
aws fail on image ecosystem

gabemontero:
ok the e2e-aws-proxy failure was an unrelated sig-arch failure @adambkaplan @coreydaley 

all the sig-builds tests passed

samples come up as managed, no image import failures, and the pod logs indicated the expected behavior

this one is ready for any final review comments and tag for merge

gabemontero:
I'll

/retest

one more time to see if we get a total green, no unrelated flakes e2e-aws-proxy; otherwise I'll skip to facilitate merge

gabemontero:
low level install fail

/retest

gabemontero:
/assign @dperaza4dustbit 

fyi - my last fix prior to ownership transfer broke OCP with proxy enabled .... this fixes it (the e2e-aws-proxy failures have been unrelated to this bug/fix, and I was able to verify in the must gather artifacts that samples was bootstrapping again as managed with proxy enabled as it should, where as the bug I introduced resulted in samples bootstrapping as removed).

pkg/stub/handler.go
```diff
@@ -430,24 +430,25 @@ func (h *Handler) tbrInaccessible() bool {
...
-				logrus.Printf("with global proxy configured assuming registry.redhat.io is accessible, bootstrap to Managed")
-				return true, nil
-			}
 		}
-		return false, errors.New("proxy is not configured")
+		return true, nil
 	})
-	if err != nil {
+
+	if len(proxy.Status.HTTPSProxy) > 0 || len(proxy.Status.HTTPProxy) > 0 {
```
coreydaley, line 447: Is there any chance that `proxy.Status` could be nil here? Or are we confident that `h.configclient.Proxies.Get` will provide that even if there is an error?

pkg/stub/handler.go
```diff
@@ -430,24 +430,25 @@ func (h *Handler) tbrInaccessible() bool {
 		h.imageConfigBlocksImageStreamCreation("quay.io") {
 		return true
 	}
-	err := wait.PollImmediate(5*time.Second, 20*time.Second, func() (bool, error) {
+
+	proxy := &configv1.Proxy{}
+	var err error
+	wait.PollImmediate(5*time.Second, 20*time.Second, func() (bool, error) {
```
coreydaley, line 436: I think it would be worth at least logging the timeout error from this showing that it never completed.

pkg/stub/handler.go
gabemontero: nope it is not a pointer ... no nil ref issues here

pkg/stub/handler.go
```diff
@@ -430,24 +430,25 @@ func (h *Handler) tbrInaccessible() bool {
 		h.imageConfigBlocksImageStreamCreation("quay.io") {
 		return true
 	}
-	err := wait.PollImmediate(5*time.Second, 20*time.Second, func() (bool, error) {
+
+	proxy := &configv1.Proxy{}
+	var err error
```
coreydaley, line 435: nit: The error from inside of the `wait.pollImmediate` is never surfaced outside of that function, so this probably isn't needed here (but is for the second one below)

gabemontero: actually no with what I am doing with proxy I either need to do this as I have it, or I have to rename the var

leaving this as is

PR #399, opened 2021-09-30
Bug 2007757: Including Template in must-gather related objects.

dperaza4dustbit:
The must-gather tool is collecting imagestreams under the openshift
namespace but it is not collecting templates. Being more prescriptive
on the groups we want to collect for sample related objects.


gabemontero:
/lgtm

/hold 

I'm holding @dperaza4dustbit so that you can inspect a must gather from one of the completed e2e like we did today during scrum and confirm that both the templates and imagestreams from the openshift namespace have been collected

Once you posted the successful results here, you can cancel the hold and we can let this one merge

thanks!!

PR #410, opened 2022-01-21
upd rhel7 Dockerfile in case still used for rhel7 worker nodes

gabemontero:


gabemontero:
/label docs-approved
/label qa-approved 
/label px-approved

given this is quite possibly obsolete

PR #413, opened 2022-01-31
WIP: Sync rails templates to ruby:2.7-ubi8

yselkowitz:
This is to test https://github.com/sclorg/rails-ex/pull/141 with image-eco.


yselkowitz:
this verified that the rails-ex changes are safe, but unfortunately they landed too late for code freeze.  this will be included in the 4.11 sync.
/close

PR #436, opened 2022-07-18
Bug 2107521: Remove EAP deprecated images from Samples Operator OCP release 4.10

fbm3307:
Remove EAP deprecated images from Samples Operator OCP release 4.10

Signed-off-by: Feny Mehta <fbm3307@gmail.com>


jmesnil:
@fbm3307 theses 3 images for EAP are all deprecated and no longer maintained.

Could we simply delete the imagestream files? If we do so, would we remove them from 4.10 upgrades?
Otherwise, should we keep the imagestream and remove *all* tags (including `latest`)?

fbm3307:
> @fbm3307 theses 3 images for EAP are all deprecated and no longer maintained.
> 
> Could we simply delete the imagestream files? If we do so, would we remove them from 4.10 upgrades? Otherwise, should we keep the imagestream and remove _all_ tags (including `latest`)?

@jmesnil these are not maintained for 4.10.x also.? if not maintained, i can modify this PR to delete these jsons.?

jmesnil:
> > @fbm3307 theses 3 images for EAP are all deprecated and no longer maintained.
> > Could we simply delete the imagestream files? If we do so, would we remove them from 4.10 upgrades? Otherwise, should we keep the imagestream and remove _all_ tags (including `latest`)?
> 
> @jmesnil these are not maintained for 4.10.x also.? if not maintained, i can modify this PR to delete these jsons.?

Actually, I think your PR is correct. Let's remove the oldest non-v2 images only. This will be less disruptive for the users of older OCP releases as these imagestreams with the most recent images will still be present.
Finding a way to notify users that the images are unsupported is out of scope of this issue.

thanks!

fbm3307:
The 3 test not passing are for OKD.
/lgtm

fbm3307:
Closing this as not required anymore

PR #459, opened 2022-07-25
Bug 2110284: Remove jboss-datagrid deprecated images from Samples Op…

pminz:
…erator OCP release 4.7


fbm3307:
/ok-to-test

@pminz can you please change the title to reference the bug like this " Bug 2110284: Remove..."

PR #495, opened 2023-03-22
OCPBUGS-10910: Add network tools imagestreams

Prashanth684:
The network tools image is used by certain tests in origin and the way that is done today [1] is by spinning up a pod with the CVO image and getting the network-tools image spec from inside the pod. This is not desirable and also doesn't work today when these tests run on a cluster with multi arch compute nodes.

This PR adds the network tools imagestreams so these tests can be changed to use the imagestream rather than inferring it through creating a pod today.

[1] https://github.com/openshift/origin/blob/master/test/extended/util/release.go#L32

### DIFF
--- a/manifests/08-openshift-imagestreams.yaml
+++ b/manifests/08-openshift-imagestreams.yaml
@@ -167,3 +167,21 @@ spec:
       from:
         kind: DockerImage
         name: quay.io/openshift/origin-hello-openshift:latest
+---
+kind: ImageStream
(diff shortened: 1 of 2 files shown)


npinaeva:
I can't verify if that is the correct implementation, but we definitely need to add network tools to imagestreams, so LGTM from the network-tools side

Prashanth684:
@fbm3307 could you please drop an approval on this?

PR #497, opened 2023-04-17
Add UBI ImageStreams

jackevans43:
Useful for:
* Injecting compiled Go/Rust binary into from a multi-stage build
* Generic images for installing 3rd party RPMs into, bash scripts etc


coreydaley:
/close
CSO is being downsized and no new items are being added.

PR #510, opened 2023-07-14
 reconcile status when clusteroperator changes

sanchezl:
Needed for https://github.com/openshift/cluster-kube-apiserver-operator/pull/1423.

If status is modified, the operator must set status to proper values. This ensures that accidents cannot permanently reset status and gives a clear indication that the operator is "live". This came up as important when operators were NOT live, during cert rotations and we had no indication of problems.

PR #512, opened 2023-07-14
add handling to reconcile when clusteroperator status is changed

deads2k:
found while investigating https://github.com/openshift/cluster-kube-apiserver-operator/pull/1423

this operator needs to keep its status up to date with reality when something messes with status.


deads2k:
closing in favor of #510

PR #516, opened 2023-08-07
Migrate from DeploymentConfig to Deployment

ardaguclu:
This operator suggests application examples in DeploymentConfig format. On the other hand, `DeploymentConfig` is deprecated and using `Deployment` is suggested instead.

This PR migrates from `DeploymentConfig` to `Deployment`. Since Deployment does not support `Triggers` field(thus, this functionality is also deprecated), this PR also removes this field from Deployment resources.

### DIFF
--- a/assets/operator/ocp-aarch64/cakephp/templates/cakephp-mysql-example.json
+++ b/assets/operator/ocp-aarch64/cakephp/templates/cakephp-mysql-example.json
@@ -139,8 +139,8 @@
 			}
 		},
 		{
-			"apiVersion": "apps.openshift.io/v1",
-			"kind": "DeploymentConfig",
+			"apiVersion": "apps/v1",
+			"kind": "Deployment",
 			"metadata": {
 				"annotations": {
 					"description": "Defines how to deploy the application server",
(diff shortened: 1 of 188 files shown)


ardaguclu:
/cc @ingvagabund 
/hold
until @ingvagabund review this PR.

fbm3307:
Hi @ardaguclu , these changes needs to take place at each imagestream/template level, the  assets/operator folder here gets overwritten/updated everytime i sync the library/official.yaml.(whateverchanges are there at each imagestream/template repo are pulled from there) you should be able to get the repo address of each imagestream/template from https://github.com/openshift/library/blob/master/official.yaml 
cc @dperaza4dustbit

ardaguclu:
> Hi @ardaguclu , these changes needs to take place at each imagestream/template level, the assets/operator folder here gets overwritten/updated everytime i sync the library/official.yaml.(whateverchanges are there at each imagestream/template repo are pulled from there) you should be able to get the repo address of each imagestream/template from https://github.com/openshift/library/blob/master/official.yaml cc @dperaza4dustbit

@fbm3307 thanks a lot for pointing this out. Could you please give me an example of where should I change then?. If I understood correctly, there will be no change in cluster-samples-operator and I'll close this PR?

fbm3307:
@ardaguclu 
https://github.com/openshift/library/blob/master/official.yaml  here 
for example 
ruby:
    imagestreams:
      - location: https://raw.githubusercontent.com/sclorg/s2i-ruby-container/master/imagestreams/ruby-rhel.json
This will give you the repo where you need to make changes.
this has to be done for each imagestream/template

PR #531, opened 2024-01-11
OCPBUGS-26992: Align builder and base for Dockerfile.okd with Dockerfile

sherine-k:
Hi team, 
OKD-SCOS for release 4.16 isn't building. Please help us update what is needed. 
Thank you


fbm3307:
Please add a valid jira bug

fbm3307:
@sherine-k , Samples Operator is in downsizing phase, and i would recommend, to remove or deviate any dependency OKD has from Samples Operator . Going forward we would not be able to update or do releases via Samples Operator

PR #537, opened 2024-03-27
SO-121: Re-sync Samples Library for OCP 4.16

fbm3307:
This for Release Activity of Samples Operator for OCP 4.16


fbm3307:
/skip 
As the test failing is for deployment config which is now deprecated

PR #541, opened 2024-04-18
WRKLDS-1015: tolerate node-role.kubernetes.io/control-plane:NoExecute 

knelasevero:
Required to be able to advise customers to follow workflow described in https://github.com/openshift/enhancements/pull/1583. If they want to taint control-plane nodes to avoid any other thing to run there (while enforcing it further with validating admission policies), control plane workloads need to tolerate that taint.


knelasevero:
Is there any other place where I would probably need to tolerate the taint in this repo?

fbm3307:
@knelasevero Samples operator is going to be removed After OCP 4.17, we are already into maintenance mode .
SO i would not recommend adding anything new to samples operator now

PR #544, opened 2024-04-23
AGENT-863: add a new image stream manifest for the fips capable installer

andfasano:
This patch is required to retrieve the baremetal installer image by the day2 add node script (see https://github.com/openshift/installer/pull/8242)

### DIFF
--- a/manifests/08-openshift-imagestreams.yaml
+++ b/manifests/08-openshift-imagestreams.yaml
@@ -59,6 +59,24 @@ kind: ImageStream
 apiVersion: image.openshift.io/v1
 metadata:
   namespace: openshift
+  name: installer-fips
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+spec:
+  tags:
+  - name: latest
+    importPolicy:
+      scheduled: true
+      importMode: PreserveOriginal
+    from:
+      kind: DockerImage
+      name: quay.io/openshift/origin-baremetal-installer:4.16
+---
+kind: ImageStream
+apiVersion: image.openshift.io/v1
+metadata:
+  namespace: openshift
   name: installer-artifacts
   annotations:
     include.release.openshift.io/ibm-cloud-managed: "true"


andfasano:
cc @zaneb

fbm3307:
/hold
looks like the imagestream is still not available ? https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-samples-operator/544/pull-ci-openshift-cluster-samples-operator-master-e2e-aws-ovn-image-ecosystem/1782678785934495744

zaneb:
Hmm, I can see it even without being logged in: https://quay.io/repository/openshift/origin-baremetal-installer

And I can pull it no problem:

```
$ podman pull quay.io/openshift/origin-baremetal-installer
Trying to pull quay.io/openshift/origin-baremetal-installer:latest...
Getting image source signatures
Copying blob e6564833bd0b done   | 
Copying blob d8190195889e done   | 
Copying blob 9444bc135ccc done   | 
Copying blob 97da74cc6d8f done   | 
Copying blob b7caac98184d done   | 
Copying blob d3540128d21d done   | 
Copying config c00a0c0cf0 done   | 
Writing manifest to image destination
c00a0c0cf0837eeb50de2a425fa84c918b20d10b65b7ac6673154c0f420b32fa
```

So it's not one of those private repos where nobody can remember if there is a reason for it to be private :sweat_smile: 

I think the issue must be that there is no 4.0 [tag](https://quay.io/repository/openshift/origin-baremetal-installer?tab=tags) in the repo, because the first version of OCP to contain this image was 4.2.

Presumably this all gets overwritten by the CVO anyway? @fbm3307 what would you recommend as the tag to use? `4.2` (first version that exists) or e.g. `4.16` (first version that will have the ImageStream) or something else?

fbm3307:
i wont be the right person to answer it, i do not have much idea onto the tags.

andfasano:
Let's go with the latest

zaneb:
All passing except for the optional OKD jobs that have been permafailing for a while now.
/lgtm

fbm3307:
Just FYI @zaneb @andfasano , Samples operator is going to be removed After OCP 4.17, we are already into maintenance mode .

andfasano:
Closed as per comment https://github.com/openshift/cluster-samples-operator/pull/544#issuecomment-2074406605

PR #548, opened 2024-07-23
[release-4.15] OCPBUGS-37464: Set required-scc for openshift workloads

kramaranya:
[OCPBUGS-37464](https://issues.redhat.com/browse/OCPBUGS-37464)

Cherry-pick of #535

### DIFF
--- a/manifests/06-operator-ibm-cloud-managed.yaml
+++ b/manifests/06-operator-ibm-cloud-managed.yaml
@@ -15,6 +15,7 @@ spec:
   template:
     metadata:
       annotations:
+        openshift.io/required-scc: restricted-v2
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
       labels:
         name: cluster-samples-operator
--- a/manifests/06-operator.yaml
+++ b/manifests/06-operator.yaml
@@ -15,6 +15,7 @@ spec:
     metadata:
       annotations:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
+        openshift.io/required-scc: restricted-v2
       labels:
         name: cluster-samples-operator
     spec:


gabemontero:
@fbm3307 @dperaza4dustbit - apologies for not finding it, but who should be approving these PRs now?  

thanks

fbm3307:
@metlos @mfrancisc should be able to help you here

mfrancisc:
@kramaranya I'm new to this project, my understanding is that this is backport fix from 4.16 correct ?

kramaranya:
@mfrancisc yep, you're right!

kramaranya:
> lgtm

@mfrancisc thanks! however the label wasn't added, could you please write it with a slash?

mfrancisc:
> @mfrancisc thanks! however the label wasn't added, could you please write it with a slash?

@kramaranya I would be more then happy to do that, but I believe I need to be in the owners file , otherwise my approval doesn't matter :) See https://github.com/openshift/cluster-samples-operator/pull/549

Meanwhile maybe @fbm3307 can help with this 🙏

kramaranya:
> I would be more then happy to do that, but I believe I need to be in the owners file

@mfrancisc oh makes sense, i guess now you can label it as that pr was merged :) thanks!

mfrancisc:
I think I need to be owner also in the release-4.15 branch 🤦‍♂️  . See https://github.com/openshift/cluster-samples-operator/pull/550

metlos:
@kramaranya, @fbm3307 is anyone of you able to add the above labels? seems like that's above my paygrade...

kramaranya:
> kramaranya, fbm3307 is anyone of you able to add the above labels? seems like that's above my paygrade...

https://github.com/openshift/release/blob/master/core-services/prow/02_config/openshift/cluster-samples-operator/_pluginconfig.yaml#L9 I guess @fbm3307 is able to add both labels

PR #550, opened 2024-07-26
OCPBUGS-38157: add mfrancisc and metlos as owners add owners 4.15

mfrancisc:


fbm3307:
@mfrancisc you would need to add @metlos in openshift org

mfrancisc:
> @mfrancisc you would need to add @metlos in openshift org

I don't think I have permissions to do that. I think he needs to send request for that! 

@metlos could you please send request for taking part of the org ?

metlos:
@mfrancisc I'm now a member of the openshift org.

PR #552, opened 2024-07-30
Update kubernetes API and go version

metlos:
Update the go version and kubernetes API version to the same as in openshift/api.


metlos:
Done in a different PR.

PR #553, opened 2024-07-31
NO-ISSUE: Only pick up the changes from the supported imagestreams and templates in library-sync.sh.

metlos:
This updates the `library-sync.sh` script to only update the imagestreams and templates of the supported samples.

This should ease the release process so that we (or any future maintainer) doesn't have to remember to manually go through the changes brought in from openshift and remove any changes to the unsupported samples.

### DIFF
--- a/library-sync.sh
+++ b/library-sync.sh
@@ -53,3 +53,23 @@ git add operator
 rm t.tar
 rm -rf library-master
 
+SUPPORTED="ruby python nodejs perl php httpd nginx eap java webserver dotnet golang"
+function reset_unsupported() {
+  for d in $(ls); do
+    if [[ "${SUPPORTED}" != *"${d}"* ]]; then
+      git reset HEAD -- "${d}"
+      # remove any changes from the working tree in this directory that reset left behind
+      git stash -u -- "${d}"
+      git stash drop
+    fi
+  done
+}
+
+ARCHS="ocp-x86_64 ocp-aarch64 ocp-ppc64le ocp-s390x okd-x86_64"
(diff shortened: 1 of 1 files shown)


mfrancisc (approved):

jupierce:
I don't see any need for docs/px in this PR, and a request was made to land this. So adding required labels.

jupierce:
Overriding old okd jobs:
/override ci/prow/okd-e2e-aws-ovn-builds
/override ci/prow/okd-e2e-aws-ovn-image-ecosystem
/override ci/prow/okd-e2e-aws-ovn-upgrade
/override ci/prow/okd-e2e-aws-operator
/override ci/prow/okd-e2e-aws-ovn
/override ci/prow/okd-e2e-aws-ovn-jenkins
/override ci/prow/okd-images

ausil:
I think this has actually made the release process harder by leaving the unsupported image streams in whatever state they are in. Any changes in the library will never get reflected, leading to bitrot, which is what has broken OKD for all releases. trying to remove the unsupported imagestreams turns out to be very difficult because there are tests in various places expecting them. We should either remove the ticking time bomb that is leaving things in an indeterminate state that may change out from under us, by removing any imagestreams that are unsupported and dealing with the CI fallout, or we need to keep syncing all things from library.

Prashanth684:
There are references to mysql in the origin repo. eg: https://github.com/openshift/origin/blob/master/test/extended/util/framework.go#L315. these need to be cleaned up if these streams are not going to be supported anymore

PR #561, opened 2024-08-26
OCPBUGS-38926: Update Go version, API version and vendored libs to the same versions as in the openshift/api.

metlos:
This is basically a copy of https://github.com/openshift/cluster-samples-operator/pull/552. I couldn't make that PR get tested with the updated testsuite from `openshift/api`, so I created a new PR to pick that up.


metlos:
oooh 😢

PR #562, opened 2024-08-27
OCPBUGS-38994: Fix library-sync.sh to handle renames in unsupported samples

metlos:
Improve the handling of changes in the unsupported samples so that no traces are left behind even when files were renamed or moved to a different location.


mfrancisc (approved):
/lgtm

PR #551, opened 2024-07-30
SO-124: Updating the assets for OCP 4.17

metlos:
Syncing the library in preparation for the SO release for OCP 4.17.

PR #567, opened 2024-08-28
OCPBUGS-39120: Updated the supported samples from openshift/library.

metlos:
`$TITLE`.

This is a remake of https://github.com/openshift/cluster-samples-operator/pull/551 but with the latest CI setup.


metlos:
This is already in through https://github.com/openshift/cluster-samples-operator/pull/569

PR #569, opened 2024-09-04
OKD-225: Sync samples - OKD CI is completely broken currently due to removed containers

ausil:
remove unsupported imagestreams from OKD.

These unsupported imagestreams and not synced by the library-sync.sh and
due to missing containers referenced from the repos the import is
failing causing all CI to be broken for OKD. so lets just remove them


Prashanth684:
`e2e-aws-ovn-image-ecosystem` test failure seems unrelated. happens in #567 as well.

metlos:
The image streams are pulled from https://github.com/openshift/library during our release and we don't maintain a separate list of images/tags that should/shouldn't be pulled. I think this is more of a problem with the OpenShift library than with the samples operator.

Prashanth684:
> The image streams are pulled from https://github.com/openshift/library during our release and we don't maintain a separate list of images/tags that should/shouldn't be pulled. I think this is more of a problem with the OpenShift library than with the samples operator.

https://github.com/openshift/cluster-samples-operator/commit/17530d21b2f06c80fcb2fbdd96fb0185f7e42b9b commit lists a set of supported imagestreams which is why imagestreams like mariadb, SQL and redis have been dropped. Given that is the case, this PR is correct? It does sync whatever is supported.

ausil:
> The image streams are pulled from https://github.com/openshift/library during our release and we don't maintain a separate list of images/tags that should/shouldn't be pulled. I think this is more of a problem with the OpenShift library than with the samples operator.

https://github.com/openshift/cluster-samples-operator/blob/master/library-sync.sh says that you do maintain a list of what is and is not synced. Ignoring and leaving in place old, unsupported content. Most of this PR comes from running that script. The second commit, removing old bits, comes from cleaning up stuff that is not updated by the script.

metlos (approved):

Prashanth684:
can we override the `e2e-aws-ovn-image-ecosystem` job @metlos ?

PR #570, opened 2024-09-10
OKD-225: Sync samples - OKD CI is completely broken currently due to removed containers

metlos:
This is a copy of https://github.com/openshift/cluster-samples-operator/pull/569 with additional commits to update the fixed samples.


ausil:
This can be closed as it is essentially the same as #569

PR #571, opened 2024-09-18
OKD-225: Add back mysql samples

ausil:
It turns out that there is a test for particular samples being imported into the imagestreams
https://github.com/openshift/origin/blob/master/test/extended/util/framework.go#L315 is the list. so lets add back mysql even though it is unsupported

### DIFF
--- a/assets/operator/okd-x86_64/cakephp/templates/cakephp-mysql-example.json
+++ b/assets/operator/okd-x86_64/cakephp/templates/cakephp-mysql-example.json
@@ -139,11 +139,12 @@
 			}
 		},
 		{
-			"apiVersion": "apps.openshift.io/v1",
-			"kind": "DeploymentConfig",
+			"apiVersion": "apps/v1",
+			"kind": "Deployment",
 			"metadata": {
 				"annotations": {
 					"description": "Defines how to deploy the application server",
+					"image.openshift.io/triggers": "[{\"from\":{\"kind\":\"ImageStreamTag\",\"name\":\"${NAME}:latest\"},\"fieldPath\": \"spec.template.spec.containers[0].image\"}]",
 					"template.alpha.openshift.io/wait-for-ready": "true"
 				},
 				"name": "${NAME}"
@@ -151,7 +152,9 @@
 			"spec": {
 				"replicas": 1,
 				"selector": {
-					"name": "${NAME}"
+					"matchLabels": {
+						"name": "${NAME}"
+					}
 				},
 				"strategy": {
 					"recreateParams": {
@@ -264,25 +267,7 @@
 							}
 						]
 					}
-				},
-				"triggers": [
-					{
-						"imageChangeParams": {
-							"automatic": true,
-							"containerNames": [
-								"cakephp-mysql-example"
-							],
-							"from": {
-								"kind": "ImageStreamTag",
-								"name": "${NAME}:latest"
-							}
-						},
-						"type": "ImageChange"
-					},
-					{
-						"type": "ConfigChange"
-					}
-				]
+				}
 			}
 		},
 		{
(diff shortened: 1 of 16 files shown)


Prashanth684:

metlos:
The failing image-ecosystem test is being fixed in openshift/origin#29116. But I wonder what breakage is https://github.com/openshift/origin/blob/master/test/extended/util/framework.go#L315 causing? From the OCP perspective, I'd much rather fix that test than bringing additional samples back to support.

Prashanth684:
> The failing image-ecosystem test is being fixed in [openshift/origin#29116](https://github.com/openshift/origin/pull/29116). But I wonder what breakage is https://github.com/openshift/origin/blob/master/test/extended/util/framework.go#L315 causing? From the OCP perspective, I'd much rather fix that test than bring additional samples back to support.

We add MySQL back to fix the breakage of that test or we remove MySQL from https://github.com/openshift/origin/blob/master/test/extended/util/framework.go#L315. @metlos are you suggesting removing MySQL from that test?

metlos:
> > The failing image-ecosystem test is being fixed in [openshift/origin#29116](https://github.com/openshift/origin/pull/29116). But I wonder what breakage is https://github.com/openshift/origin/blob/master/test/extended/util/framework.go#L315 causing? From the OCP perspective, I'd much rather fix that test than bring additional samples back to support.
> 
> We add MySQL back to fix the breakage of that test or we remove MySQL from https://github.com/openshift/origin/blob/master/test/extended/util/framework.go#L315. @metlos are you suggesting removing MySQL from that test?

Well, even though some of the imagestreams in that test are not supported, we keep them in the assets, even though we don't update them.

So at a glance, I didn't see what breakage that test might be causing. But I assumed I didn't look deep enough in the code so I asked what breakage is being fixed by updating the unsupported image streams.

The test doesn't need to be removed, just updated as soon as we "loose" some of the image streams for good (which IMHO didn't happen yet). Loosing the image stream would only happen if we removed all its versions due to EOL.

Prashanth684:
> > > The failing image-ecosystem test is being fixed in [openshift/origin#29116](https://github.com/openshift/origin/pull/29116). But I wonder what breakage is https://github.com/openshift/origin/blob/master/test/extended/util/framework.go#L315 causing? From the OCP perspective, I'd much rather fix that test than bring additional samples back to support.
> > 
> > 
> > We add MySQL back to fix the breakage of that test or we remove MySQL from https://github.com/openshift/origin/blob/master/test/extended/util/framework.go#L315. @metlos are you suggesting removing MySQL from that test?
> 
> Well, even though some of the imagestreams in that test are not supported, we keep them in the assets, even though we don't update them.
> 
> So at a glance, I didn't see what breakage that test might be causing. But I assumed I didn't look deep enough in the code so I asked what breakage is being fixed by updating the unsupported image streams.
> 
> The test doesn't need to be removed, just updated as soon as we "loose" some of the image streams for good (which IMHO didn't happen yet). Loosing the image stream would only happen if we removed all its versions due to EOL.

Right. So, we are adding MySQL back because the e2e test checks for its existence.

metlos:
There may be breakage in other parts of the testsuite in origin, if we update the samples from DeploymentConfig to Deployment. I'm not sure if mysql is used in any of the tests in https://github.com/openshift/origin/tree/master/test/extended/image_ecosystem or https://github.com/openshift/origin/tree/master/test/extended/builds or anywhere else.

Prashanth684:
> There may be breakage in other parts of the testsuite in origin, if we update the samples from DeploymentConfig to Deployment. I'm not sure if mysql is used in any of the tests in https://github.com/openshift/origin/tree/master/test/extended/image_ecosystem or https://github.com/openshift/origin/tree/master/test/extended/builds or anywhere else.

For okd, the extended test suite is not run and yes, we do need to be aware that if we ever do - this is a potential problem to solve. But I believe as far the conformance e2es are concerned, we are good

PR #572, opened 2024-09-18
OKD-225: Sync OKD to library for 4.17

ausil:
Sync all samples for 4.17 for okd including unsupported ones

### DIFF
--- a/assets/operator/okd-x86_64/cakephp/templates/cakephp-mysql-example.json
+++ b/assets/operator/okd-x86_64/cakephp/templates/cakephp-mysql-example.json
@@ -139,11 +139,12 @@
 			}
 		},
 		{
-			"apiVersion": "apps.openshift.io/v1",
-			"kind": "DeploymentConfig",
+			"apiVersion": "apps/v1",
+			"kind": "Deployment",
 			"metadata": {
 				"annotations": {
 					"description": "Defines how to deploy the application server",
+					"image.openshift.io/triggers": "[{\"from\":{\"kind\":\"ImageStreamTag\",\"name\":\"${NAME}:latest\"},\"fieldPath\": \"spec.template.spec.containers[0].image\"}]",
 					"template.alpha.openshift.io/wait-for-ready": "true"
 				},
 				"name": "${NAME}"
@@ -151,7 +152,9 @@
 			"spec": {
 				"replicas": 1,
 				"selector": {
-					"name": "${NAME}"
+					"matchLabels": {
+						"name": "${NAME}"
+					}
 				},
 				"strategy": {
 					"recreateParams": {
@@ -264,25 +267,7 @@
 							}
 						]
 					}
-				},
-				"triggers": [
-					{
-						"imageChangeParams": {
-							"automatic": true,
-							"containerNames": [
-								"cakephp-mysql-example"
-							],
-							"from": {
-								"kind": "ImageStreamTag",
-								"name": "${NAME}:latest"
-							}
-						},
-						"type": "ImageChange"
-					},
-					{
-						"type": "ConfigChange"
-					}
-				]
+				}
 			}
 		},
 		{
@@ -308,11 +293,12 @@
 			}
 		},
 		{
-			"apiVersion": "apps.openshift.io/v1",
-			"kind": "DeploymentConfig",
+			"apiVersion": "apps/v1",
+			"kind": "Deployment",
 			"metadata": {
 				"annotations": {
 					"description": "Defines how to deploy the database",
+					"image.openshift.io/triggers": "[{\"from\":{\"kind\":\"ImageStreamTag\",\"name\":\"mysql:${MYSQL_VERSION}\"},\"fieldPath\": \"spec.template.spec.containers[0].image\"}]",
 					"template.alpha.openshift.io/wait-for-ready": "true"
 				},
 				"name": "${DATABASE_SERVICE_NAME}"
@@ -320,7 +306,9 @@
 			"spec": {
 				"replicas": 1,
 				"selector": {
-					"name": "${DATABASE_SERVICE_NAME}"
+					"matchLabels": {
+						"name": "${DATABASE_SERVICE_NAME}"
+					}
 				},
 				"strategy": {
 					"type": "Recreate"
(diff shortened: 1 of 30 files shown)


Prashanth684:

metlos:
Why do we need to update the unsupported samples?

Prashanth684:
> Why do we need to update the unsupported samples?

The update is needed because the samples images for centos contained the old centos7 images which are now not available.

ausil:
On top of what Prashanth said. We see that tests are expecting the unsupported, at least some unsupported imagestreams to be there and given it is a stable release we are doing the least invasive change.

assets/operator/okd-x86_64/cakephp/templates/cakephp-mysql-example.json
```diff
@@ -139,19 +139,22 @@
 			}
 		},
 		{
-			"apiVersion": "apps.openshift.io/v1",
-			"kind": "DeploymentConfig",
```
metlos, line 143: This is going to cause breakage in origin that uses this template in a test and assumes `DeploymentConfig`. See https://github.com/openshift/origin/blob/master/test/extended/image_ecosystem/sample_repos.go#L173

ausil: The change came about from running library-sync. so the bug is likely elsewhere @Prashanth684 says that in the tests we run we should not hit that check

Prashanth684: @metlos does the `e2e-aws-ovn-image-ecosystem` job run this test ? or is the extended test suite not being tested in presubmits? I'm wondering how all tests passed

Prashanth684: test logs [here](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-samples-operator/572/pull-ci-openshift-cluster-samples-operator-release-4.17-e2e-aws-ovn-image-ecosystem/1836546856549617664/artifacts/e2e-aws-ovn-image-ecosystem/openshift-e2e-test/artifacts/e2e.log) for this run show that this test passed

metlos: This might be because I believe our tests don't use the OKD templates? Not sure.. @fbm3307 do you know?

Prashanth684: it probably uses the examples that are sourced locally: https://github.com/openshift/origin/tree/master/examples/quickstarts

metlos: that's weird, too, because for example rails quickstart uses ruby:3.0-ubi8 in its buildconfig and that should no longer be available...

metlos: if only there were logs even for successful runs...

metlos: Digging through https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-samples-operator/572/pull-ci-openshift-cluster-samples-operator-release-4.17-e2e-aws-ovn-image-ecosystem/1836546856549617664/artifacts/e2e-aws-ovn-image-ecosystem/openshift-e2e-test/artifacts/junit/resource-events_20240919-002049.zip, there's "e2e-rails-postgresql-repo-test-l6k4b" directory with events and those seem to show that the test used DeploymentConfig for the rails app. That means it definitely did not use the templates defined in this PR, nor did it use the templates from master.

Prashanth684: > This might be because I believe our tests don't use the OKD templates
This is correct because the changes are made to the okd templates which are only used when the cluster under test has okd installed on it. all the passing tests are ocp tests. I tried the test on my local system and it does use the imagestream on the cluster. This will be a problem if we add okd-scos jobs in the extended test suite for samples.

Prashanth684: i think we have a few options on how to pursue these backports:

- Merge this and make a note that if we ever add extended test suite jobs to test samples for OKD, they will break and tests need updating
- update ocp samples and update origin test cases which is a bigger effort . Given that this is a backport this option is not desirable
- make a PR manually just removing the centos7 bits for the OKD samples without running the library script

Prashanth684: @metlos as far as OKD is concerned - we do not have the image ecosystem test jobs for OKD. If we go ahead with this change, we just have to note the caveat that that job cannot be added for 4.17 and below without changes to the test suite. But it is a bit awkward for the libraries of 4.17 and 4.16 to be "ahead" of 4.18. Hence I proposed the 3rd option - which will do the bare minimum of removing the centos7 images.

I am not sure if samples are synced hereafter for GA'ed releases, but i'd like to hear from you which decision you would prefer.

metlos: @Prashanth684 I agree. Let's go with the minimal option and just remove the offending imagestreams/samples.

Prashanth684: thanks @metlos . https://github.com/openshift/cluster-samples-operator/pull/575 for 4.17

Prashanth684: i just noticed that the rails-postgresql had also changed from DeploymentConfig to Deployment through the original [PR](https://github.com/openshift/cluster-samples-operator/pull/569/files#diff-defc51d2ab4428824c502c965fc188a5902011b82ab83dfe4cac6ea3205dab2dR141) and might fail ecosystem-tests: https://github.com/openshift/origin/blob/master/test/extended/image_ecosystem/sample_repos.go#L128

metlos: Interestingly, image-ecosystem is passing in https://github.com/openshift/origin/pull/29116 (which is my PR to fix the issues with rails and perl).

metlos:
https://github.com/openshift/cluster-samples-operator/pull/575 has been approved, so I believe this can be closed, right?

PR #573, opened 2024-09-18
OKD-225: Sync OKD to library  for 4.16

ausil:
Sync all samples for 4.16 for okd including unsupported ones


Prashanth684:

metlos:
/close

I think this can be closed. The minimal 4.16 changes for OKD are defined here: https://github.com/openshift/cluster-samples-operator/pull/576

PR #574, opened 2024-09-20
OCPBUGS-42265: Fix the rails template before the upstream accepts or PR

metlos:
Fix the rails template before the upstream accepts our PR (https://github.com/sclorg/rails-ex/pull/178).


metlos:
I am going to override the image ecosystem test so that our changes in https://github.com/openshift/origin/pull/29116 have a chance to pass. 

Currently we're in a Head-22 situation with these PRs. 

The image ecosystem test in this PR uses the master of CI in openshift/origin that looks for a DeploymentConfig in Rails and uses perl:5.30-el7 for the Perl test. This won't pass because Rails now uses Deployment and per:5.30-el7 is not available anymore.

On the other hand, https://github.com/openshift/origin/pull/29116 will not pass until we fix the rails templates in this PR.

PR #575, opened 2024-09-20
OKD-225: remove only the EOL CentOS 7 images

ausil:
Remove all references to the EOL CentOS 7 images that have been removed, update the redis image to a  scl.org c9s one, as the CentOS 7 one was the only one available


ausil:
@metlos this is I think the smallest change we could make to get 4.17 moving again

PR #577, opened 2024-09-23
OCPBUGS-39120: Sync the non-OKD assets from master to release-4.17

metlos:
This brings the 4.17 branch up-to-date with master prior to 4.17 release.


metlos:
As reasoned in https://github.com/openshift/cluster-samples-operator/pull/574#issuecomment-2364581403 I am going to override the image ecosystem test because the failures it exhibits are fixed in the PR https://github.com/openshift/origin/pull/29126 which itself requires changes introduced in this PR.

The gordian knot needs to be cut.

PR #581, opened 2024-10-07
SO-137: Support updating the library separately for OKD and for OCP.

metlos:
This changes `library-sync.sh` such that by default, it only updates the supported samples in OCP for all architectures.

If the `--okd` parameter is supplied to the script, all samples in OKD are updated (there are no unsupported samples in OKD).

### DIFF
--- a/library-sync.sh
+++ b/library-sync.sh
@@ -3,6 +3,20 @@
 # utility script to gather template/imagestream content from https://github.com/openshift/library
 # and store it in this repo (cannot access other repos with dist git, and advised against git submodules
 
+# If the --okd CLI argument is supplied, this script only updates all the samples in the assets/operator/okd-x86_64.
+# When the argument is not supplied, it only updates the OCP samples.
+# It never updates both at the same time, so if you need that, you have to run the script twice.
+
+PROCESS_OKD="false"
+while [[ $# -gt 0 ]]; do
+  case $1 in
+  --okd)
+    PROCESS_OKD="true"
+    ;;
+  esac
+  shift
+done
+
 pushd assets
 wget https://github.com/openshift/library/archive/master.zip -O library.zip
 unzip library.zip
@@ -53,23 +67,43 @@ git add operator
 rm t.tar
 rm -rf library-master
 
+function reset_directory() {
+  git checkout HEAD -- "$1"
+  # remove any changes from the working tree in this directory that checkout left behind
+  git stash -a -- "$1"
+  git stash drop
+}
+
 SUPPORTED="ruby python nodejs perl php httpd nginx eap java webserver dotnet golang rails"
 function reset_unsupported() {
+  # there are no unsupported samples in OKD
+  if $PROCESS_OKD; then
+    return 0
+  fi
+
   for d in $(ls); do
     if [[ "${SUPPORTED}" != *"${d}"* ]]; then
-      git checkout HEAD -- "${d}"
-      # remove any changes from the working tree in this directory that checkout left behind
-      git stash -a -- "${d}"
-      git stash drop
+      reset_directory "${d}"
     fi
   done
 }
 
-ARCHS="ocp-x86_64 ocp-aarch64 ocp-ppc64le ocp-s390x okd-x86_64"
+ALL_ARCHS="ocp-x86_64 ocp-aarch64 ocp-ppc64le ocp-s390x okd-x86_64"
+if $PROCESS_OKD; then
+  SUPPORTED_ARCHS="okd-x86_64"
+else
+  SUPPORTED_ARCHS="ocp-x86_64 ocp-aarch64 ocp-ppc64le ocp-s390x"
+fi
+
 pushd operator
-for arch in $ARCHS; do
-  pushd "${arch}"
-  reset_unsupported
-  popd # $arch
+for arch in $ALL_ARCHS; do
+  if [[ "${SUPPORTED_ARCHS}" != *"${arch}"* ]]; then
+    reset_directory "${arch}"
+  else
+    pushd "${arch}"
+    reset_unsupported
+    popd # $arch
+  fi
 done
+
 popd # operator
--- a/library-sync.sh
+++ b/library-sync.sh
@@ -3,20 +3,71 @@
 # utility script to gather template/imagestream content from https://github.com/openshift/library
 # and store it in this repo (cannot access other repos with dist git, and advised against git submodules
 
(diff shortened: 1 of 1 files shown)


metlos:
@Prashanth684 does this look likes something that would improve our "release experience" the next time? 

For OKD, you'd run `./library-sync.sh --okd` which would update all the samples in OKD (not resetting any "unsupported" ones).

library-sync.sh
```diff
@@ -53,23 +67,43 @@ git add operator
...
+function reset_directory() {
+  git checkout HEAD -- "$1"
+  # remove any changes from the working tree in this directory that checkout left behind
+  git stash -a -- "$1"
+  git stash drop
+}
+
 SUPPORTED="ruby python nodejs perl php httpd nginx eap java webserver dotnet golang rails"
 function reset_unsupported() {
+  # there are no unsupported samples in OKD
```
Prashanth684, line 79: how do we define what is supported vs unsupported? We would want this to be the same for OKD and OCP. If the images are no longer maintained and supported, that should be the case for both. I don't think OKD should behave differently from OCP.

Prashanth684: Also if it is unsupported, why are they not removed? Wouldn't the images eventually be removed if they are not supported? wouldn't this be a problem for automation tests?

metlos: I was basing this assumption on what was done prior to the 4.17 release. Upon your request, all of the samples from the library were updated in OKD, while only the supported ones were updated for OCP. Supported here means "receiving customer support by Red Hat".

The decision to let the unsupported samples stay in without any updates in OCP was made prior to me joining the project, so I'll let @fbm3307 or @dperaza4dustbit comment on that.

fbm3307: @Prashanth684 , there are only 12 s2i samples for OCP which are being supported as per the list given above.
As for removing them, thats something of a larger effort, and since the whole samples operator is being deprecated and discussion going around for its removal, we did not remove them previously. This same has been announced to the customers also in 4.13 and 4.16 releases

library-sync.sh
```diff
@@ -3,6 +3,71 @@
...
+
+while [[ $# -gt 0 ]]; do
+  case $1 in
+  --okd)
+    PROCESS_OKD="true"
+    ;;
+  --ocp)
+    PROCESS_OCP="true"
+    ;;
+  --ocp-unsupported)
```
mfrancisc, line 31: minor - I would maybe call this `--ocp-all` as this names might suggest we are updating ONLY the unsopported ones.

library-sync.sh
```diff
@@ -3,6 +3,71 @@
...
+function reset_directory() {
+  git checkout HEAD -- "$1"
+  # remove any changes from the working tree in this directory that checkout left behind
+  git stash -a -- "$1"
+  git stash drop
+}
+
+function reset_ocp_unsupported() {
+  for d in $(ls); do
+    if [[ "${OCP_SUPPORTED_SAMPLES}" != *"${d}"* ]]; then
```
mfrancisc, line 64: might this reset the okd supported ones as well ?

library-sync.sh
```diff
@@ -3,6 +3,71 @@
...
+OKD_ARCHS="okd-x86_64"
+ALL_ARCHS="$OCP_ARCHS $OKD_ARCHS"
+if $PROCESS_OKD; then
+  if $PROCESS_OCP; then
+    SUPPORTED_ARCHS=$ALL_ARCHS
+  else
+    SUPPORTED_ARCHS=$OKD_ARCHS
+  fi
+else
+  SUPPORTED_ARCHS=$OCP_ARCHS
```
mfrancisc, line 51: it seems like by default it will synch all ocp archs , I mean if no flags are provided . Is this intentional ?

mfrancisc (commented):
Nice job 👍 

I have few minor comments/questions only.

library-sync.sh
metlos: This will just keep the changes only to the samples that are supported in OCP, regardless of the architecture. In another words, this function doesn't care about the architecture, it just resets the unsupported samples in whatever arch it happens to find itself in.

library-sync.sh
metlos: I will rename the `SUPPORTED_ARCHS` to `PROCESSED_ARCHS` or something, because the "supported" is being used for 2 different things in the script which is confusing. There is a limited set of "supported ocp samples" and there's a list of architectures that is being processed in particular script run.

metlos: Forgot to answer your actual question. Yes, by default the script will only process supported OCP samples in all OCP architectures. It leaves out OKD by default.

library-sync.sh
Prashanth684: > Upon your request, all of the samples from the library were updated in OKD, while only the supported ones were updated for OCP. Supported here means "receiving customer support by Red Hat".

I do understand this stance. While I like the flexibility of OKD deciding the samples it wants to import, I'm worried about the skew between OCP and OKD samples, which might be a moot point if samples is being deprecated soon.

Prashanth684: in other words..i would like OCP and OKD to be in sync. In understand that this requires updates in CI tests etc..but that would prevent this need.

@fbm3307 is the samples deprecation happening anytime soon?

library-sync.sh
mfrancisc: right it's minor thing - but it seems that you run this function also for the okd folder at line 129, if there is a bug in the logic you end up resetting those as well. maybe it's an edge case, but you could explicitly exclude okd folders or include only the ocp unsupported folder in the loop which calls this function. 

Anyway feel free to ignore my comment, as said it's not a big issue.

library-sync.sh
```diff
@@ -3,6 +3,71 @@
 # utility script to gather template/imagestream content from https://github.com/openshift/library
 # and store it in this repo (cannot access other repos with dist git, and advised against git submodules
 
+# Without any commandline arguments, this script only updates the OCP supported samples. This behavior
+# can be modified using these commandline arguments:
+# * --okd - the OKD samples are updated
+# * --ocp - the OCP supported samples are updated
+# * --ocp-all - all OCP samples are updated including the unsupported ones
```
mfrancisc, line 10: minor thing - I would include these descriptions in a `-h` message or something like that, if the script is meant also for local usage as I believe.

mfrancisc (approved):
Looks good to me 👍 

Thanks for addressing my comments and answering my questions 🙏

PR #584, opened 2024-11-04
Update the k8s API version

metlos:
update `k8s.io/api`, `k8s.io/apimachinery`, `k8s.io/apiserver` and `k8s.io/client-go` to `v0.31.1` to be in line with OpenShift.


mfrancisc (approved):
/lgtm

PR #587, opened 2024-11-25
Revert "update k8s.io/api"

metlos:
This reverts commit 26090283fcf0f4da2b90bc6a6d24d51c3e13a241.

There is a regression in OpenShift tracked by https://issues.redhat.com/browse/OCPBUGS-44693 which might be caused by this. Let's try to revert to the older version of the k8s api to see if it fixes the problem.


mfrancisc (approved):

metlos:
I think we might be reproducing the issue with high watch api request counts here, too. The `e2e-aws-operator` job has 49 watch requests on imagestreams and 40 on templates where a normal default installation of 4.17 has 8 and 21 respectively.

metlos:
Maybe even better comparison is with the `e2e-aws-operator` from the https://github.com/openshift/cluster-samples-operator/pull/584 (the e2e-aws-operator build of that PR is here https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-samples-operator/584/pull-ci-openshift-cluster-samples-operator-master-e2e-aws-operator/1853432224435671040). There the counts are very low - 8 imagestreams and 10 templates.

metlos:
Finally, in PR https://github.com/openshift/cluster-samples-operator/pull/583 which has the same k8s API version as this PR, the counts are 5 for imagestreams and 5 for templates.

metlos:
I'm holding this PR from merging. Chances are the regression is outside of the operator and we won't need this PR.

metlos:
I'm closing this PR because the increased watch counts are not directly caused by the samples operator. See the JIRA issue for further details.

PR #588, opened 2024-11-26
DO NOT MERGE - experimental revert to 4.17 state

mfrancisc:
This PR reverts the commits to the 4.17 state at 66ea9ab5.


metlos:
/hold

This probably won't be needed because the regression seems to be outside of the samples operator.

metlos:
/close

I'm closing this because the regression in OpenShift is not caused by the samples operator. See https://issues.redhat.com/browse/OCPBUGS-44693 for further details.

PR #631, opened 2025-04-14
OCPBUGS-59324: Periodic sync of supported OCP samples

aroyoredhat:
Syncing the latest supported OCP samples from the library.

Created using https://github.com/aroyoredhat/trigger-test-samples-build

### DIFF
--- a/assets/operator/ocp-aarch64/ruby/imagestreams/ruby-rhel-aarch64.json
+++ b/assets/operator/ocp-aarch64/ruby/imagestreams/ruby-rhel-aarch64.json
@@ -35,27 +35,6 @@
 				}
 			},
 			{
-				"name": "3.1-ubi8",
-				"annotations": {
-					"description": "Build and run Ruby 3.1 applications on UBI 8. For more information about using this builder image, including OpenShift considerations, see https://github.com/sclorg/s2i-ruby-container/blob/master/3.1/README.md.",
-					"iconClass": "icon-ruby",
-					"openshift.io/display-name": "Ruby 3.1 (UBI 8)",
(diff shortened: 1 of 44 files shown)


assets/operator/ocp-aarch64/ruby/imagestreams/ruby-rhel-aarch64.json
```diff
@@ -34,27 +34,6 @@
 					"type": "Local"
 				}
 			},
-			{
```
metlos, line 37: This is gonna break a lot of stuff:
* rails samples depend on this, e.g.: https://github.com/openshift/cluster-samples-operator/blob/main/assets/operator/ocp-x86_64/rails/templates/rails-pgsql-persistent.json#L534 (which is also referenced in `openshift/origin` https://github.com/openshift/origin/blob/main/test/extended/image_ecosystem/s2i_ruby.go#L16, https://github.com/openshift/origin/blob/main/test/extended/image_ecosystem/sample_repos.go#L143 and possibly elsewhere.
* This ruby version is also used in the `openshift/origin` tests like https://github.com/openshift/origin/blob/main/test/extended/image_ecosystem/s2i_images.go#L36

assets/operator/ocp-x86_64/php/imagestreams/php-rhel.json
```diff
@@ -98,19 +98,19 @@
 				}
 			},
 			{
-				"name": "8.1-ubi9",
```
aroyoredhat, line 101: This needs to be fixed on origin as it is making the tests to fail and start using 8.2-ubi9 on origin

PR #684, opened 2026-04-02
CNTRLPLANE-3176: implements operator tls config compliance

ricardomaraschini:
Refactor metrics server to accept a `configv1.GenericControllerConfig` for configuration. Key improvements include early bind failure detection, certificate validation at construction time, and support for TLS version and cipher suite configuration.

This is part of PQC compliance work.


cmd/cluster-samples-operator/main.go
```diff
@@ -27,28 +32,86 @@ func main() {
...
+	if err := srv.Start(); err != nil {
+		logrus.Fatal(err)
+	}
+	defer func() {
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		if err := srv.Stop(ctx); err != nil {
+			logrus.Error(err)
+		}
+		cancel()
+	}()
```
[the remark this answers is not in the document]
ricardomaraschini: This isn't necessary. The cancel will always be called.

cmd/cluster-samples-operator/main.go
```diff
@@ -27,28 +32,86 @@ func main() {
...
+		return config, nil
+	}
+
+	content, err := os.ReadFile(path)
+	if err != nil {
+		return config, fmt.Errorf("failed to read config file: %w", err)
+	}
+
+	if err := kubeyaml.Unmarshal(content, &config); err != nil {
+		return config, fmt.Errorf("failed to unmarshal config content: %w", err)
```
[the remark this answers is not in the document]
ricardomaraschini: We can't use the strict unmarshal on this case due to type meta.

cmd/cluster-samples-operator/main.go
```diff
@@ -27,28 +32,86 @@ func main() {
...
+	}
+
+	// make sure we always have a bind address present in the config. This
+	// is just for the case where the config has an explicitly empty
+	// bindAddress value.
+	if config.ServingInfo.BindAddress == "" {
+		config.ServingInfo.BindAddress = fmt.Sprintf(":%d", metrics.MetricsPort)
+	}
+
+	return config, nil
```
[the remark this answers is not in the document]
ricardomaraschini: Architecturally speaking the config can't be edited. The Cluster Version Operator will roll it back if someone edits it so I don't think this is a real concern we need to have here.

pkg/metrics/server_test.go
```diff
@@ -1,135 +1,133 @@
...
-	if err != nil {
-		return "", "", err
+func newInsecureClient() *http.Client {
+	return &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: true,
+			},
+		},
 	}
```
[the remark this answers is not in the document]
ricardomaraschini: Done.

gangwgr:
Pre-merge testing completed

manifests/06-operator-ibm-cloud-managed.yaml
```diff
@@ -72,6 +78,11 @@ spec:
             drop:
             - ALL
         terminationMessagePolicy: FallbackToLogsOnError
+        volumeMounts:
+        - mountPath: /etc/secrets
```
ingvagabund, line 82: `/etc/secrets` has not been mounted before? Is this file even used/rendered by CVO?

ricardomaraschini: Yes, you read it right. It was not mounted and we have a container watching `/etc/secrets/tls.crt` and `/etc/secrets/tls.key`. This is true for both `06-operator.yaml` and `06-operator-ibm-cloud-managed.yaml`. I don't see how it would be deployed without CVO rendering at the very least `06-operator.yaml`.

ricardomaraschini: It seems like this container was introduced in the manifest without the `volumeMount`. See https://github.com/openshift/cluster-samples-operator/commit/712740b2ca5d07b91e379abcd16d67f0cde7bbea . How did it pass the QE testing at the time is the question.

**edit**: rephrase without losing essence.

cmd/cluster-samples-operator/main.go
```diff
@@ -27,28 +33,86 @@ func main() {
...
+		return config, nil
+	}
+
+	content, err := os.ReadFile(path)
+	if err != nil {
+		return config, fmt.Errorf("failed to read config file: %w", err)
+	}
+
+	if err := kubeyaml.Unmarshal(content, &config); err != nil {
+		return config, fmt.Errorf("failed to unmarshal config content: %w", err)
```
vincentdephily: Should we even return a `config` here instead of `nil` ? AFAIK it could have been partially updated by `Unmarshal()`. Just a nit-pick, as the caller exits on `err!=nil` anyway.

ingvagabund: The same for `os.ReadFile`

ingvagabund (commented):
Just the nits. Looks good otherwise.

cmd/cluster-samples-operator/main.go
ricardomaraschini: done.

ricardomaraschini:
This test seems to be failing across the board, even on different PRs. As ti has already passed in the past on this PR I am overriding it.

/override ci/prow/e2e-aws

